PR #73, opened 2014-08-13
Date expression

Rodac:
Add support for PostgreSQL syntax

``` sql
EXTRACT(datepart FROM date 'yyyy-mm-dd')
```

or 

``` sql
EXTRACT(datepart FROM date'yyyy-mm-dd')
```


wumpz:
Already solved.

PR #77, opened 2014-09-24
Enable parsing of escaped single quotes within string values.

Maascamp:
The current parser breaks when parsing values of the form `'There\'s an escaped single quote here.'` This change respects backslash escapes of single quotes within string values.


wumpz:
First try of modified character literal parsing accepts your SQL.

PR #118, opened 2015-03-26
oracle join (+) vs. ( + ) fix #117

### LINKED ISSUE #117: oracle join (+) vs. ( + )
I see, there is support for (+) syntax, but if in the statement "( + )" is instead of "(+)" then the parser throws ParseException. In other words:

The parser is good for this query:

        Select * From dual d1, dual d2 where d1.dummy = d2.dummy(+);

But the parser throws ParseException for this query:

        Select * From dual d1, dual d2 where d1.dummy = d2.dummy( + );

karsany:
the problem is that this patch does not fix the (+ ) ( +) (     +  ) etc. forms. - I don't know how to do it in javacc


wumpz:
I will look into it.

wumpz:
I made the implementation myself. It will all the forms, you mentioned.

karsany:
thank you

PR #148, opened 2015-07-09
Fix inline usage of foreign keys in CREATE TABLE statements

weiznich:
The parser is crashing if a CREATE TABLE statement contains a inlined foreign key(see testcases).


wumpz:
Thx for contributing. I will look into it.

PR #155, opened 2015-07-25
Implement support for "SELECT SKIP <OFFSET> FIRST <LIMIT>..." co…

matthiasblaesing:
…nstruct

This is the informix analogue to the mysql "limit <OFFSET>, <LIMIT>" construct. Unittests were added.


wumpz:
Isn't your implementation to limited using only numbers and JdbcParameters? I am not familiar with Informix but could one use some variables there (limit @a)? Would'nt it be better to use an expression there? Cheers Tobias

PR #197, opened 2015-11-24
Issue 195: support for ORDER BY and LIMIT, in UPDATE and DELETE statements

chiastic-security:
Add support for ORDER BY and LIMIT in UPDATE and DELETE statements (as supported by MySQL).

LimitDeparser and OrderByDeParser have been pulled out into separate classes to avoid code duplication from SelectDeParser.


wumpz:
Nice.

PR #202, opened 2015-12-07
table functions

ftaras:
Added support of H2 table functions (http://www.h2database.com/html/features.html section 'Using a function as a table')
Ex: 'SELECT * FROM MATRIX(4)'


wumpz:
Why did you not use a FunctionItem?

ftaras:
@wumpz Fixed. Now TableFunction extends FunctionItem and implements FromItem. 
The other option is to remove TableFunction class and use FunctionItem, but it requires FunctionItem to implement FromItem. Not sure if it is a good idea.

PR #216, opened 2016-01-30
Added reference options foreign keys support (ON UPDATE/DELETE NO ACT…

pabloa:
…ION/CASCADE) and Full text indexes (FULLTEXT idx(text1))


pabloa:
Here it is a new version of the patch in 1 commit operation and without merged commits as you requested.

PR #254, opened 2016-04-26
Oracle relational operators with spaces & updates to MERGE INTO

davelindquist-egistix:
Includes two changes:

**Added ability to have operators like '>=' or '<=' separated by a space.  This includes:**

- Modifying the JJT syntax to support the 'space in the middle' versions of operators.
- Modifying the various operators to inherit from a new 'ComparisonOperator' class, which handles the (previously NotEqualsTo-only) logic for capturing the form of the operator.
- Giving each of the various operators a constructor that accepts the syntax used.
- Modifying TestUtils to strip comments out before comparing SQL text (necessary because condition07.sql is now passing, and has a comment).
- Updating SpecialOracleTest to indicate 130 tests passing now (condition7.sql now passes).
- Adding a new test specifically for operators into SpecialOracleTest.

**Corrected "MERGE INTO" parsing for more complicated statements.  Specifically:**

- Changed "Condition" to "Expression" for the "ON" clause -- this is needed to handle "ON" clauses that have "a = y AND b = z" or other more complicated expressions (basically the same as the "ON" clause in a SELECT query).
- Corrected the toString on the MergeUpdate clause, which was missing a comma between the fields.
- Added a new, more complicated MERGE INTO statement to the MergeTest class.

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -284,7 +284,7 @@ Statement SingleStatement() :
 	(
     stm = Select()
     |
-    stm =  Update()
+    stm = Update()
     |
     stm = Insert()
     |
@@ -292,7 +292,8 @@ Statement SingleStatement() :
     |
     stm = Replace()
     |
-	stm = Alter()
+    LOOKAHEAD(2)
+	stm = AlterTable()
     |
     stm = Merge()
 	|
@@ -304,6 +305,8 @@ Statement SingleStatement() :
     |
 	LOOKAHEAD(2)
 	stm = CreateView()
+	|
+    stm = AlterView()
 	|
     stm = Drop()
     |
@@ -2026,15 +2029,15 @@ Expression PrimaryExpression():
 
     | LOOKAHEAD([sign="+" | sign="-"] JsonExpression()) [sign="+" | sign="-"] retval=JsonExpression()
 
-	| LOOKAHEAD(["+" | "-"] Function())       [sign="+" | sign="-"] retval=Function()
+	| LOOKAHEAD(["+" | "-"] Function()) [sign="+" | sign="-"] retval=Function()
 
-	| LOOKAHEAD(["+" | "-"] <S_DOUBLE>)       [sign="+" | sign="-"] token=<S_DOUBLE>  { retval = new DoubleValue(token.image); }
+	| LOOKAHEAD(2) [sign="+" | sign="-"] token=<S_DOUBLE>  { retval = new DoubleValue(token.image); }
 
-	| LOOKAHEAD(["+" | "-"] <S_LONG>)         [sign="+" | sign="-"] token=<S_LONG> { retval = new LongValue(token.image); }
+	| LOOKAHEAD(2) [sign="+" | sign="-"] token=<S_LONG> { retval = new LongValue(token.image); }
 
-    | LOOKAHEAD(["+" | "-"] <S_HEX>)          [sign="+" | sign="-"] token=<S_HEX>  { retval = new HexValue(token.image); }
+    | LOOKAHEAD(2) [sign="+" | sign="-"] token=<S_HEX>  { retval = new HexValue(token.image); }
 
-	| LOOKAHEAD(["+" | "-"] CastExpression()) [sign="+" | sign="-"] retval=CastExpression()
+	| LOOKAHEAD(2) [sign="+" | sign="-"] retval=CastExpression()
 
 	| LOOKAHEAD(["+" | "-"] Column())         [sign="+" | sign="-"] retval=Column()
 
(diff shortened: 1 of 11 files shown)


src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
```diff
@@ -1616,9 +1616,13 @@ Expression RegularCondition():
 	| "<" { result = new MinorThan(); }
 	| "=" { result = new EqualsTo(); }
 	| ">=" { result = new GreaterThanEquals(); }
+	| "> =" { result = new GreaterThanEquals("> ="); }
 	| "<=" { result = new MinorThanEquals(); }
+	| "< =" { result = new MinorThanEquals("< ="); }
 	| "<>" { result = new NotEqualsTo(); }
+	| "< >" { result = new NotEqualsTo("< >"); }
```
wumpz: Isnt this only half the way? Oracle allows, as I learned right now :), multiple whitespaces between < and =. Addionaly wouldnt it be better to map <..= to <=? Is there a case one need to reproduce oracles special space separated syntax?

davelindquist-egistix: Yes, you are correct that I was handling only a single space.  Ideally, I should handle any whitespace between it – any tips for me on what that would look like in JavaCC?  (I’m having difficulty reconciling the ‘skip’ characters with the insertion of variable whitespace in the token…)

I had originally started by mapping “< =” to just GreaterThanEquals(), but that unfortunately loses information about the ‘style’ of the operator (for example, I can’t go back and slap the hands of the developer for using non-standard relational operators).  I elected to follow the pattern of “NotEqualsTo”, which passed the actual operator text into the constructor.  Is that OK?  Did you have a preference for an alternate way of doing this?

Thanks!

tejoe:
There are two errors here.
First oracle actually allows an arbitrary number of spaces in operators, so 
"<=" equals "< =" equals "<                 ="
you only considered a single space

Second and this one is much more remarkable "! =" in oracle is not the same like "! =" in postgresql. In Postgres "! =" is not an operator. It is two operators faculty and equals. 

**ORACLE**
select * from dual where 1 ! = 1
returns no row, because the condition evaluates to false

**POSTGRESQL**
select 1 where 1 ! = 1
returns a row, because the faculty of 1 equals 1

davelindquist-egistix:
Hmm, that is interesting – I had not realized that postgres would actually interpret that as something.

Can you explain to me what the “! =” does in postgres?  (Or point me to the docs?)  I’m not quite clear on how it functions, and wasn’t able to turn anything up on a quick google.

In that case, it seems to me that we’d have to handle “! =” in the postgresql way?  (As two operators?)

Thanks!

tejoe:
here is the operator docu for postgres:
http://www.postgresql.org/docs/9.5/static/typeconv-oper.html
example 10-1 will show you the "!" (factorial) operator

If there is an blank between ! and = postgres will always use the ! as factorial. I have no documentation for this, this is what I observed.

davelindquist-egistix:
Thanks!  I think I turned up the same page a second before I got your reply. :-)

src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
davelindquist-egistix: I'll provide an updated commit to address:
a. The 'any amount of whitespace' issue (between parts of the operators)
b. Removal of "space between ! and =", as this means something different in PostgreSQL

davelindquist-egistix:
OK, I've pushed another couple of commits -- this one allows for truly variable whitespace in the operators, and retracts the "! =" case, as it means something different in PostgreSQL  (I haven't tackled adding the "factorial" operator, which sounds fun.)

Feedback would be appreciated! :-)

wumpz:
Again the question, why would one preserve the whitespaces? Isn't it for parsing purposes enough to map <  = to <= and therefore to less and equal?

Could you split your pull request into two. It is much easier to handle if I get one for one task or solution.

davelindquist-egistix:
Sure, I'll split the pull request into two.

With regards to preserving the whitespace:
a. The test cases currently require preservation of the whitespace -- the parse/deparse test structure will fail if the whitespace is stripped.  If we elect to strip the whitespace, I imagine we'll need to modify the parse/deparse tests to do a regex-replace to remove whitespace?
b. The pre-existing NotEqualsTo structure established a pattern of preserving the operator (that is, it does not collapse "!=" into "<>"), so I followed that pattern.
c. For my particular usage, it is nice to be able to detect that a 'non-standard' form was used for the operator -- preserving the exact form of the token parsed as the operator was the easiest way to do so.

In order to prevent rework, could you give me an idea of what you'd like to see?  For example:

1. The existing structure, preserving the token that was parsed, OR
2. A modified structure that does not preserve the token that was parsed, but does provide a flag indicating that the operator was "non-standard", OR
3. Neither, just collapse the operators into standard form regardless of style used.

Additional Question: In either of cases 2 or 3, should we then collapse "!=" into "<>" as well?

(I don't have any super-strong preferences in either direction -- I'd prefer to have 1 or 2 over 3, but I can live with 3 if it is what you'd like.)

Thanks!

PR #269, opened 2016-05-19
Add support for Hive's LEFT SEMI JOIN

nhanitvn:
HiveQL provides a type of JOIN named LEFT SEMI JOIN as a way to write IN/EXIST clause.

> LEFT SEMI JOIN implements the uncorrelated IN/EXISTS subquery semantics in an efficient way. As of Hive 0.13 the IN/NOT IN/EXISTS/NOT EXISTS operators are supported using subqueries so most of these JOINs don't have to be performed manually anymore. The restrictions of using LEFT SEMI JOIN is that the right-hand-side table should only be referenced in the join condition (ON-clause), but not in WHERE- or SELECT-clauses etc.

More information [here](https://cwiki.apache.org/confluence/display/Hive/LanguageManual+Joins)

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/select/Join.java
+++ b/src/main/java/net/sf/jsqlparser/statement/select/Join.java
@@ -39,6 +39,7 @@ public class Join {
 	private boolean inner = false;
 	private boolean simple = false;
 	private boolean cross = false;
+	private boolean semi = false;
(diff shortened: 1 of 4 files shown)


wumpz:
Thx for improving JSqlParser

nhanitvn:
Awesome, thanks @wumpz

PR #276, opened 2016-06-01
Date expressions

ChrissW-R1:
Added ANSI SQL scalar functions CURRENT_TIMESTAMP, CURRENT_TIME and CURRENT_DATE


wumpz:
Please sqash all connected commits to one. Make a pull request for each change (means issue). You know although github does state your PR as mergable, I still want to check the changes. 

Why should something like **CURRENT_TIMESTAMP** be a keyword? This could be easily handeled as a normal identifier or column name. This would be the more general approach too, since **CURRENT_TIMESTAMP** is a very SQLServer specific one. So each time an oracle user defines a column **CURRENT_TIMESTAMP** he will get it as a function.

ChrissW-R1:
As you can see in http://users.atw.hu/sqlnut/sqlnut2-chp-4-sect-4.html#sqlnut2-CHP-4-SECT-4.1 not only the MS SQL Server supports this scalar functions.
We make this changes, because these functions were returned as column names!

wumpz:
As I said (column names), were is the problem? Saying that I at the moment learned that CURRENT_TIMESTAMP is ANSI Sql. So provide a single squashed commit pr for exactly this issue and I will merge it.

ChrissW-R1:
New pull request: #280

PR #280, opened 2016-06-08
date expressions of ANSI SQL

ChrissW-R1:


ChrissW-R1:
New pull request, because squash were not possible #276

src/main/java/net/sf/jsqlparser/expression/TimeKeyExpression.java
```diff
@@ -0,0 +1,49 @@
...
+ * GNU General Lesser Public License for more details.
+ * 
+ * You should have received a copy of the GNU General Lesser Public
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ * #L%
+ */
+package net.sf.jsqlparser.expression;
+
+public class TimeKeyExpression implements Expression {
```
wumpz: Wouldnt it be better if TimeKeyExpression extends function, or why implement an extra expression for it and not just accept it as a function? That is my last question, I swear ;).

PR #305, opened 2016-07-11
Add a double precision cast type

opeongo:
Postgres SQL allow casting to a "double precision" type.  This modification implements this syntax.


opeongo:
I will add a test for this if you can give me some direction as to what
file I should modify.

wumpz:
I added a test alreade :).

PR #322, opened 2016-08-06
Delete from table using join to another table

luksrn:
Support for parse delete from table using join to another table like:

DELETE posts
FROM posts
INNER JOIN projects ON projects.project_id = posts.project_id
WHERE projects.client_id = :client_id

This necessitated some changes to the DeleteTest class, specifically:

JSqlParserCC.jjt - changes on grammar of Delete statements.
Delete toString and DeleteDeParser.


luksrn:
I was thinking of treating the specific situation of Postgres (**using** instead **join**) in a second pull request , but decided to send this before to get some feedback necessary . Even if this request will be approved.

DELETE FROM films USING producers
  WHERE producer_id = producers.id AND producers.name = ' foo' ;

PR #323, opened 2016-08-11
Update Alter to support Postgres ADD/DROP expressions

wrobstory:
Hi JsqlParser maintainers!

Postgres currently supports multiple ADD/DROP expressions in a single ALTER statement: https://www.postgresql.org/docs/9.5/static/sql-altertable.html with a syntax like 
```sql
ALTER TABLE mytable ADD COLUMN col1 varchar (255), ADD COLUMN col2 integer;
```

This PR adds support for multiple expressions. The highlights: 
* Most of the logic that previously lived in the `Alter` class now lives in the `AlterExpression` class, and the `Alter` class is more or less a container for multiple `AlterExpression`s. 
* Each `AlterExpression.toString` call will return the entire expression, and the `Alter` class builds the statement by stacking these expression strings. 
* The .jjt file has been updated to move most of the `Alter` logic into `AlterExpression` as well, and now `Alter` basically just builds these expressions: https://github.com/JSQLParser/JSqlParser/compare/master...wrobstory:PR_alter-expressions?expand=1#diff-6ae5229b13d3feb0f9fdd13137ab69e4R3121
* All of the current `AlterTest` namespace passes and I added a couple more tests to ensure the multiple-expression parsing works.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/alter/Alter.java
+++ b/src/main/java/net/sf/jsqlparser/statement/alter/Alter.java
@@ -22,7 +22,8 @@
 package net.sf.jsqlparser.statement.alter;
 
 import java.util.ArrayList;
-import java.util.List;
+import java.util.Iterator;
+
 import net.sf.jsqlparser.schema.Table;
 import net.sf.jsqlparser.statement.Statement;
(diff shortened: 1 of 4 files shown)


wumpz:
thx for improving

PR #325, opened 2016-08-15
#modify net.sf.jsqlparser.statement.insert.Insert   Method:toString()…

superqtqt:
… if itemsList and useSelectBrackets together not null will error

#modify PlainSelect.getStringList change sql append to StringBuilder

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/select/PlainSelect.java
+++ b/src/main/java/net/sf/jsqlparser/statement/select/PlainSelect.java
@@ -407,25 +407,29 @@ public class PlainSelect implements SelectBody {
      * @return comma separated list of the elements in the list
      */
     public static String getStringList(List<?> list, boolean useComma, boolean useBrackets) {
-        String ans = "";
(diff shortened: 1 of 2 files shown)


src/main/java/net/sf/jsqlparser/statement/select/PlainSelect.java
```diff
@@ -407,25 +407,29 @@ public static String getStringList(List<?> list) {
...
         if (list != null) {
             if (useBrackets) {
-                ans += "(";
+                ans.append("(");
+//                ans += "(";
             }
 
             for (int i = 0; i < list.size(); i++) {
-                ans += "" + list.get(i) + ((i < list.size() - 1) ? comma + " " : "");
+                ans.append(list.get(i)).append(((i < list.size() - 1) ? comma + " " : ""));
```
wumpz: To construct valid comma lists the constructs are nearly always weird :). You could use a commalist generator method within Plainselect. Nevertheless it was not your statement but the older one I am refering to.

PR #328, opened 2016-08-25
Support additional Postgres column types for alter table statements

wrobstory:
Hi again JSqlParser maintainers!

The following types are valid column types for Postgres: 
```
time without time zone
time with time zone
timestamp without time zone
timestamp with time zone
bit varying
xml
interval
```

This PR adds support for those types in `ALTER TABLE` statements, along with tests ensuring that the types can be parsed. I believe that _all_ Postgres types will be supported once this PR is merged.


wrobstory:
This PR has been rebased on master- should be good for review + merge.

PR #365, opened 2016-12-03
support getting jdbc parameter index

hellojavaer:
support getting jdbc parameter index.
 - For sql 'select * from tab where id = ? ', support get the '?' index when traversing sql nodes.


wumpz:
Wouldn't it be better to use already existing adapters in your complete visitor to implement the standard traversal only once? Kind of delegator pattern?

hellojavaer:
oh, I just pay attention to  the class 'TablesNamesFinder' and I didn't notice the existing adapters before you say. In my project I need to traversal the full sql ast nodes to do some tables conversion and collect some column informations based on my input condition. In this case I can just override some methods of  'TablesNamesFinder' to implement my function. But the function of 'getTableList' isn't what I need for this case, So I provide a more basic class which based on  'TablesNamesFinder' to support traversing of the full ast nodes. I checked the existing adapters, and didn't find an adapter which implement the traversing of the full ast nodes. Maybe there is better a naming for 'JSQLBaseVisitor'. Or will there be any better ways to implement the traversing of the full ast nodes? how about you?

hellojavaer:
I reverted the second commit and updated the title. Please check it again.

wumpz:
Did you change Linefeeds? Reviewing it I have some much source blocks removed and added with no visible change at all. At least I did not find some.

hellojavaer:
Yes, there have been some source codes changed after I fork. Now I have synchronized to upstream repository and  re-implemented support getting jdbc parameter index. Please check it again.

PR #369, opened 2016-12-11
Add support for Apache Drill CSV column indices

davidhaller:
Apache Drill is able to perform queries like this on CSV files:

```sql
select columns[0], columns[1] from dfs.`/tmp/drill/test.csv`
```
JSqlParser could not parse column identifiers with an array index, so I modified the grammar. I'm not sure that I did everything right, as I never worked with JavaCC before. Tests still pass, and the debug output says that columns[0] and columns[1] were recognized as identifiers.


wumpz:
The problem I see is, that there is already some kind of array indexing. Additionally the array index will be part of an identifier, which is IMHO not correct.

davidhaller:
Actually this not really array indexing, as "columns" is not an array and not a valid identifier. Apache Drill just names columns from CSV files columns[0], columns[1] and so on which are usually renamed and casted to the desired data type, like this:

```sql
select cast(columns[0] as int) as user_id, columns[1] as username,
cast(columns[2] as timestamp) as registration_date
from `users.csv`
```

PostgreSQL, for example, has arrays which can be used within the "select" clause (https://www.postgresql.org/docs/9.4/static/arrays.html). If array support already exists in JSQLParser, what do I need to change to allow array access within "select"?

manticore-projects:
Obsolete, since it is working fine with JSQLParser 4.7. [You can test it online here.](http://jsqlformatter.manticore-projects.com/jsqlformatter/demo.html?args=-c%20MoUQMiDCAqAEsGMD2AbArgWwHYGcDaADALoBQ855ANIqprngIykBiASgPICysAJgGY4AdAAMA9ABcMAB1E8ATgEsUKCQFMc4wQhwA3YSQDcQA)

PR #381, opened 2017-01-12
fixes #288

### LINKED ISSUE #288: limit with jdbc index parameter not support
Hi ,
the parser can work with jdbc index param where the params are in where, but it can't works when I set jdbc index parameter in limit.
this works fine,
`select * from my_tbl where a = ? limit ?`
and also this:
`select * from my_tbl where a = ?1 limit ?`
while this throws a ParseException:
`select * from my_tbl where a = ?1 limit ?2`

jujusharp:
add support for limit with jdbc index parameter. e.g. limit ?1, ?2


wumpz:
Merged ... thx for improving JSqlParser

PR #382, opened 2017-01-13
support to add jdbc name parameter after LIMIT and TOP keywords

jujusharp:
This is a enhancement fix for bug #288, now **limit** can work with jdbc index parameter and jdbc name parameter. 
I also create some test cases for this.

     SELECT * FROM mytable WHERE mytable.col = 9 LIMIT :some_name, 2
     SELECT * FROM mytable WHERE mytable.col = 9 LIMIT 1, :some_name
     SELECT * FROM mytable WHERE mytable.col = 9 LIMIT :name1, :name2
     SELECT * FROM mytable WHERE mytable.col = 9 LIMIT ?1, :name1
     SELECT * FROM mytable WHERE mytable.col = 9 LIMIT :name1, ?1

and the same with **Top**

     SELECT TOP ?1 * FROM mytable WHERE mytable.col = 9
     select top :name1 foo from bar
     select top ? foo from bar

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/select/Limit.java
+++ b/src/main/java/net/sf/jsqlparser/statement/select/Limit.java
@@ -21,7 +21,7 @@
  */
 package net.sf.jsqlparser.statement.select;
 
-import net.sf.jsqlparser.expression.JdbcParameter;
+import net.sf.jsqlparser.expression.Expression;
 
 /**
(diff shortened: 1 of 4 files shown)


jujusharp:
I refactor `net.sf.jsqlparser.statement.select.Limit` and make rowCount and offset to `Expression` type, then the rowCountJdbcParameter has been removed. 
and now Limit has less methods. this make the code coverage decreased.

Issue #200, opened 2015-12-01
Could not parse 'case  when' expression in where expression of a select query.


aronica:
For example, parse the sql below will cause a JSQLParserException:
select \* from t1,t2 where case when t1.id=1 then t2.name='Marry' when t1.id=2 then t2.age=10 end

wumpz:
What should this Sql do? I don't think it is a valid one. Within a where clause there are conditions. Or is the some extension to case when to rewrite "a and b or c and d"?

-------- Ursprüngliche Nachricht --------
Von: aronica <notifications@github.com> 
Datum: 01.12.2015  04:56  (GMT+01:00) 
An: JSQLParser/JSqlParser <JSqlParser@noreply.github.com> 
Betreff: [JSqlParser] Could not parse 'case  when' expression in where
  expression of a select query. (#200) 

For example, parse the sql below will cause a JSQLParserException:

select * from t1,t2 where case when t1.id=1 then t2.name='Marry' when t1.id=2 then t2.age=10 end

aronica:
Both mysql/hive could accept this syntax.
Beside , i found case expression as a param of a function also cause problem.

wumpz:
Strange. You are right. But isn't boolean syntax much more readable? Or do I not understand the evaluation of this construct? Tobias

aronica:
Hope next version will add solution to fix this.

devender-yadav:
I am using JsqpParser in my application & got issue while parsing SQL with `case` statement. Its support should be added.

-Dev

wumpz:
@devender_yadav: case statements are supported a long long time. Or do you have the same more concrete problem discussed in this issue?

devender-yadav:
@wumpz 

Initially I though `CASE` is not supported. My bad..

I tired Query = 

    "SELECT Store_Name, CASE Store_Name WHEN 'abc','xyz' THEN Sales * 2  ELSE Sales END FROM Store_Information";

Got the exception:

    net.sf.jsqlparser.JSQLParserException
	at net.sf.jsqlparser.parser.CCJSqlParserUtil.parse(CCJSqlParserUtil.java:51)
	at com.dev.App.getTableNamesFromQuery(App.java:20)
	at com.dev.App.main(App.java:83)
    Caused by: net.sf.jsqlparser.parser.ParseException: Encountered " "CASE" "CASE "" at line 1, column 20.
    Was expecting one of:
    "*" ...
    "*" ...
    
[...]

PR #396, opened 2017-02-17
Potential solve for #200

JeroennC:
Hi,

I've changed around the grammar a bit in order to support CASE WHENs being used as Conditions on their own (see #200). This means that the THENs and ELSEs have to be able to contain full expressions. One of the problems that made this not work was that Condition() would be okay with turning into a Column or a 0 or 1. I've removed this and added Conditions becoming SimpleExpressions. This does however mean that it will accept any SimpleExpression in place of a condition, for instance:
```
SELECT * FROM t1 WHERE 5
```
or
```
SELECT * FROM t1 WHERE a + b
```

Which is also why I removed one of the tests what ensure this is not parse-able. I would argue that while it may indeed not be valid SQL (depending on the DB's implementation), it is no harm if it is actually parse-able. What do you think of this?


sh-tomer:
@JeroennC, Thanks for providing this fix. I merged it to my own custom branch to see if it's working and it worked perfectly.
@wumpz - did you consider pulling this request?

wumpz:
already solved

Issue #374, opened 2016-12-23
support "force index"


sealedbook:
e.g
SELECT 
    tel.*, ti.task_type
FROM
    HTtask.task_info AS ti,
    HTtask.task_exercise_log AS tel FORCE INDEX (IDX_T_E_L_MODIFY_DATE)
WHERE
    ti.id = tel.task_info_id

dhlin:
Will `force/ignore/use index` key words be supported in the future?

PR #429, opened 2017-04-18
Introduce support for mysql index hints (fixing issue #374)

joeymart:
Introduce support for MySQL style index hints (defined in mysql docs [here](https://dev.mysql.com/doc/refman/5.7/en/index-hints.html) ).

Index hints immediately follow the table and optional correlation alias in the target table and join clauses like:
`SELECT t.field FROM table t USE INDEX (index_name) WHERE ...`
...or in joins like:
`SELECT t2.field FROM table t INNER JOIN table_2 t2 IGNORE INDEX (index_name) WHERE ...`

This commit checks for the optionally present index hint as part of the `Table` definition, and I tried to make it as small and minimally invasive as possible.

This fixes issue #374 (which was specifically logged for `FORCE INDEX`) and handles the other flavors of mysql index hinting.

Thanks!

### DIFF
--- a/src/main/java/net/sf/jsqlparser/expression/MySQLIndexHint.java
+++ b/src/main/java/net/sf/jsqlparser/expression/MySQLIndexHint.java
@@ -25,27 +25,28 @@ import java.util.List;
 
 public class MySQLIndexHint  {
 
-	private final String action;
-	private final String indexQualifier;
-	private final List<String> indexNames;
+    private final String action;
+    private final String indexQualifier;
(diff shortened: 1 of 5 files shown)


joeymart:
I see there's a build error on Travis. Looks to be a checkstyle violation.
I'm fixing now...

joeymart:
Ok, fixed the checkstyle errors.
Had some tabs where spaces were expected and didn't have curly braces on a single-line if statement.

PR #433, opened 2017-04-24
Add new feature: convert to CNF form

messfish:
Greetings, Here is my thought about adding a new feature:

Basically I get the inspiration from this article: 
https://www.google.com/patents/US6353821

So I wonder whether I could do an optional function, that is by converting the expression INTO CNF form for optimization.  In a nutshell it will only changes the condition pattern of the expression so it could quite fit with the following updates.

I wrote the algorithm for the conversion at the definition of the class and made some test cases. Please take a look and check it to see whether my approach makes sense.


wumpz:
First of all, I think it would be good to introduce something like JSqlParser-Utils project. I don't think it fits well into the main library, does it?

src/main/java/net/sf/jsqlparser/expression/operators/conditional/MultiAndExpression.java
```diff
@@ -0,0 +1,53 @@
...
+
+import net.sf.jsqlparser.expression.Expression;
+import net.sf.jsqlparser.expression.ExpressionVisitor;
+
+/**
+ * This helper class is mainly used for handling the CNF conversion.
+ * @author messfish
+ *
+ */
+public class MultiAndExpression extends MultipleExpression {
```
wumpz: Are these helpers realy needed? Isnt the whole computation possible using the binary form or am I missing something?

src/main/java/net/sf/jsqlparser/statement/select/CNFConverter.java
```diff
@@ -0,0 +1,614 @@
...
+ *
+ */
+public class CNFConverter {
+
+    private Expression root; 
+    // the variable that stores the newly generated root.
+    private Expression dummy;
+    // this variable mainly serves as the dummy root of the true root.
+    // generally it will be a multi and operator with root as the child.
+    private Expression temp1;
```
wumpz: Due to this variables your class will not be threadsafe. You should prevent double usage, shoudnt you?

messfish:
Thanks for the feed back, here is what I thought:
(1) So does this mean I should put everything in a new package named "utils"?
(2) It is because the algorithm I wrote in the definition of the class has a step that will merge all the adjacent same condition expressions together. I cannot say it is impossible but it will be much harder to implement. The fourth step needs to push one and expression (which is the child of the or expression) up, which could be handled much easier if I could use an expression that could hold multiple elements. Besides, Everything will move back to binary form at last. After all, a CNF form is one and element that has multiple or elements which has multiple children expression.
(3) So is it a good solution to add a lock in the convert() method? If not, what do you think is a good solution to prevent double usage? Put the variables in the convert() method instead of global variable?
(4) Are there any problems on the test cases or the algorithm I use?

wumpz:
(1) Yes. 
Maybe even a separate project. I am not convinced to put this into the main parser jar. But I may be wrong. Unsure ...
(2) I did not dig very deep into the algorithm. So probably my not yet complete understanding.
(3) Do you want to reuse the CNFConverter? Instance creation is not a performance issue, its a small class. I would rigorously prevent this by setting a class used flag and then throwing an IllegalState at reuse. But its a matter of taste.
(4) I find your tests very hard to read. Its a lot of stuff. Do you know the util CCJSqlParserUtil.parseExpression. This should at least improve your step one tests. As you are comparing the resulting Sql Strings, I at the moment see no benefit of building all these in between versions, that result in the same String. 

Still this is all derived from only reading over your sources.

By the way, did you have any problems using the coding conventions by checkstyle I integrated recently?

messfish:
Ok. So I think I will create a package in the jsqlparser and put the code and test cases in it. 
In essence this work is mainly for optimization. 

For the third step, does that mean I declare a condition variable and if the condition changed, the class should throw an illegal argument exception?

For the test case, is it sounds fine if I only keep the final result of those test cases and leave out those intermediate steps?

messfish:
Oops... sorry for missing the last question. I do not experience any hard times on the style checking.

wumpz:
Regarding style checking .. I already got the first complaints about it. It is like this with every restriction. 
But I see it as a massive improvement regarding git history, merge checking and cleaness.

wumpz:
(3) Yes, why not? Again the question, would you like to reuse. But rethinking it a better option would be some kind of builder pattern: 

    CNFTransformer.transform(expression)   returning the corrected expression

The internals of CNFTransfomer should be hidden from the user. Why are the intermediate methods public?

For the tests: I must checkout a running example. Only from reading it, its difficult to tell, if only the first and last step should be tested.

PR #434, opened 2017-04-29
conversion to CNF

messfish:
Greetings, I follow your advice and do the major changes:
(1) I put every modifications in the util.CNFexpression package.
(2) I add a variable in the class to make sure it cannot be reused.
(3) For the test cases, I put them in an SQL statement instead of building expression trees.

Please let me know if there are still some improvements in the code.


src/test/java/net/sf/jsqlparser/util/CNFexpression/CNFTest.java
```diff
@@ -0,0 +1,171 @@
...
+        cnf5 = new CNFConverter();
+    }
+    
+    /**
+     * this method is used to test the full progress from the original
+     * expression tree to the expression tree in CNF form. In other words,
+     * this is used to test the last case.
+     */
+    @Test
+    public void testAll() {
```
AnEmortalKid: could we split these into individual tests? I think if we had that, we could have clearer tests with a structure:
```
// Tests CNF convertion with a NOT ( a AND b)
@Test
public void test simpleNotAnd{

String statement = "SELECT * from S WHERE NOT (1.2 < 2.3 AND 3.5 = 4.6) "
Select untransformed = (Select) JSqlTestUtil.parse(statement);

String expected = "SELECT * from S WHERE OR ( NOT ( 1.2 > 2.3 ) AND 3.5 != 4.6)" 
Select transformed = CNFConverter.convert(untransformed);
assertEquals(expected, transformed.toString()); // or even cnf.getRoot().toString();
```

However from your current tests (without looking at StepLastHelper), I have no idea what the actual expected value is, I know what your input is since I can track buildOriginalExpression1/2/3/4. However, I don't see why you can't write the test method so that it calls buildOriginalExpression1() and stores it in a local variable instead of storing it as class field. 

just my 2 cents.

Also, your CNFConverter.java class has some good doc with some trees, it would be neat if the javadoc for those tests had trees like that with the current input and the expected CNF tree.

wumpz:
@AnEmortalKid @messfish : good point. Writing separate tests would also remove the need for some of those helper methods.

messfish:
I updated the test cases and include the expected trees in it. Please take a look.

src/main/java/net/sf/jsqlparser/util/CNFexpression/MultiAndExpression.java
```diff
@@ -0,0 +1,44 @@
...
+ * #L%
+ */
+package net.sf.jsqlparser.util.CNFexpression;
+
+import java.util.List;
+
+import net.sf.jsqlparser.expression.Expression;
+
+/**
+ * This helper class is mainly used for handling the CNF conversion.
```
AnEmortalKid: Maybe a better description could be "Groups expressions that should be ANDED together for CNF Conversion" (same with the MultiOrExpression

src/main/java/net/sf/jsqlparser/util/cnfexpression/CNFConverter.java
```diff
@@ -19,7 +19,7 @@
  * <http://www.gnu.org/licenses/lgpl-2.1.html>.
  * #L%
  */
-package net.sf.jsqlparser.util.CNFexpression;
+package net.sf.jsqlparser.util.cnfexpression;
```
AnEmortalKid: would you mind just adding a link to https://en.wikipedia.org/wiki/Conjunctive_normal_form in your doc where you say "the CNF Form." :) other than that I really like the javadoc that you have in the trees

wumpz:
I would merge your code. Do you mind if I use githubs squash and merge? If you would like to have control over your history, you should squash it by yourself.

messfish:
Sounds good, thanks for your approval! I don't mind if you use squash and merge since my history code is a little bit ugly...

Issue #438, opened 2017-05-04
National character ( N'') syntax not supported 


freepath-it:
Hi,
the following syntas is not parsable:
SELECT N'national character' FROM table;

the N prefix on string is meant to identify an unicode string. I didnt find a better reference then this one:
http://stackoverflow.com/questions/18521928/the-n-in-the-where-clause-of-sql

The syntax is supported, and in some case needed, for example union with a nvarchar and a literal string in the same position on two different sql of the union) both from oracle and postgresql (didnt test other databases).

Regards,
Silvano

freepath-it:
Hi,
i have attached a pull request for this issuse. The parser is now able to correclty parse the query, but the test fails. The fix is missing something, any idea/suggestion how to properly fix it ?

thomaszbz:
SQL Server (TSQL) also uses that (see #267).

PR #439, opened 2017-05-04
Potential fix for #438

freepath-it:
Potentianl fix for #438. The parser is able to parse and rebuilds the statement, but the test fails. I am missinng something, but as of now i dont know what.


AnEmortalKid:
I would have implemented a new key type. Like `K_UNICODE` (look at K_AS https://github.com/freepath-it/JSqlParser/blob/2b45a14fce05ae74d7e44f344667669fb6697777/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt#L117) that is just defined like ` K_UNICODE: ("n"|"N") <LITERAL>`

Then you'd have a UnicodeString token that just wraps the literals. 

I am not sure why your code didn't work as you expected it, but It feels a bit hacky that we're saying that a literal is also now a `N'blah'` instead of `'blah'`

freepath-it:
I dont understand your proposal, but i believe its because i have no experieence at all with grammars, so please excuse the total ignorance...

The K_AS you use as an example defines a new token, "AS", but its not what is needed here: i understand your suggestion of introducing a new type, but shouldnt it be more similar to S_CHAR_LITERAL ? like S_UNICODE_CHAR_LITERAL: <"n"|"N"> <S_CHAR_LITERAL>.
But then, everywhere S_CHAR_LITERAL is referenced in the other definitions, it should be changed to (<S_CHAR_LITERAL>|<S_UNICODE_CHAR_LITERAL>).
Is it what you meant ? or is there a better solution to avoid rewriting a lot of definitions ?

wumpz:
Solved

Issue #446, opened 2017-05-14
Use provided ExpressionVisitor when de-parsing execute statements


chrycheng:
As [mentioned](https://github.com/JSQLParser/JSqlParser/issues/442#issuecomment-300189618) in issue #442, `ExecuteDeParser` takes an `ExpressionVisitor` but currently doesn't use it to de-parse `Expression`s.

PR #447, opened 2017-05-14
Fix issue #446

chrycheng:
Fixes issue #446.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -37,6 +37,12 @@
             <version>4.12</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>2.7.22</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <developers>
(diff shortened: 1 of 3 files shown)


chrycheng:
My pleasure!  BTW, I think the Codacy rule checking for asserts in the tests should be updated to consider Mockito method invocation verifications as asserts.

AnEmortalKid:
+1, I think we should update the codacy rules to include mockito verify

wumpz:
Have to lock into codacy. Till now I did no configuration regarding issue types. Any clue how to do it but in addition to the assert fail check?

chrycheng:
Sorry, no.  Not familiar with Codacy too.

AnEmortalKid:
Let me ask at work, a friend worked with some codacy.

AnEmortalKid:
So, coworker said no :(

However, I dug a bit into codacy and I believe it is using PMD (which is where that rule comes in). 

1. We could disable that PMD rule (bad idea).
2. We could try adding `@SuppressWarnings("PMD.JUnitTestShouldIncludeAssert")` on the tests that need verifies. (could work).
3. We could combine the warning suppression and one of us could write a custom PMD rule to check that there's a mockito.verify if that rule is excluded (i looked a bit into rules, it would take some time)

wumpz:
It seems to be an open issue for pmd at least for Mockitos verify.

Issue #442, opened 2017-05-07
Allow an ExpressionDeParser instance to be injected into StatementDeParser


chrycheng:
Is there a reason why, unlike other `DeParser` s like `InsertDeParser` and `SelectDeParser`, you aren't allowed to inject an `ExpressionDeParser` instance?  I could see this being useful, for example, when you want to create a translator from one SQL dialect to another.  You would code the translation of, for example, string concatenation from `||` to `concat(...)` in the `visit(Concat)` method of an `ExpressionDeParser`.  Then in order to apply this translation to all statements where it could occur, you would use the proposed enhanced `StatementDeParser` and then pass your `ExpressionDeParser` instance.  This appears to be a straightforward enough change so I think I can manage a pull request if approved.

wumpz:
You know,  JSqlParser is feature request driven. That's why it is not there. I can follow your argumentation. Looking forward to your pr.

chrycheng:
Of course.  I wanted to create an issue first so we could take about it if necessary.  Working on the PR now.

chrycheng:
Would it be acceptable to you if I picked only one `ExpressionDeParser` method to verify to have been called in the test?  Also, would it be OK if I use [Mockito](http://site.mockito.org/) in the test?

chrycheng:
Sorry about that.  Clicked the wrong button.  It seems I don't have privileges to assign myself the issue.  Could you do that for me please?

wumpz:
Assigned you ... , I have no problem including some kind of mocking framework, as long as it is only a test dependency.
I am not quit sure if I understand your question about the method called once. Could you make an example, what do you want to test.

chrycheng:
Of course.  I will make sure that the Mockito dependency is added only within the test scope.

Please refer to [`StatementDeParserTest`](https://github.com/chrycheng/JSqlParser/blob/30ef321710ce124c480eb66332233662c5e0c332/src/test/java/net/sf/jsqlparser/util/deparser/StatementDeParserTest.java) for the following paragraphs.

The `shouldUseProvidedExpressionDeparserWhenDeParsingInsert` test is what I meant in my previous comment.  I give a concrete SQL statement as input and then verify that the `ExpressionDeParser` is interacted with as expected.  However, I only checked the interaction through one `visit` overload.  I think it would be too cumbersome to test for all the `visit` overloads in this style.

[...]

PR #451, opened 2017-05-17
Fix issue #442

chrycheng:
Fixes issue #442.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,6 +38,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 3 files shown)


chrycheng:
Fixed the only issue in the first Codacy check that could be fixed.  Remaining ones are the known PMD issues with Mockito.

wumpz:
Agreed to //nopmd.

chrycheng:
I chose to annotate instead in order to scope down the suppression to that particular rule only.

Issue #452, opened 2017-05-17
Use the ExpressionDeParser to de-parse AnalyticExpressions


chrycheng:
Revision 066199c of `ExpressionDeParser` de-parses `AnalyticExpression`s by simply taking the value of the `toString` method.  Shouldn't the `ExpressionDeParser` be made to visit the `Expression`s that make up the `AnalyticExpression`?

I'd like to take a crack at this so please go ahead and assign it to me if you approve.

wumpz:
You are digging deep 👍 .

chrycheng:
Fixed by 81de7a6.

wumpz:
So are these github squashed commits ok for you?

chrycheng:
Actually, I personally prefer `--no-ff` merges but it's your repo so it's your rules 😃

wumpz:
The question was about the squashing. Some developers prefer to squash by themselfs. And I prefer one commit PRs. So I use happily githubs Squash & Merge.

chrycheng:
I see.  No problem!  I'll squash all subsequent PRs for this repo.

PR #453, opened 2017-05-18
Fix issue #452

chrycheng:
Fixes issue #452.  I basically moved the logic of `AnalyticExpression`'s `toString` to `visit(AnalyticExpression)` in `ExpressionDeParser` and replaced most calls to `toString` on the components to calls to `accept`.  I also introduced a few members that I think were required in order to be able to test the fix.  I limited the access to the package level to stress the fact that they are internal-only.  I hope you don't mind I also organized the imports on the existing classes.


wumpz:
@chrycheng The separation of deparsers and toString was intentionally. If one like to test a deparser then you have to construct the sql without it. The standard form was not done only within tests but within toString. Please leave this intact. 
For the replacement of toString call within the ExpressionDeparser you are right. That should be replaced.

chrycheng:
Yeah I had a feeling that might be an issue :)  It's a good thing I left it till the end in its own commit.  Easily reverted.  I felt that I had to [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) out the code but I guess in this case they are two very different cases.

chrycheng:
I disagree but it appears that PMD dislikes package-private members.  If you want, we could just expose those setters as public members and add corresponding public getters.  This seems to be the pattern for other `DeParser`s anyway.

wumpz:
I removed this pmd check.

PR #459, opened 2017-06-05
Add upsert grammar

messfish:
Greetings. Right now I am interestead in reading documents related to phoneix SQL so I want to see what I could contribute from it. It seems there is no "upsert" involved in the parser. So I write the 
implementation of "Upsert" and several basic test cases.


wumpz:
Could you commit your changes at the last master version. You mixed some cnf regarding stuff that was already merged using sqash and merge with new stuff. The squash creates a new commit that is not identical with your original ones.

messfish:
Oops, my bad. I will try to fix it.

messfish:
Ok I commit the changes to the last master version. Please take a look.

PR #494, opened 2017-08-21
replace support multiple values once

packageOk:


wumpz:
Looks ok for me. 

I would use *Squash and merge*, if thats ok for you. But you need to actualize your current fork, due to a new commit is created for your changes. If you dont like that, please squash your pr into one commit.

PR #504, opened 2017-08-28
Implements #509

KyongSik-Yoon:
Parsing failed when sql contain MySql hint SQL_CALC_FOUND_ROWS so I added rule for parsing that. 
I referred following documents.
https://dev.mysql.com/doc/refman/5.7/en/select.html


src/test/java/net/sf/jsqlparser/expression/mysql/MySqlSqlCalcFoundRowsTest.java
```diff
@@ -0,0 +1,56 @@
...
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+
+/**
+ * @author sam
+ */
+public class MySqlSqlCalcFoundRowsTest {
+    @Test
+    public void testPossibleParsingWithSqlCalcFoundRowsHint() throws JSQLParserException {
+        AtomicReference<MySqlSqlCalcFoundRows> ref = new AtomicReference<MySqlSqlCalcFoundRows>();
```
AnEmortalKid: do you really need it to be an atomic reference? could you just assign it to a variable?

src/test/java/net/sf/jsqlparser/expression/mysql/SelectVisitorFactory.java
```diff
@@ -0,0 +1,54 @@
+/*
```
AnEmortalKid: could both of these live in just your test class? (otherwise the name might throw people off since it is quite generic)

src/main/java/net/sf/jsqlparser/expression/MySqlSqlCalcFoundRows.java
```diff
@@ -0,0 +1,33 @@
...
+ * 
+ * You should have received a copy of the GNU General Lesser Public
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ * #L%
+ */
+package net.sf.jsqlparser.expression;
+
+/**
+ * @author sam
```
AnEmortalKid: Could you add the reference link here? And that this object supports: `SQL_CALC_FOUND_ROWS` for the MySQL driver?

KyongSik-Yoon: Yes. SQL_CALC_FOUND_ROWS keyword is hint syntax only for MySQL driver.
You can find here. 
https://dev.mysql.com/doc/refman/5.7/en/select.html

src/test/java/net/sf/jsqlparser/expression/mysql/MySqlSqlCalcFoundRowsTest.java
KyongSik-Yoon: I will do that.

src/test/java/net/sf/jsqlparser/expression/mysql/SelectVisitorFactory.java
KyongSik-Yoon: It's created only for unit test.
I'll change the class name to containing means of test.

src/main/java/net/sf/jsqlparser/expression/MySqlSqlCalcFoundRows.java
AnEmortalKid: Mind adding this to the doc:
```
Supports the {@code SQL_CALC_FOUND_ROWS} keyword for <a href="https://dev.mysql.com/doc/refman/5.7/en/select.html">MySQL driver.</a>
```

Just so there's a reference in case future devs want to know more about what this does (then they don't have to track this PR down) :)

PR #507, opened 2017-08-28
Fixed #506

### LINKED ISSUE #506: TableNamesFinder does not process ExtractExpression
TableNamesFinder does detect tables names inside EXTRACT.
e.g `SELECT EXTRACT(EPOCH FROM table.column)`

yakout:


wumpz:
Could you add a test? Multiple commits will be squashed using githubs merge and squash. If you want full control over it, provide a single commit pr.

yakout:
I found it's hard to extract tables from EXTRACT expression without implementing Column visitor which will cause other tests to fail because of the tables alias names.
But indeed I also found that it's not really needed to implement **ExtractExpression** visitor, because AFAIK any tables inside **EXTRACT** will be inside the from clause.
I don't know why **JSQLparser** allows not having FROM clause in case there is a table name inside EXTRACT
e.g `SELECT EXTRACT(MONTH table.date)` is not valid in PostgreSQL

wumpz:
JSqlParser povides more or less a structured view on your SQL. It is not a database sql validator. Because its grammar is a merge of multiple database specialities, that is not possible.

In your example you do not know, if "table" is indeed a table or a complex variable using in some PL/Sql script.

PR #517, opened 2017-09-27
Fix for some cases where TableNamesFinder can't find tables

natzcam:
Some edge cases that TableNamesFinder fails. For your review...


wumpz:
Thx for improving JsqlParser 👍

PR #532, opened 2017-10-27
Linking structures to their AST nodes to have access to their positions

sh-tomer:
This far there were only 3 types of structures linked to their AST nodes. Now adding some more expressions and literals to their AST node to have access to their token's position in the query.
Why this can be useful? If you're using JSqlParser to eventually present a query that was modified or maybe show different parts of it in different colors. In that case, you'll have to know the exact position of each of the tokens.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/expression/AllComparisonExpression.java
+++ b/src/main/java/net/sf/jsqlparser/expression/AllComparisonExpression.java
@@ -21,9 +21,10 @@
  */
 package net.sf.jsqlparser.expression;
 
+import net.sf.jsqlparser.parser.ASTNodeAccessImpl;
 import net.sf.jsqlparser.statement.select.SubSelect;
 
-public class AllComparisonExpression implements Expression {
+public class AllComparisonExpression extends ASTNodeAccessImpl implements Expression {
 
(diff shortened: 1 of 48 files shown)


sh-tomer:
@wumpz, this code compiled properly couple of weeks ago. Any idea if something changed in the parser generation process? It seems that the code "SimpleNode jjtn000 = new SimpleNode ....." is not generated for some of the structures, and therefore the compilation errors.

wumpz:
I will look into it. Did you run any memory consumption tests? That was one point to discard a complete last three.

sh-tomer:
@wumpz, thank you for your comment.
Any recommended way to test for memory change in JSqlParser that were used this far?
I have a test set of ~2000 real world queries that I'm testing my changes with. I'm running 100 iterations of that set to make sure nothing changes over time and relatively high volume.
Running the tests on a machine with low resources (1 CPU & 1GB RAM), to make sure there is no excessive use of memory - if something works on that machine, it provides a good indication it will  work on strong production environment.
Please let me know your thoughts.

sh-tomer:
Will recreate pull requests based on master - had some issues inside my fork, sorry.

PR #534, opened 2017-10-29
Linking structures to their AST nodes to have access to their positions

sh-tomer:
This far there were only 3 types of structures linked to their AST nodes. Now adding some more expressions and literals to their AST node to have access to their token's position in the query.
Why this can be useful? If you're using JSqlParser to eventually present a query that was modified or maybe show different parts of it in different colors. In that case, you'll have to know the exact position of each of the tokens.

Regarding memory consumption - I have a test set of ~2000 real world queries that I'm testing my changes with. I'm running 100 iterations of that set to make sure nothing changes over time and relatively high volume.
Running the tests on a machine with low resources (1 CPU & 1GB RAM), to make sure there is no excessive use of memory - if something works on that machine, it provides a good indication it will work on strong production environment.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/expression/AllComparisonExpression.java
+++ b/src/main/java/net/sf/jsqlparser/expression/AllComparisonExpression.java
@@ -21,9 +21,10 @@
  */
 package net.sf.jsqlparser.expression;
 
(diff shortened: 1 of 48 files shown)


wumpz:
Agreed and merging right now ...

PR #539, opened 2017-10-31
Add support for "NOT LIKE ..." expressions

sh-tomer:
The parser is able to parse expressions such as "a NOT LIKE '%pattern%'", but is not able to parse expressions where the NOT part is before the entire expression.
For example: "NOT a LIKE '%pattern%'.

When parsing the latter, the error is:
Caused by: net.sf.jsqlparser.parser.ParseException: Encountered " "LIKE" "LIKE "" at line 1, column 32.
Was expecting one of: ...

The reason this is important is both because these syntaxes are both valid, and also because the deparser uses the second method.
Therefore, if you parse a query with the first type of expression, then deparse it and parse again, you'll get the same error.

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -2077,9 +2077,18 @@ Expression LikeExpression() :
     Expression rightExpression = null;
 }
 {
-    leftExpression=SimpleExpression()
-    [<K_NOT> { result.setNot(true); } ] ( <K_LIKE> | <K_ILIKE> { result.setCaseInsensitive(true); } ) rightExpression=SimpleExpression()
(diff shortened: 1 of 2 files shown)


wumpz:
Thx for adding this. I was surprised that this very common expression was not supported.

sh-tomer:
Thanks, glad to have the chance to contribute.

wumpz:
@sh-tomer I hope it is not a problem that I use always squash and merge. I see there are problems due to your branch ist not merged into the master branch, but a new squashed commit is created. So you have to be sure to start from master again.

sh-tomer:
@wumpz, don't see a problem with squash merges, feel free.
Not sure I understood the last part though. I started all branches from the master. Is there anything else you recommend?

wumpz:
@sh-tomer No that's all. I mentioned only to go back to master since there are still pull requests open from developers that tend to not to go to master and therefore I get a PR with thousands of changes but only one relevant commit. This type I will not merge, I cannot merge. So fine if you do not see a problem there.

PR #546, opened 2017-11-22
Allow identifiers such as aliases, to start with digits

sh-tomer:
Mysql allows identifiers to start with a digit, but JSqlParser's parser fails when parsing such aliases.
Documentation:
https://dev.mysql.com/doc/refman/5.7/en/identifiers.html
Search for "Identifiers may begin with a digit".

An example for a query that will fail before this change: SELECT a AS 1_a FROM tbl WHERE c = d


wumpz:
Isn't the valid aproach regarding sql standards to quote those names?

"1_col" or [1_col] or `1_col`

sh-tomer:
Thanks for the reply. It's the standard in some cases, but not in MySQL's case.
The reference I mentioned in the pull request description shows that MySQL allows identifiers to start with a digit, even without quotes around it.
Also, I tested it and MySQL runs a query like the following without any problem:
select inv_item_sk as 1_a from inventory limit 1000;

wumpz:
That would collide with e.g. postgresql 

`select 1e`

where it is a scientific number.

sh-tomer:
Bummer, it kind of blocks a way to make both cases work. Any thoughts on how to move forward?

PR #548, opened 2017-11-26
Allow single quotes around aliases in SELECT clause

sh-tomer:
MySQL supports single quotes around aliases (in addition to double quotes and backticks). They are only allowed in aliases in the SELECT clause. In other cases, only backticks are supported.

From the documentation:
_"The identifier quote character is the backtick (`).
"In the select list of a query, a quoted column alias can be specified using identifier or string quoting characters"._

https://dev.mysql.com/doc/refman/5.7/en/identifiers.html
The example in the documentation shows an alias with single quotes around it as valid - search for 'two' in the url above.

The following query is valid in MySQL, but JSqlParser currently fails when parsing it:
`SELECT mycolumn AS 'My Column Name' FROM mytable`


wumpz:
fixed in #583

PR #577, opened 2018-02-01
Add SUBSTRING(string FROM start_pos [FOR length])

artemsnisarenko:
Yet another pull request for ANSI SQL-92 substring. 
```
<character substring function> ::=
              SUBSTRING <left paren> <character value expression> FROM <start position>
                          [ FOR <string length> ] <right paren>
```
http://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt

### DIFF
--- a/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitor.java
+++ b/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitor.java
@@ -178,4 +178,5 @@ public interface ExpressionVisitor {
 
     public void visit(NotExpression aThis);
 
+    void visit(SubstringExpression substringExpression);
 }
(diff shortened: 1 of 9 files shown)


wumpz:
Isn't the a from b syntax used as parameter for other functions as well? Wouldn't it be better to improve the function rule to accept this kind of parameter?

artemsnisarenko:
In SQL:2003 there are four functions that have a `from` sub-expression. Besides `substring` they are `overlay`, `trim` and `extract`. The last one already has a dedicated expression class.  I'm not sure about merits of either approach but changing the function rule would be inconsistent with what's already there.

wumpz:
Oracle has extract as well (#579).

PR #581, opened 2018-02-06
Lomboked the code.

wheredevel:
As discussed at https://gitter.im/JSQLParser/JSqlParser on Feb 5 2018 @wheredevel -> @wumpz .
Here's the "lomboked" code of JSQLParser.  
All the 659 tests are green.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/select/AllColumns.java
+++ b/src/main/java/net/sf/jsqlparser/statement/select/AllColumns.java
@@ -28,8 +28,6 @@ import net.sf.jsqlparser.parser.ASTNodeAccessImpl;
  */
 public class AllColumns extends ASTNodeAccessImpl implements SelectItem {
 
-    public AllColumns() {}
-
     @Override
     public void accept(SelectItemVisitor selectItemVisitor) {
         selectItemVisitor.visit(this);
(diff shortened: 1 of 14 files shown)


wumpz:
The start of Lombok was fairly simple using Netbeans, which is my favourite IDE. The problems were comming while using it. That's why at the moment I have more objections in using it:

- Somehow stacktraces are corrupted:  sorry I don't know how to describe it better.
- Netbeans "test file" does not work stable
- Testng + Maven somehow instable (tests do not run every time) (without lombok no problems)
- refactoring is problematic, due to Netbeans does not see all usages
- no more "find usages" in Netbeans (dont know about eclipse)
- debugging getter not possible anymore

Here is a discussion of pros and cons I found useful. There are a lot more I know.

https://stackoverflow.com/questions/3852091/is-it-safe-to-use-project-lombok

wheredevel:
I can't comment on Netbeans vs. Lombok, because I never used it.  
In the mentioned Stackoverflow discussion there're some tips, like cleaning cache - maybe those can help.

I'm an Eclipse user, and I never encountered any problem, due to Lombok generated constructors and methods, with either stack traces, tests, Maven, finding reference and call hierarchies.   

I agree that refactoring the generated constructors and methods is not straight-forward. But I think, this is the flaw in the Eclipse's implementation of the refactoring. But I'm sure Lombok here is the one that corrects the inherent problem, here's an example:
Let's say I have a field named `count`. The generated boilerplate code would have `get/setCount()` methods. Now I decide to rename the field into `total` - there's no mechanism to force me to rename its getter and setter(!) Lombok enforces this by managing 3 references at 1 place. In turn, what Eclipse developers should do - is to add a checkbox (default) saying "modify getter/setter refernces as well".
The workaround to this: override getter/setter, rename 1.field, 2. getter 3. setter, than remove getter/setter - back to Lombok.  

Same with Debugging getter/setter: if you do not override them, why debugging them at the first place? Their implementation isn't interesting. And if you do override - you can step into, as usual.

Issue #584, opened 2018-02-20
Mysql Parser error 


rmahnovetsky:
I get this parser error in mysql
select chequeoute0_.cheque_book_id as cheque_b1_11_, chequeoute0_.cheque_number as cheque_n2_11_, chequeoute0_.created_utc as created_3_11_ from vw_cheque_out chequeoute0_ where (chequeoute0_.cheque_book_id, chequeoute0_.cheque_number)=(76, 1)

`Caused by: net.sf.jsqlparser.parser.ParseException: Encountered " "(" "( "" at line 1, column 179.
Was expecting one of:
    "ALL" ...
    "NOT" ...
    "NULL" ...
    "CASE" ...
    "INTERVAL" ...
    "GROUP_CONCAT" ...
    <K_TIME_KEY_EXPR> ...
    <S_CHAR_LITERAL> ...
    "{d" ...
    "{t" ...
    "{ts" ...
    "NULL" ...
    "CASE" ...
    "GROUP_CONCAT" ...
    <K_TIME_KEY_EXPR> ...
    <S_CHAR_LITERAL> ...
    "{d" ...
    "{t" ...
    "{ts" ...
    "INTERVAL" ...
    "NOT" ...
    "NULL" ...
    "CASE" ...
    "GROUP_CONCAT" ...
    <K_TIME_KEY_EXPR> ...
    <S_CHAR_LITERAL> ...
    "{d" ...
    "{t" ...
    "{ts" ...
    "INTERVAL" ...
    "NOT" ...
    "NULL" ...
    "CASE" ...
    "GROUP_CONCAT" ...
    <K_TIME_KEY_EXPR> ...
    <S_CHAR_LITERAL> ...
    "{d" ...
    "{t" ...
    "{ts" ...
    "INTERVAL" ...
    "NOT" ...
    "ALL" ...
    "NULL" ...
    "CASE" ...
    "GROUP_CONCAT" ...
    <K_TIME_KEY_EXPR> ...
    <S_CHAR_LITERAL> ...
    "{d" ...
    "{t" ...
    "{ts" ...
    "INTERVAL" ...
    "NULL" ...
    "CASE" ...
    "GROUP_CONCAT" ...
    <K_TIME_KEY_EXPR> ...
    <S_CHAR_LITERAL> ...
    "{d" ...
    "{t" ...
    "{ts" ...
    "INTERVAL" ...
    "NOT" ...
    "NULL" ...
    "CASE" ...
[...]

PR #585, opened 2018-02-26
JSQLPARSER-584: adds support for MySQL (a,b,...)OP(c,d,...) expression

adriil:
Fixes issue #584

### DIFF
--- a/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitor.java
+++ b/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitor.java
@@ -167,6 +167,8 @@ public interface ExpressionVisitor {
     void visit(KeepExpression aexpr);
 
     void visit(MySQLGroupConcat groupConcat);
+    
+    void visit(MySQLValueListExpression valueList);
 
     void visit(RowConstructor rowConstructor);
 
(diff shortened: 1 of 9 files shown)


adriil:
I'll add the missing tests as soon as possible.

adriil:
I realize that I could get rid of my new class, which is useless, actually, as it only contains an expressionList...
I will rework my PR, do you want me to add commit here once it's done, or to close this one and provide a new clean PR ?

wumpz:
Modify this pr. I will make s squash merge at the end, if you don't mind.

adriil:
Hi,
Sorry for the silence, I had busy days last week.

Actually I'm facing an issue.
In order to get rid of my class which just contains an expressionList, I tried to use directly the expressionList as a ComparisonItem, instead of my class. But It can't be done because the ExpressionList is not regarded as an Expression itself.
I felt dumb because I already faced this issue, this was the reason I've created a class implementing Expression to get through this.
A solution to me, without creating useless class, would be to make ExpressionList implementing Expression, like a Composite. But this may be a big step as almost everything relies on Expression and ExpressionList.

wumpz:
Hmm. Same here. Leave your class. Rename it to ValueListExpression and everything is fine, at least regarding this pull request.

PR #589, opened 2018-03-15
Fixed: column names in INSERT statement should be simple, not FQN.

wheredevel:


src/main/java/net/sf/jsqlparser/schema/Column.java
```diff
@@ -62,41 +62,55 @@ public void setColumnName(String string) {
...
     /**
-     * Get name with out without using aliases.
-     *
-     * @param aliases
+     * Get FQN.
+     * @param aliases - use aliases.
      * @return
      */
-    public String getName(boolean aliases) {
+    public String getFullyQualifiedName(boolean aliases) {
```
AnEmortalKid: I believe this would force us to do a major version bump right?

src/main/java/net/sf/jsqlparser/schema/Column.java
```diff
@@ -62,41 +62,55 @@ public void setColumnName(String string) {
...
         }
+
         return fqn.toString();
     }
 
+    /**
+     * Get name.
+     * @return 
+     */
+    public String getName() {
```
AnEmortalKid: Maybe update doc with “@return the simple name of the column. “

wumpz:
I will not merge this, due to injecting a guava dependency for it. 

@AnEmortalKid Why do you think we should make a main version update? The renaming looks reasonable and for this minimal change it would be to much I think to introduce version 2.

AnEmortalKid:
@wumpz the method with signature `getName(boolean)` disappeared. Which is going to cause compilation errors for consumers. If you’re consuming it directly, that’s fine. However if you are pulling it in transitively in a war or assembly of sort, developers might not be wary of going to 1.2 and having things fail at runtime because the method disappeared. 


From semver.org

Given a version number MAJOR.MINOR.PATCH, increment the:

MAJOR version when you make incompatible API changes,
MINOR version when you add functionality in a backwards-compatible manner, and
PATCH version when you make backwards-compatible bug fixes.
Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.

wumpz:
@anemortalkid wenn already had this discussion some time ago. 😋

PR #594, opened 2018-03-25
Add support for MySQL SQL_NO_CACHE

sh-tomer:
Adding support for MySQL's SQL_NO_CACHE flag.
When the flag is used, the MySQL server does not use the query cache. It neither checks the query cache to see whether the result is already cached, nor does it cache the query result.

docs: https://dev.mysql.com/doc/refman/5.7/en/query-cache-in-select.html

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/select/PlainSelect.java
+++ b/src/main/java/net/sf/jsqlparser/statement/select/PlainSelect.java
@@ -60,6 +60,7 @@ public class PlainSelect extends ASTNodeAccessImpl implements SelectBody {
     private boolean useBrackets = false;
     private Wait wait;
     private boolean mySqlSqlCalcFoundRows = false;
+    private boolean sqlNoCacheFlag = false;
 
     public boolean isUseBrackets() {
         return useBrackets;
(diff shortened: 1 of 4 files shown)


src/main/java/net/sf/jsqlparser/statement/select/PlainSelect.java
```diff
@@ -465,7 +469,15 @@ public void setMySqlSqlCalcFoundRows(boolean mySqlCalcFoundRows) {
         this.mySqlSqlCalcFoundRows = mySqlCalcFoundRows;
     }
 
+    public void setMySqlSqlNoCache(boolean sqlNoCacheFlagSet) {
```
AnEmortalKid: Should the field name also be mySql like the other MySQL specific one? Or do other dbs provide a no cache mode

sh-tomer: Thanks for the review.
As far as I know, there are no other databases which support this flag. Seems like an extension of MySQL.

AnEmortalKid: Ah, then id name the field like the other MySQL one (makes it clearer in this class)

sh-tomer: I'm not sure I understood the suggestion. can you clarify please?

AnEmortalKid: Yeah, since this field is for mySQL support only, we should rename it to be `mySqlSqlNoCacheFlag` like the `mySqlSqlCalcRows` field (that way we can distinguish that this is MySQL specific.

sh-tomer:
@AnEmortalKid, done

AnEmortalKid:
@wumpz I'm happy with this, are you

PR #629, opened 2018-06-20
Refactor to make the code more readable and testable

A4Vision:
Refactor the class `AnalyticExpression`.
Extract the "partition by" and "order by" logics out of `AnalyticExpression` into two new classes, in order to make the code more testable and more readable.


wumpz:
Agreed ...

PR #631, opened 2018-07-01
IN clause without parentheses is not working but work in Oracle DB

skylark-nam:
Hi, Oracle expressions are possible.  expression IN value1

ex) select 1 from dual where a in 1


wumpz:
I have two problems with this commit:

1. something like "in ( expr"  would be accepted. If there is an opening parenthesis there must be a closing one. I know it is not yet done for all of JSqlParser, but it should.
2. The deparser should reconstruct exactly the same statement from the parse tree. But without any change, it would forget the parenthesis.

manticore-projects:
Obsolete now, since I accidentally implemented this somewhere down the 4.7 Snapshot (reworking `ExpressionList`).
[You can test it here.](http://jsqlformatter.manticore-projects.com/jsqlformatter/demo.html?args=-c%20MoUQMiDCAqAECMAoAYgJQPIFlYBMCuAhgDaIDqAEiKiLAbAJIByCiA3EA)

PR #635, opened 2018-07-03
add support for && operator

sh-tomer:
MySQL allows to use the `&&` operator as an alias to the AND operator.
https://dev.mysql.com/doc/refman/8.0/en/logical-operators.html#operator_and

Example:
`SELECT col FROM tbl WHERE tbl.pid = 2 && tbl.status = 1`


wumpz:
Again, the deparser should reconstruct the original statement, shouldn't it.

wumpz:
Changed my mind. I think you are right about the deparsers.

PR #640, opened 2018-07-11
fix parsing and rendering of Truncate

ted-johnson:
- the method for rendering a Truncate command in StatementDeParser.java is empty.  This can cause problems, e.g. I wrote a program to split a file of SQL statements into individual statements and the TRUNCATE commands were missing.  I filled in the module.

- The parser does not accept a CASCADE directive in a TRUNCATE DDL.  I modified the parser to accept CASCADE, added code to Truncate.java to record the CASCADE directive and added code to the toString method to render the CASCADE if present.

example SQL (postgres):
   truncate table global_uids CASCADE;

The modified code passes all test


ted-johnson:
I added a test of TRUNCATE deparsing
and addressed the minor style issue.

PR #641, opened 2018-07-11
support teradata shortcut for Select

ted-johnson:
Teradata SQL allows the user to use "sel" in place of "select".
Example:
   sel name, status from accounts

I modified JSqlParserCC.jjt to accept the "sel" keyword as synonymous with Select.

The modified code passes all tests


ted-johnson:
I have added a test, the code passes.

I disagree that the deparser should reconstruct the SEL.
SEL is a non-standard shorthand for SELECT.
Replacing SEL with SELECT creates the exact same parse (there is no information loss),
and has the benefit of being more standard.

wumpz:
Agreed

Issue #694, opened 2018-10-12
Round tripping of "NOT LIKE" where clause element does not work


thebiguno:
### Actual Behavior
Parse the statement:
```sql
SELECT VALUE1, VALUE2 FROM FOO WHERE BAR NOT LIKE '*%'
```
This parses properly.

When you call toString on the parsed statement, it outputs:
```sql
SELECT VALUE1, VALUE2 FROM FOO WHERE NOT BAR NOT LIKE '*%'
```

Note the 'NOT' in front of BAR.  This is invalid DB syntax.

### Expected Behavior

The output should not include a spurious NOT.

### Code Changes
The issue is from BinaryExpression line 71.  That line checks whether it is a 'not' statement and adds the NOT:
```java
    public String toString() {
        return (not ? "NOT " : "") + getLeftExpression() + " " + getStringExpression() + " " + getRightExpression();
    }
```
However the NOT is also added in LikeExpression line 49 in getStringExpression() (which is called from the BinaryExpression toString() method):
```java
    public String getStringExpression() {
        return (isNot() ? "NOT " : "") + (caseInsensitive ? "ILIKE" : "LIKE");
    }
```

I have fixed this on a fork and will submit a pull request shortly.

  - Version: 1.4 latest (pulled right now)

thebiguno:
Pull request:
https://github.com/JSQLParser/JSqlParser/pull/695

tmortagne:
Is there any plan for a 1.4 or 1.3.1 release in a short time ? This bug is quite important for XWiki but we would like so is #163 so no a big fan of going back to 1.2 :)

Is there anything I could do to help ?

wumpz:
@tmortagne  #731

tmortagne:
Thanks for considering it :)

PR #695, opened 2018-10-12
Fix to issue #694 (spurious NOT in NOT LIKE elements within the WITH clause)

thebiguno:
Removes a spurious NOT from the BinaryExpression toString() element.  All the Maven tests pass.


wumpz:
I changed your PR a bit. You are removing nots from non like statements as well.

thebiguno:
Awesome, thanks!

Issue #555, opened 2017-12-09
No support the function of "IF"


ghost:
No support the function of "IF”, For example, 
SELECT
     t.req_url AS url,
     COUNT(0) AS total,
     **IF(t.http_code = 200, 1, 0) AS normal**
     FROM
	 test_table t
     WHERE
          t.req_start_time BETWEEN 0 and 1612097162291
     GROUP BY t.req_url

ShipAlex:
Yes, it is

CodingFabian:
```
net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: "IF" "IF"
    at line 4, column 1.

Was expecting one of:

    "*"
    "+"
    "-"
    "?"
    "@"
    "@@"
    "ACTION"
    "ANY"
    "CASCADE"
    "CASE"
    "CAST"
    "COLUMN"
    "COMMIT"
    "DO"
    "ENABLE"
    "END"
    "EXTRACT"
    "FIRST"
    "FOLLOWING"
    "GROUP_CONCAT"
    "INDEX"
    "INSERT"
    "INTERVAL"
    "KEY"
    "LAST"
    "MATERIALIZED"
    "NO"
    "NULL"
    "NULLS"
    "OPEN"
    "OVER"
    "PARTITION"
    "PERCENT"
    "PRECISION"
    "PRIMARY"
    "PRIOR"
    "RANGE"
    "REPLACE"
    "ROW"
    "ROWS"
    "SEPARATOR"
    "SIBLINGS"
    "TABLE"
    "TEMP"
    "TEMPORARY"
    "TRUNCATE"
    "UNSIGNED"
    "VALUE"
    "VALUES"
    "XML"
    "{d"
    "{t"
    "{ts"
    "~"
    <K_DATETIMELITERAL>
    <K_TIME_KEY_EXPR>
    <S_CHAR_LITERAL>
    <S_DOUBLE>
    <S_HEX>
    <S_IDENTIFIER>
    <S_LONG>
    <S_QUOTED_IDENTIFIER>
```
"IF" is not yet supported. We also would like to have it

wumpz:
I know, its a long requested feature. Maybe one could extend function parameters to conditional expressions. Then "if" would be accepted as well. 

Pull requests are welcome. At the moment it is problematic to work steadily on JSqlParser unfortunately.

Issue #511, opened 2017-09-04
function fail to parse


packageOk:
Hi, when I parse "update food set num=if(num > 1, 1, 0);", it throw a exception. but if I try "update food set num = if(1, 1, 0);", it works well. Is there anything wrong?

wumpz:
JSqlParser does not (yet) support conditions as function parameters.

zhztheplayer:
@wumpz 
I also wonder if the boolean expr parameter (or as select items) could be supported in later version of Jsqlparser.

As I known, some popular storages like postgreSQL supports this feature, and SQL-2003 standard explicitly declares this also.

wumpz:
@zhztheplayer 
JSqlParser will support this. It is already a requested feature. But til now I had no plan, when to start.

tserr:
@wumpz 
I'd like to see this feature implemented ;-)
Did you start implementing in the meantime? Is there already some code, I could base on to help you implementing this feature (in the bounds of my possibilities). Should I start a try on my own?

wumpz:
Did not start implementing in this direction. Sorry.

wumpz:
now supported for specific function names (e.g. if)

Issue #301, opened 2016-06-29
Failed to parse properly formatted MSQL syntax.


gir489:
I am trying to parse the following SQL select with the SQL parser:

`String stmt = "SELECT CAST((MAX(CAST(IIF(isnumeric(license_no) = 1, license_no, 0) AS INT )) + 2) as varchar) FROM lcps.t_license WHERE profession_id = 60 and license_type = 100 and YEAR(issue_date) % 2 = case when YEAR(issue_date) % 2 = 0 then 0 else 1 end and ISNUMERIC(license_no) = 1";
Statements stmts = CCJSqlParserUtil.parseStatements(stmt);`

The exception is given:

`Caused by: net.sf.jsqlparser.parser.ParseException: Encountered " "(" "( "" at line 1, column 12.
Was expecting one of:
    <EOF> 
    "AS" ...
    "DO" ...
    "ANY" ...
    "KEY" ...
    "PERCENT" ...
    "INTO" ...
    "FROM" ...
    "OPEN" ...
    "WHERE" ...
    "FOR" ...
    "XML" ...
    "GROUP" ...
    "VALUE" ...
    "HAVING" ...
    "VALUES" ...
    "REPLACE" ...
    "TRUNCATE" ...
    "CAST" ...
    "OVER" ...
    "PARTITION" ...
    "EXTRACT" ...
    "MATERIALIZED" ...
    "START" ...
    "CONNECT" ...
    "PRIOR" ...
    "SIBLINGS" ...
    "COLUMN" ...
    "NULLS" ...
    "FIRST" ...
    "LAST" ...
    "ROWS" ...
    "RANGE" ...
    "FOLLOWING" ...
    "ROW" ...
    "SEPARATOR" ...
    "CASCADE" ...
    "NO" ...
    "ACTION" ...
    <S_IDENTIFIER> ...
    <S_QUOTED_IDENTIFIER> ...
    ";" ...
    "," ...
    "+" ...
    "||" ...
    "|" ...
    "&" ...
    "-" ...
    "^" ...
    "::" ...
    "::" ...
    "^" ...
    "+" ...
    "-" ...
    "||" ...
    "|" ...
    "&" ...
    "AS" ...
    <S_IDENTIFIER> ...
    <S_QUOTED_IDENTIFIER> ...
    "CAST" ...
    "DO" ...
    "EXTRACT" ...
    "FIRST" ...
    "FOLLOWING" ...
    "LAST" ...
[...]

PR #697, opened 2018-10-16
Function boolean param

dfulachier:
Hello,
First congratulation for this simple-to-use and efficient API. This is a very powerfull tool.
I faced to the "famous" issue with boolean expression in function call (typically: "if(a<b, c, d)").
It does not seem to be so easy to fix, due to grammar complexity. I faced issue with the "ALL" optional clause we can found in both condition and function call parsing branches.
I propose a first quick fix, taking into account only simple boolean expressions (<, >, etc. simple operators.). 
Modified JavaCC grammar, plus a unitary test.

This fix addresses at least the following tickets: #301, #511, #555.

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -2329,6 +2329,20 @@ ExpressionList SimpleExpressionList():
     }
 }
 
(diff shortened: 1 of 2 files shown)


wumpz:
Have to check. Thx for using Jsqlparser.

wumpz:
I am on a modification to allow conditions as function parameters. This would solve this issue not only for if statements. Give me some time.

wumpz:
I implemented a different approach but included your test file.

PR #702, opened 2018-10-23
support named parameters

ted-johnson:
Some standard SQL functions, especially string functions, use keywords to separate parameters.
I have encountered 
For example,
SELECT 
  substring(id, 2, 3),
  substring(id from 2 for 3),
  substring(id from 2),
  trim(BOTH ' ' from 'foo bar '),
  trim(LEADING ' ' from 'foo bar '),
  trim(TRAILING ' ' from 'foo bar '),
  trim(' ' from 'foo bar '),
  position('foo' in 'bar'),
  overlay('foo' placing 'bar' from 1),
  overlay('foo' placing 'bar' from 1 for 2)
FROM mytable

This pull request adds support for these functions.
I have added a test for these named parameters in SelectTest.java
The code passes all tests.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitorAdapter.java
+++ b/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitorAdapter.java
@@ -347,6 +347,14 @@ public class ExpressionVisitorAdapter implements ExpressionVisitor, ItemsListVis
     }
 
     @Override
+    public void visit(NamedExpressionList namedExpressionList) {
+        for (Expression expr : namedExpressionList.getExpressions()) {
+            expr.accept(this);
+        }
+    }
+
+
+    @Override
     public void visit(MultiExpressionList multiExprList) {
         for (ExpressionList list : multiExprList.getExprList()) {
             visit(list);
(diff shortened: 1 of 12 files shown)


wumpz:
Great job. Never thought about a solution like your named parameters.

manticore-projects:
Dear All,

unfortunately I have been able to isolate this particular PR as the root cause of a massive(!) performance deterioration. The problematic part seems to be
```
LOOKAHEAD(NamedExpressionListExprFirst()) namedExpressionList = NamedExpressionListExprFirst()
```
in the Fragment `Function InternalFunction(Function retval)`

Just to illustrate my concern: If you remove this part (sacrificing the function with named parameters), then the nested function calls become light speed fast:
```
// With NamedExpressionListExprFirst() compiled in
 doIncreaseOfParseTimeTesting("concat($1,'B')", "'A'", 20); 

--> The test passed. (0.205 s)

// With NamedExpressionListExprFirst() compiled in
 doIncreaseOfParseTimeTesting("concat($1,'B')", "'A'", 20); 

--> The test passed. (30.676 s)
```

If I am not mistaken, the PR makes this particular Test running 600 times slower. That is an issue at least to me.

So how can we get the cow off the ice together?
I would like to suggest to rewrite the Grammar for the NamedExpressionListExprFirst() fragment and narrow it down to the actually allowed Functions with named parameters. Any better suggestions?

Unfortunately I am "a Oracle guy" an not experienced with these kind of Functions/Expressions, so can someone please shared a complete List of Functions that would be supported?

manticore-projects:
I have worked around that with PR #1200 at the cost that Function With Named Parameters are only supported when the functions are not nested deeper than a certain threshold (currently defined as 7). Although it would not parse in this case anyways because of the slowdown.

PR #709, opened 2018-11-07
Support Alter Table Drop Constraint If Exists

rfscholte:
Support parsing `ALTER TABLE Persons DROP CONSTRAINT IF EXISTS UC_Person` as allowed by PostgreSQL

### DIFF
--- a/src/test/java/net/sf/jsqlparser/statement/alter/AlterTest.java
+++ b/src/test/java/net/sf/jsqlparser/statement/alter/AlterTest.java
@@ -147,6 +147,14 @@ public class AlterTest {
     }
 
     @Test
+    public void testAlterTableDropConstraintIfExists() throws JSQLParserException {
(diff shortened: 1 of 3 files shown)


src/test/java/net/sf/jsqlparser/statement/alter/AlterTest.java
```diff
@@ -146,6 +146,14 @@ public void testAlterTableDropConstraint() throws JSQLParserException {
         assertEquals(alterExpression.getConstraintName(), "YYY");
     }
 
+    @Test
+    public void testAlterTableDropConstraintIfExists() throws JSQLParserException {
+        Statement stmt = CCJSqlParserUtil.parse("ALTER TABLE Persons DROP CONSTRAINT IF EXISTS UC_Person");
+
+        // not yet interesting enough to remember 'IF EXISTS' flag on constraint
```
wumpz: The deparser should handle this. Therefore it should remember it.

rfscholte: So you want an ifExists-flag on the constraint as well?

PR #722, opened 2018-12-11
Support KSQL's WITHIN

lmontrieux:
[KSQL](https://github.com/confluentinc/ksql) supports the 'WITHIN' keyword for stream-stream joins. The syntax is as follows:

```
WITHIN [(before TIMEUNIT, after TIMEUNIT) | N TIMEUNIT]
```

See the [syntax reference](https://docs.confluent.io/current/ksql/docs/developer-guide/syntax-reference.html#create-stream-as-select)

This PR adds support for the WITHIN statement inside a join.


rcillo:
@wumpz if a new SNAPSHOT has not yet been release, would you be so kind to release it?

wumpz:
@rcillo Sorry. Try to do it tonight.

PR #724, opened 2018-12-12
Support Alter Table Alter Column Type

rfscholte:


wumpz:
Thx for using Jsqlparser.

PR #414, opened 2017-03-24
Issue 391 expression for interval.

ryanmurf:

PR #725, opened 2018-12-14
Added support for expressions as arguments of an IntervalExpression. …

Sebarl:
…Based on #414


manticore-projects:
This has been implemented already and is supported now (at least in JSQLParser 4.6 Snapshot).

Issue #571, opened 2018-01-22
Error on table rename, column rename, add constraints with cascade


Golg013:
Hello, Wumpz!

**I understand that you are busy, so you can just ignore this message.**

I have some issues:

`CREATE TABLE table_qwert
(
    column_1 INT,
    column_2 INT,
    column_3 INT,
    CONSTRAINT table_qwert_db_object_id_fk FOREIGN KEY (column_1) REFERENCES db_object (id)
  ON UPDATE CASCADE
  ON DELETE CASCADE
);

Caused by: net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: "ON" "ON"
    at line 8, column 3.

Was expecting one of:

    ")"
    ","

	at net.sf.jsqlparser.parser.CCJSqlParser.generateParseException(CCJSqlParser.java:16995) ~[jsqlparser-1.2-20180105.233414-16.jar:na]
	at net.sf.jsqlparser.parser.CCJSqlParser.jj_consume_token(CCJSqlParser.java:16848) ~[jsqlparser-1.2-20180105.233414-16.jar:na]
	at net.sf.jsqlparser.parser.CCJSqlParser.CreateTable(CCJSqlParser.java:9451) ~[jsqlparser-1.2-20180105.233414-16.jar:na]
	at net.sf.jsqlparser.parser.CCJSqlParser.SingleStatement(CCJSqlParser.java:171) ~[jsqlparser-1.2-20180105.233414-16.jar:na]
	at net.sf.jsqlparser.parser.CCJSqlParser.Statements(CCJSqlParser.java:250) ~[jsqlparser-1.2-20180105.233414-16.jar:na]
	at net.sf.jsqlparser.parser.CCJSqlParserUtil.parseStatements(CCJSqlParserUtil.java:152) ~[jsqlparser-1.2-20180105.233414-16.jar:na]
	... 97 common frames omitted

##################################################################################

ALTER TABLE attachment CHANGE content test LONGBLOB;

Caused by: net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: "CHANGE" <S_IDENTIFIER>
    at line 1, column 24.

Was expecting one of:

    "ADD"
    "DROP"
    "MODIFY"

	at net.sf.jsqlparser.parser.CCJSqlParser.generateParseException(CCJSqlParser.java:16995) ~[jsqlparser-1.2-20180105.233414-16.jar:na]
[...]

PR #741, opened 2019-02-04
Added support for 'ALTER TABLE CHANGE COLUMN'

### LINKED ISSUE #380: Alter tab support for 'change'
This script cannot be parsed : Alter table X change column a b Double null default null;
It throws :
Caused by: net.sf.jsqlparser.parser.ParseException: Encountered " <S_IDENTIFIER> "CHANGE "" at line 5, column 1.
Was expecting one of:
    "DROP" ...
    "ADD" ...

Please add support for CHANGE.

Regards

Gikkman:
Added support for 'ALTER TABLE CHANGE COLUMN oldName newName columnDefinition'. 
Please see https://dev.mysql.com/doc/refman/8.0/en/alter-table.html for reference.

I was not really sure how to modify the AlterStatement class. I found introducing these two new fields was the easiest solution, but I am open to suggestions should there be a better way of doing it.
Also, this is my first time writing anything in JavaCC, so if I've made some obvious mistakes or bad practises, I would gladly take feedback. 

This should fix #380 and parts of #571

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java
+++ b/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java
@@ -37,7 +37,9 @@ import net.sf.jsqlparser.statement.select.PlainSelect;
 public class AlterExpression {
 
     private AlterOperation operation;
+    private String optionalSpecifier;
     private String columnName;
+    private String columnOldName;
     //private ColDataType dataType;
 
     private List<ColumnDataType> colDataTypeList;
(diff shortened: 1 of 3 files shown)


wumpz:
Could you write tests using **assertSqlCanBeParsedAndDeparsed** as well. The deparser should be able to completely reconstruct the source SQL.

Gikkman:
Oh, yes, sorry. I had them as test earlier but apparently managed to remove them along the way. I added the requested tests back in.

wumpz:
The reduction on coverage or the coverage value itself seems to be some kind of black magic to me as well. Sometimes the increase or decrease has no clear source.

Issue #586, opened 2018-03-13
net.sf.jsqlparser.JSQLParserException about index issue


Hackeruncle:
VERSION: 1.1
**SQL:**
CREATE TABLE biz_add_fee (
  id bigint(20) NOT NULL AUTO_INCREMENT,
  operation_time timestamp ,
  warehouse_code varchar(32) ,
  customerid varchar(32) ,
  fees_type varchar(32),
  external_no varchar(128) ,
  PRIMARY KEY (id),
  UNIQUE KEY operation_time (operation_time)
) ;

ALTER TABLE BIZ_ADD_FEE DROP INDEX OPERATION_TIME , 
ADD UNIQUE INDEX OPERATION_TIME (OPERATION_TIME, WAREHOUSE_CODE, CUSTOMERID, FEES_TYPE, EXTERNAL_NO) USING BTREE;

**Error:**
net.sf.jsqlparser.JSQLParserException

PraTrick:
Created [PR-752](https://github.com/JSQLParser/JSqlParser/pull/752)

PR #752, opened 2019-02-17
Added support for DROP INDEX, ADD UNIQUE INDEX, ALGORITHM and USING

PraTrick:
Fix handles the following cases: 1) DROP INDEX 2) ADD UNIQUE INDEX 3) ALGORITHM 4) USING <index type>

> `ALTER TABLE biz_add_fee DROP INDEX operation_time, ADD UNIQUE INDEX operation_time (`operation_time`, `warehouse_code`, `customerid`, `fees_type`, `external_no`) USING BTREE, ALGORITHM = INPLACE`

Fixes [586](https://github.com/JSQLParser/JSqlParser/issues/586)


wumpz:
Please integrate some tests using: the deparsers as well.

PraTrick:
May be I've missed something here, but the support for Alter de-parser is rather weak [AlterDeParser](https://github.com/JSQLParser/JSqlParser/blob/jsqlparser-1.4/src/main/java/net/sf/jsqlparser/util/deparser/AlterDeParser.java). Please let me know how to proceed henceforth.

wumpz:
You are right. Maybe it is something you want to implement. The deparsers and toString are two ways to get to the parsed sql, therefore it is a nice check.

PR #769, opened 2019-03-27
KSql 'Within' Clause  Time Unit Bug Fix

Kunal-Jha:
The new release of JSQLParser breaks in case of singular Time units usage. For example usage of `MINUTE` was failing but `MINUTES` was fine. This bug fix would allow us to use both singular and plural time units.


Kunal-Jha:
Could someone please review this PR? It's quite important for our use case.

Kunal-Jha:
@wumpz Sorry for the tag. But I would appreciate if you or any of the maintainers could take a look on this big fix?

esabol:
Pinging @wumpz again

AnEmortalKid (approved):

PR #800, opened 2019-05-21
support postgresql CREATE INDEX syntax

ted-johnson:
The postgresql CREATE INDEX statement uses the USING clause to indicate the indexing method, e.g.
   CREATE INDEX myindex1 ON mytab USING GIST (mycol)
This PR adds support for this syntax.

The PR passes all tests, and a test has been added for the USING syntax.
testCreateIndexIssue633 in CreateIndexTest.java will likely now pass.

Most other SQL dialects indicate the indexing method in the parameters between CREATE and INDEX
     CREATE (parameter)* INDEX
and in some cases, there can be two parameters, e.g. in TransactSQL one can say
    CREATE Unique Clustered INDEX
The current code will only pick up "Unique" for Index.type and will drop "Clustered".
In general, Index.type should be List<String>

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/create/index/CreateIndex.java
+++ b/src/main/java/net/sf/jsqlparser/statement/create/index/CreateIndex.java
@@ -4,7 +4,19 @@
  * %%
  * Copyright (C) 2004 - 2019 JSQLParser
  * %%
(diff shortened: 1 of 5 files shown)


wumpz:
Could you keep the license header intact?

ted-johnson:
fixed the license header

PR #804, opened 2019-05-30
Add generated sources to classpath.

matthieu-vergne:
When importing the Maven project into Eclipse, the sources generated are not automatically included. This commit aims to do that.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,6 +84,25 @@
     <build>
         <plugins>
             <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>build-helper-maven-plugin</artifactId>
+                <executions>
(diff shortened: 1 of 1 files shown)


gitmotte:
@wumpz - i'm using eclipse too, could we merge this ? How can i review this merge-request within github - do i need special rights?

wumpz:
@gitmotte to be honest I don't know. You should be able to write a comment.

wumpz:
The merge will take awhile. But it sounds reasonable and should not interfere with a normal maven build. Or am I wrong?

gitmotte:
> The merge will take awhile. But it sounds reasonable and should not interfere with a normal maven build. Or am I wrong?

Yes it will not affect the normal build.

PR #805, opened 2019-05-30
Fix: Plugin execution not covered by lifecycle configuration.

matthieu-vergne:
When opening project into Eclipse, some goals are not recognized in the POM and generate errors. This commit aims to fix that.

This is not a manual fix, this is automatically generated through a quick fix from Eclipse.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -246,6 +246,60 @@
                 <version>2.15</version>
             </plugin>
         </plugins>
+        <pluginManagement>
+        	<plugins>
+        		<!--This plugin's configuration is used to store Eclipse m2e settings only. It has no influence on the Maven build itself.-->
+        		<plugin>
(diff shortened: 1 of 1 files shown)


wumpz:
I have no problem, if Eclipse would add some ini file, like netbeans does. However, I don't like to include IDE specific plugins within the pom.xml.

matthieu-vergne:
I understand the point. Unfortunately, the only other way provided by Eclipse is to include it in files which are not committed. We may store the information, but it still requires a manual operation from the user. If the user is able to fix it this way, then it means he knows about the issue and how to fix it, in which case he can use one of the quick fixes provided by Eclipse already. In this case, there is no added value to add anything at all.

If someone else knows about another solution, please suggest one. Since I don't see another way, I close it.

PR #819, opened 2019-07-02
Adding support for IS [NOT] TRUE/FALSE expressions

tomershay:
Example: SELECT col FROM tbl WHERE col IS TRUE


wumpz:
Wouldn't it be better to merge IsNullExpression and IsBooleanExpression? As I see so far there is a
IS UNKOWN expression as well.

tomershay:
Thank you for your feedback. Will try to address both points:
1. I don't think it's a good idea to join both classes, as they are not really related. Also, when using the visitor pattern, one may want to visit all objects of IsNullExpression, without visiting the IsBooleanExpression, and it will make the entire process more cumbersome.
2. This was a quick fix, so I didn't implement the IS UNKNOWN part yet, but might add it in the future. This pull request can stand on its own though, even if IS UNKNOWN isn't yet supported.

PR #822, opened 2019-07-04
Add support for full text search (MATCH..AGAINST)

tomershay:
Example:
`SELECT MATCH (col1,col2,col3) AGAINST ('test' IN NATURAL LANGUAGE MODE) relevance FROM tbl`

Documentation: 
https://dev.mysql.com/doc/refman/5.7/en/fulltext-search.html


wumpz:
Thx for improving JSqlParser

PR #825, opened 2019-07-11
Pull request for two changes needed for the integration of JSqlParser in SQuirreL SQL

gerdwagner:
Hi,

thanks for your great work on JSqlParser.

I just uploaded a new snapshot version of SQuirreL SQL client http://www.squirrelsql.org that replaces its former parser by JSqlParser, see 
https://sourceforge.net/projects/squirrel-sql/files/3-snapshots/snapshot-20190711_2226/

There were two things I needed, which are included in this pull request.

One more feature that would be nice but that I found a workaround for is if TablesNamesFinder would have a getter for Table objects instead of table names only.

Thanks again
Gerd

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -504,7 +504,10 @@ Statements Statements() #Statements :
 JAVACODE
 void error_skipto(int kind) {
   ParseException e = generateParseException();
-  System.out.println(e.toString());
+
+  // Could this please be made switchable by some verbose flag. It would even be Ok to me if was switched on by default.
+  // System.out.println(e.toString());
+
   Token t;
   do {
     t = getNextToken();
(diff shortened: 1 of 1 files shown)


wumpz:
I added some review comments. 

Thx for using and improving JSqlParser.

wumpz:
The problem with **getTables** is, that the result of **TableNamesFinder** would change in a way, (duplicates). There could be the same table name for two different **Table** instances. I think the right way to tackle this is to override **protected String extractTableName(Table table)** and collect it by your own, or the internal structure of TableNameFinder should be modified to something like Map<String, List<Table>> where hash key is the fully qualified table name.

wumpz:
Any progress here?

gerdwagner:
Hm, probably closing the request wasn't the right choice.

I request want my changes to be merged. 
If there's something I've got to do please let me know.

wumpz:
I changed the two lines from your PR by hand. So we can close this PR. Thx for using JSqlParser.

PR #835, opened 2019-08-01
Suggest default interface method to all of visitor.

KyongSik-Yoon:
When create new instance of visitor we should to implement all method. 
But sometimes it's seems too much boilerplate code in case that only want implement limited functionality. 
As I know recently this project decide use JDK 8. So this time is chance to feeling benefit of modern java.


wumpz:
I am not sure if I like this massive usage of `default`. IMHO it is a mechanism to extend existing interfaces due to the lack of multi inheritance in java. Here it is a replacement for a standard empty implementation. For this standard implementation there are the adapter classes.

KyongSik-Yoon:
My point is a little different.
We can find why default method suggested to modern java on Oracle documentation.
(https://docs.oracle.com/javase/tutorial/java/IandI/defaultmethods.html)
`Default methods enable you to add new functionality to the interfaces of your libraries and ensure binary compatibility with code written for older versions of those interfaces.`

All of instance implements interface that added some new method always could not compiled unless implements each of all that. 
It's not inconvenient on point of maintenance. 
When need to add new functionality to interface then we just add default method.
It will not breaking all of various instance of interface.

wumpz:
You are right: .. **to add new functionality to existing interfaces**.

AnEmortalKid:
I particularly like the requirement of needing to implement new methods when I upgrade my version of jsqlparser. 

I have abstract classes at the root of my hierarchy that throw an unsupported operation exception, which I always copy paste to new methods. 

If we had the default no op, we might spend some time wondering why things stopped working without realizing there’s a new method that we should have overridden but we instead took an implementation that did nothing 

I could see the use case for defaulting new methods starting with the java 1.8 version but not one for retrofitting everything.

wumpz:
After a long thought I come to the conclusion, this will not be included. Sorry.

PR #861, opened 2019-10-06
Adding support for STRAIGHT_JOIN in the select clause

tomershay:
In MySQL, the STRAIGHT_JOIN flag can be used in both the FROM clause and the SELECT clause.
This pull request adds support for this flag in the SELECT clause.

A sample of a valid query:
`SELECT STRAIGHT_JOIN col, col2 FROM tbl INNER JOIN tbl2 ON tbl.id = tbl2.id`

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/select/PlainSelect.java
+++ b/src/main/java/net/sf/jsqlparser/statement/select/PlainSelect.java
@@ -35,6 +35,7 @@ public class PlainSelect extends ASTNodeAccessImpl implements SelectBody {
     private Fetch fetch;
     private OptimizeFor optimizeFor;
     private Skip skip;
+    private boolean straightJoin;
     private First first;
     private Top top;
     private OracleHierarchicalExpression oracleHierarchical = null;
@@ -170,6 +171,14 @@ public class PlainSelect extends ASTNodeAccessImpl implements SelectBody {
         this.skip = skip;
     }
 
+    public boolean getStraightJoin() {
+        return this.straightJoin;
+    }
+
+    public void setStraightJoin(boolean straightJoin) {
+        this.straightJoin = straightJoin;
+    }
+
     public First getFirst() {
         return first;
     }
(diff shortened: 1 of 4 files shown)


wumpz:
Are you sure, your SQL is valid. I look up some mysql documentation:

https://www.w3resource.com/mysql/advance-query-in-mysql/mysql-straight-join.php

there this **straight_join** is used like a normal join at the same position:

```
SELECT table113.id,table113.cval1,
table111.id,table111.aval1
FROM table113
STRAIGHT_JOIN table111
ON table113.id=table111.id;
```

tomershay:
@wumpz, yes, this is valid MySQL extension.
Look at the syntax description on MySQL's documentation: https://dev.mysql.com/doc/refman/8.0/en/select.html

> SELECT
>     [ALL | DISTINCT | DISTINCTROW ]
>       [HIGH_PRIORITY]
>       **[STRAIGHT_JOIN]**
>       [SQL_SMALL_RESULT] [SQL_BIG_RESULT] [SQL_BUFFER_RESULT]
>       [SQL_NO_CACHE] [SQL_CALC_FOUND_ROWS]
>     select_expr [, select_expr ...]
>     [FROM table_references
>       [PARTITION partition_list]
>     [WHERE where_condition]
>     [GROUP BY {col_name | expr | position}, ... [WITH ROLLUP]]

Also, a quote from that page that describe this capability:
> Following the SELECT keyword, you can use a number of modifiers that affect the operation of the statement. HIGH_PRIORITY, STRAIGHT_JOIN, and modifiers beginning with SQL_ are MySQL extensions to standard SQL.

wumpz:
OK. So this seems to be an optimizer hint for mySql only. You should rename the objects instance variable and methods to something like **mySqlHintStraightJoin**.

tomershay:
@wumpz, thanks for the suggestion, done.

tomershay:
@wumpz, anything else you think should be done to get this PR ready for merge?

wumpz:
@tomershay Sorry. Overlooked your last commit.

PR #886, opened 2019-10-27
Column no longer final.

lscoughlin:
Because it would super awesome for column to no longer be final :)


pom.xml
```diff
@@ -2,7 +2,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.github.jsqlparser</groupId>
     <artifactId>jsqlparser</artifactId>
-    <version>3.1-SNAPSHOT</version>
+    <version>3.1-LSC</version>
```
wumpz: Please don't change the version number.

wumpz:
integrated differently

PR #910, opened 2019-12-19
Adding support for simple expressions in INTERVAL expressions

tomershay:
This far, only columns or constants were supported in an INTERVAL expression.

Now adding support for simple expressions, such as function calls and addition/subtraction expressions.
Parsing this query failed before this fix:
`SELECT DATE_SUB(mydate, INTERVAL DAY(anotherdate) - 1 DAY) FROM tbl`

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -3095,7 +3095,7 @@ IntervalExpression IntervalExpression() : {
 {
     
 { interval = new IntervalExpression(); }
-    <K_INTERVAL> ["-" {signed=true;}] (token=<S_LONG> | token=<S_DOUBLE> | token=<S_CHAR_LITERAL> | expr = Column()) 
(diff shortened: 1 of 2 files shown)


wumpz:
Thx

wumpz:
Unfortunately I had to rewind this pull request. It was not complete. Thee are multiple Lookahead problems. The result is unforeseeable behaviour.

tomershay:
@wumpz, can you please provide an example of when this change may cause an issue?

wumpz:
If you create the parser with your change, a lot of Lookup warnings are shown. There the parser cannot decide, which way to go. It takes the first, I think. If I remember right, now it depends on rule position in the jjt file.

wumpz:
I tried to get around this by rewriting the interval rule to allow only simple expressions when the whole is ended by a time literal, but without success so far. This is preferred since every lookup is a slow down.

wumpz:
By the way, which database supports this? I checked Postgresql til 9.6, Oracle, MySql 8 but could not run any simple command like this.

After digging around I could run it with MySql 5.7.

wumpz:
I added a new branch **solution-issue-910** to solving this one. I put another effort into it but was not successfull. I think there is a circular problem within SimpleExpression which itself could produce another **interval** expression.

PR #914, opened 2019-12-21
#913 SQL Server variable assignment in a select does not parse | SELECT @var = 1

### LINKED ISSUE #913: SQL Server variable assignment in a select does not parse | SELECT @var = 1
**Describe the bug**
SQL Server supports assigning a value to a variable from a selected expression.
For example:
```
DECLARE @var INT;
CREATE TABLE tableName (id INT);

SELECT @var = 1;
SELECT @var = id FROM tableName;
```

**To Reproduce**
` assertSqlCanBeParsedAndDeparsed("SELECT @var = 1");`

**System**
 - SQL Server 2017
- JSqlParser 3.1

cdandoy:
This fixes #913 but breaks SelectTest.testMultiPartNamesIssue608().
Unfortunately, I don't know JavaCC enough to fix both.

My test works: `SELECT @var = 1`
but it breaks `SELECT @@sessions.tx_read_only`


wumpz:
The function is integrated, but not with your pull request.

PR #975, opened 2020-04-20
Add a formatter

AnEmortalKid:
I configured the https://github.com/revelc/formatter-maven-plugin with the [sun style](https://www.genuitec.com/forums/topic/where-to-get-sun-standard-style-templates-for-eclipse/) conventions. The only two changes I made to those formatter settings were to set "spaces" as the only indentation and to set the indent level to `4`. 

The change of formatting all the sources is in this commit: fbdd711

Hopefully this makes it easier for contributors to not have to worry about sun style!

I might have to adjust the line length it seems since a bunch of the javadoc got changed. If you have the number you use, I'll set it to that.


AnEmortalKid:
@wumpz are you interested in a formatter for the project?

PR #978, opened 2020-04-29
Fix CYCLE and GLOBAL token matching tests

AnEmortalKid:
Fixes the regressions introduced by #977


AnEmortalKid:
@wumpz fixed here. No idea why coveralls/travis isn't kicking in, if there's anything on my part I can do to fix that lmk. 

It might help to have `master` be a protected branch that requires the travis check to always succeed to avoid this in the future.

Issue #112, opened 2015-02-24
Support SEQUENCES


itadventurer:
``` sql
CREATE SEQUENCE foo_seq
    START WITH 1
    INCREMENT BY 1
    NO MINVALUE
    NO MAXVALUE
    CACHE 1;


ALTER SEQUENCE foo_seq OWNED BY foo.id;
```
- http://www.postgresql.org/docs/9.4/static/sql-createsequence.html
- http://www.postgresql.org/docs/9.4/static/sql-altersequence.html

pumpkinlink:
+1 for PL/SQL SEQUENCE

AnEmortalKid:
I believe in some fashion this has been completed @wumpz !

wumpz:
@itadventurer @AnEmortalKid This one has completed indeed. Thanks for your patch, @AnEmortalKid. I have to admit, that I did not deploy the changes. The deployment runs as I write.

PR #980, opened 2020-05-03
Support ALTER SEQUENCE 

AnEmortalKid:
Part of #112 

Did a little refactoring mostly around the deparser (sharing a common class) and made a common production for SequenceProperties (both alter/create can use it).

Once this is merged, I'll follow up with a DROP SEQUENCE pr, and that should complete the sequences feature (at least for oracle syntax)


src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
```diff
@@ -464,8 +465,12 @@ Statement SingleStatement() :
         LOOKAHEAD(CreateView())
         stm = CreateView()
         |
+        LOOKAHEAD(AlterView())
```
AnEmortalKid, line 468: I believe this is needed since we now have 2 productions that match on `K_ALTER`, so we need to peek and check if it is one or the other

Issue #993, opened 2020-06-14
Unable to parse `EXPLAIN VERBOSE ...`


auntyellow:
**Describe the bug**
Unable to parse `EXPLAIN VERBOSE ...`

**To Reproduce**
1. `EXPLAIN VERBOSE SELECT * FROM test`
2. Parsing this SQL using JSqlParser with this statements
3. Exception: `Encountered unexpected token: "VERBOSE" <S_IDENTIFIER> at line 1, column 9. Was expecting one of: ...`

**Expected behavior**
Should be able to parse statement.

**System**
- Database: PostgreSQL
- Java Version: 1.8
- JSqlParser version: 3.1

wumpz:
This *verbose* keyword is not (yet) allowed there. PRs are wellcome.

AnEmortalKid:
@auntyellow mind checking that pr out and letting me know what version of postgres you're using. I found some conflicting syntax between requiring `explain (verbose buffers) ...` or `explain ANALYZE ...`

auntyellow:
Sorry for missing system context.

I'm implementing a Java Database with PostgreSQL compatibility, and testing it with different SQL clients. One client `Valentina Studio` will send this kind of SQL.

I tested my database with your PR `AnEmortalKid:explain-verbose` and the SQL is well parse. Thanks for your help.

AnEmortalKid:
This was fixed by #996

wumpz:
@AnEmortalKid if you comment your commits with **fixes #1234** it will close the issue if your pr is merged.

AnEmortalKid:
Yeah... I forgot to do it for that one PR :(

PR #996, opened 2020-06-19
Support options for Explain

AnEmortalKid:
Supports some of the postgres options for Explain based on: https://www.postgresql.org/docs/current/using-explain.html

I made the classes take values just in case feedback from #993 comes back as needing to know what those values are (or if they are needed in the future). Might need some clarification there since there's a piece of doc I found that mentioned the order was [ANALYZE VERBOSE] or something with parenthesis, so depending on the postrgres version this might change.


auntyellow:
It's better to add `getOptionType` in `ExplainStatement.Option`, or simply add a `Map<OptionType, String>` into `ExplainStatement`.
I have to write too much codes to detect this option, e.g.:

			CCJSqlParser parser = new CCJSqlParser(new StringProvider(sql));
			Statement st = parser.Statement();
			if (st instanceof Select) {
				....
			} else if (st instanceof ExplainStatement) {
				ExplainStatement es = (ExplainStatement) st;
				List<Option> options = es.getOptions();
				for (int i = 0; i < options.size(); i ++) {
					if (options.get(i).formatOption().equals("VERBOSE")) {
						...
                                        }
				}
			}

It can be simplified like:

				ExplainStatement es = (ExplainStatement) st;
				if (st.getOptionMap().containsKey(OptionType.VERBOSE)) {
					...
                                }

AnEmortalKid:
@auntyellow how would you feel about

```
Option verbose = ExplainStatement.getOption(OptionType.VERBOSE);
```

Since they _can_ have a value of like TRUE,FALSE,0,1,NO,YES (hadn't gotten to that parsing if it wasn't needed)

auntyellow:
> ```
> Option verbose = ExplainStatement.getOption(OptionType.VERBOSE);
> ```
> 
That's good.

AnEmortalKid:
Cool i'll work on getting that part updated then :)

pom.xml
```diff
@@ -413,6 +413,29 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
```
wumpz, line 416: Why did you add all those profiles? The main build seems to be fast enough even with all this enabled.

src/main/java/net/sf/jsqlparser/statement/ExplainStatement.java
```diff
@@ -11,9 +11,16 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * An {@code EXPLAIN} statement
+ */
 public class ExplainStatement implements Statement {
 
     private Select select;
+    private List<Option> options;
```
wumpz, line 23: I don't know exactly if an option can be here multiple times. So if then a Map or Set would be better here. Due to direct access to this list one is able to build inconsistent multiple options of the same type using getOptions or setOptions.

pom.xml
AnEmortalKid: Ah, when i initially try to do something with the parser I like to have the DEBUG_PARSER and DEBUG_LOOKAHEAD options set to true. When you run `mvn clean package` with those, the build takes quite a while (mostly from all the printing). I can revert these since they're only just for my own benefit when trying to add new things, and it's mostly cause i set all the debug options!

src/main/java/net/sf/jsqlparser/statement/ExplainStatement.java
AnEmortalKid: Umm yeah, I think you're right, it can only be a LinkedSet (order preserved). I'll change that up!

pom.xml
wumpz: No if you find them useful keep them, but keep the standard behaviour untouched.

AnEmortalKid: cool, updated!

src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
```diff
@@ -635,13 +643,109 @@ DescribeStatement Describe(): {
...
+
+/**
+ * The output format, which can be TEXT, XML, JSON, or YAML
+ */
+String ExplainFormatOption():
+{
+  Token tk = null;
+}
+{
+   // TODO support Text
```
AnEmortalKid, line 685: ran into issues with conflicting "TEXT", so i've just left it out atm until someone requests it.

PR #1000, opened 2020-06-21
Minimal function/procedure support

### LINKED ISSUE #111: Recognize CREATE FUNCTION statement
``` sql

CREATE FUNCTION fun(query_from_time date) RETURNS TABLE(foo double precision, bar double precision)
    LANGUAGE plpgsql
    AS $$
      BEGIN
       RETURN QUERY
      WITH bla AS (
        SELECT * from foo)
      Select * from bla;
      END;
      $$;
```

http://www.postgresql.org/docs/9.1/static/sql-createfunction.html

For me it would be enough if JSql sees that its a function and has the content of the function as a string.

This seems quite hard because we need to be able to recognize changes of escape characters and so on.

### LINKED ISSUE #251: Failed to parse CREATE PROCEDURE statement
Statement statement = CCJSqlParserUtil.parse("CREATE PROCEDURE Test" +
                "AS" +
                "BEGIN" +
                "SELECT \* FROM MyTable" +
                "END");

Above code failed.

Exception in thread "main" net.sf.jsqlparser.JSQLParserException
    at net.sf.jsqlparser.parser.CCJSqlParserUtil.parse(CCJSqlParserUtil.java:54)
    at com.zone24x7.Main.main(Main.java:14)
Caused by: net.sf.jsqlparser.parser.ParseException: Encountered " "_" "_ "" at line 1, column 36.
Was expecting one of:
    "ON" ...
    "KEY" ...
    "NOT" ...
    "NULL" ...
    "INDEX" ...
    "DELETE" ...
    "UPDATE" ...
    "PRIMARY" ...
    "FOREIGN" ...
    "REFERENCES" ...
    "ROWS" ...
    "COMMIT" ...
    "UNIQUE" ...
    "CASCADE" ...
    <S_DOUBLE> ...
    <S_LONG> ...
    <S_IDENTIFIER> ...
    <S_CHAR_LITERAL> ...
    "=" ...
    "(" ...
    "+" ...
    "-" ...

AnEmortalKid:
Fixes #251 
Fixes #111

Provides very minimal support for function/procedure parsing, returning everything after `FUNCTION` and `PROCEDURE` as a String (or List<Strings> if desired for other formatting)

### DIFF
--- a/src/main/java/net/sf/jsqlparser/util/TablesNamesFinder.java
+++ b/src/main/java/net/sf/jsqlparser/util/TablesNamesFinder.java
@@ -56,18 +56,7 @@ import net.sf.jsqlparser.expression.operators.conditional.OrExpression;
 import net.sf.jsqlparser.expression.operators.relational.*;
 import net.sf.jsqlparser.schema.Column;
 import net.sf.jsqlparser.schema.Table;
-import net.sf.jsqlparser.statement.Block;
-import net.sf.jsqlparser.statement.Commit;
-import net.sf.jsqlparser.statement.DeclareStatement;
-import net.sf.jsqlparser.statement.DescribeStatement;
-import net.sf.jsqlparser.statement.ExplainStatement;
-import net.sf.jsqlparser.statement.SetStatement;
-import net.sf.jsqlparser.statement.ShowColumnsStatement;
-import net.sf.jsqlparser.statement.ShowStatement;
-import net.sf.jsqlparser.statement.Statement;
-import net.sf.jsqlparser.statement.StatementVisitor;
-import net.sf.jsqlparser.statement.Statements;
-import net.sf.jsqlparser.statement.UseStatement;
+import net.sf.jsqlparser.statement.*;
 import net.sf.jsqlparser.statement.alter.Alter;
 import net.sf.jsqlparser.statement.alter.sequence.AlterSequence;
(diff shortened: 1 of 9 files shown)


wumpz:
I always avoided implementing this due to the different programming syntaxes of databases. But I think you have found a good compromise.

src/main/java/net/sf/jsqlparser/util/TablesNamesFinder.java
```diff
@@ -56,18 +56,7 @@
...
-import net.sf.jsqlparser.statement.DescribeStatement;
-import net.sf.jsqlparser.statement.ExplainStatement;
-import net.sf.jsqlparser.statement.SetStatement;
-import net.sf.jsqlparser.statement.ShowColumnsStatement;
-import net.sf.jsqlparser.statement.ShowStatement;
-import net.sf.jsqlparser.statement.Statement;
-import net.sf.jsqlparser.statement.StatementVisitor;
-import net.sf.jsqlparser.statement.Statements;
-import net.sf.jsqlparser.statement.UseStatement;
+import net.sf.jsqlparser.statement.*;
```
hola-holo, line 59: our static analysis engine detects that your pull request include change from an explicit import to a wildcard import. According to the best practice (http://www.javadude.com/articles/importondemandisevil.html), import star (import on demand) is a bad practice, wildcard import can sometimes cause name conflicts and ambiguities.

PR #1007, opened 2020-07-03
let all deparsers extend AbstractDeParser

gitmotte:
add extend AbstractDeParser<T> to all deparsers

### DIFF
--- a/src/main/java/net/sf/jsqlparser/util/deparser/AlterViewDeParser.java
+++ b/src/main/java/net/sf/jsqlparser/util/deparser/AlterViewDeParser.java
@@ -13,25 +13,25 @@ import net.sf.jsqlparser.statement.create.view.AlterView;
 import net.sf.jsqlparser.statement.select.PlainSelect;
 import net.sf.jsqlparser.statement.select.SelectVisitor;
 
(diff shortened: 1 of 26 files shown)


src/main/java/net/sf/jsqlparser/util/deparser/AlterViewDeParser.java
```diff
@@ -13,25 +13,25 @@
...
 
-public class AlterViewDeParser {
+public class AlterViewDeParser extends AbstractDeParser<AlterView> {
 
-    protected StringBuilder buffer;
     private SelectVisitor selectVisitor;
 
     public AlterViewDeParser(StringBuilder buffer) {
+        super(buffer);
         SelectDeParser selectDeParser = new SelectDeParser();
```
AnEmortalKid, line 22: should we make SelectDeParser also take the buffer instead of calling `setBuffer` ?

gitmotte: I don't care, but if it's for consistency, we could implement it...

wumpz:
Thx for providing this pull request and thx @AnEmortalKid for reviewing this.

PR #1021, opened 2020-07-23
Remove unnecessary parenthesis in ON clause of MERGE statement

beikov:


beikov:
Not sure I understand the question. AFAICS an expression can have parenthesis around for "grouping" purposes anyway, so removing the explicit parenthesis should be fine.

wumpz:
That's what I ment. At the moment I am not able to merge. Give me some time.

wumpz:
Did you check your changes? I had many failing tests after merging it. So for now I reverted it.

beikov:
I did not check the tests. I thought that this was an easy fix because optional parenthesis are already defined by the expression rule, but if this is more involved, I guess it's not worth pursuing this further, especially since Oracle mandates the same syntax.
Would be great to have feedback on the PR about the test runs though. Have you thought about setting up Github Actions or Travis CI for this?

Issue #985, opened 2020-05-16
Support Foreign Key ON UPDATE CASCADE 


dprutean:
Two more statements to check ( ON UPDATE [RESTRICT | CASCADE | SET NULL | NO ACTION | SET DEFAULT})

ALTER TABLE `texto`
  ADD CONSTRAINT `texto_autor_id_foreign` FOREIGN KEY (`autor_id`) REFERENCES `users` (`id`) ON UPDATE CASCADE,
  ADD CONSTRAINT `texto_tipotexto_id_foreign` FOREIGN KEY (`tipotexto_id`) REFERENCES `tipotexto` (`id`) ON UPDATE CASCADE

Encountered unexpected token: "UPDATE" "UPDATE"
    at line 2, column 97.

ALTER TABLE `texto_fichero`
  ADD CONSTRAINT `texto_fichero_fichero_id_foreign` FOREIGN KEY (`fichero_id`) REFERENCES `fichero` (`id`) ON DELETE CASCADE ON UPDATE CASCADE,
  ADD CONSTRAINT `texto_fichero_texto_id_foreign` FOREIGN KEY (`texto_id`) REFERENCES `texto` (`id`) ON DELETE CASCADE ON UPDATE CASCADE

Encountered unexpected token: "ON" "ON"
    at line 2, column 126.

wumpz:
JSqlParser supports only **ON DELETE** (at the moment). Pull requests are welcome.

gitmotte:
this is merged already an can be closed.

PR #1025, opened 2020-07-26
Support Foreign Key ON UPDATE CASCADE

gitmotte:
* implement COMMENT ON VIEW
* testcase "testCommentOnView"
https://github.com/JSQLParser/JSqlParser/issues/985


gitmotte:
@wumpz i fixed this issue, because i' using ON UPDATE too

gitmotte:
> Did you replace all occurances of this? If I remember right there was one direct in the alter expression and another in an ForeignIndex class.

Yes, i replaced both. Added the same unit-tests for the AlterExpression variant.

gitmotte:
ok, it's ready to merge now :)

Issue #991, opened 2020-06-11
Unable to parse `SELECT 0 IS NULL`


auntyellow:
**Describe the bug**
Unable to parse `SELECT 0 IS NULL`

**To Reproduce**
1. `SELECT 0 IS NULL` or `SELECT 1 IS NOT NULL` or `SELECT 2 IN (0, 1)`
2. Parsing this SQL using JSqlParser with this statements
3. Exception: `net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: "IS" at line 1, column 9. Was expecting one of: ...`

**Expected behavior**
Should be able to parse statement.

**System**
- Database: MySQL (returns `0`) and PostgreSQL (returns `false`)
- Java Version: 8
- JSqlParser version: 3.1, 3.2-SNAPSHOT (with latest `master`)

auntyellow:
Looks duplicated with #720

wumpz:
JsqlParser 4 parses this

Issue #720, opened 2018-12-06
`SELECT column IS NOT NULL` not parsed


datto-aparrill:
Example query:
```sql
SELECT column IS NOT NULL FROM table
```
`CCJSqlParserUtil.parse` throws an error:

```plain
[...]

PR #1032, opened 2020-08-10
bugfix #720 #991: supporting SELECT "conditions"

mo3000:
As it is in the issues above, sqls below won't parse.
```sql
SELECT column IS NOT NULL FROM table;
SELECT 0 IS NULL;
```
Then I figured out that in function `SelectExpressionItem()` -> `SimpleExpression()` should be replaced by `Condition()` since null expression, bool expression and SimpleExpression itself are handled in `Condition()`.
Because function `PivotSingleInItems()` also use `SelectExpressionItem()`, I copyed the original `SelectExpressionItem()` so that the "pivot" test doesn't break.

All tests passed, but "SELECT 1 < 2 AS a, (0 IS NULL) AS b" won't work because something is wrong with aliasing an  expression with parentheses. 
"SELECT (1 < 2) AND (3 < 4)" won't work too, I belive this can be fixed somehow in function `Condition()`.


wumpz:
Did you run all tests? I had a similar approach and the performance dropped significantly.

mo3000:
I run "mvn test" before and after this pr. 
Before:
```[INFO] Total time:  01:17 min```
After:
```[INFO] Total time:  01:16 min```
I tried several times, running time is around 1:20 on my mac (open jdk 14), no performance dropped. mvn says "[INFO] BUILD SUCCESS" , is there anything I should run specifically?

wumpz:
Strange, but very fortunate. Maybe some of the recent changes corrected this behaviour.

wumpz:
The test on my machine is nearly 50% slower with your modification. Strange that you did not experience such a performance drop.

wumpz:
Now I modified some productions and gained more speed. Nearly the same as before.

mo3000:
teriffic! I'll try that on my machine.

PR #1057, opened 2020-09-28
Retain original value in TimestampValue

eolivelli:
Currently we only keep a reference to a parsed java.sql.Timestamp, but java.sql.Timestamp.valueOf uses local machine Timezone to parse the timestamp value, this is not good for statement rewriting if jSQLParser is running on a separate machine (like a server) with different JVM configuration


eolivelli:
Thanks for accepting my patch.
At HerdDB we hope can can see current master released soon 👍

wumpz:
I will try it, however, I have no much time at the moment.

eolivelli:
Thank you much!

PR #1065, opened 2020-10-18
support IN with value

### LINKED ISSUE #574: IN clause without parentheses is not working but work in Oracle DB
The following works on Oracle 11g:
`update AsyncTaskManager set lastActive = SYS_EXTRACT_UTC(SYSTIMESTAMP) where sessionId in sessionId`

AnEmortalKid:
Fixes #574 

I looked at pull https://github.com/JSQLParser/JSqlParser/pull/631 and I think I addressed both of the concerns there:

> something like "in ( expr" would be accepted. If there is an opening parenthesis there must be a closing one. I know it is not yet done for all of JSqlParser, but it should.
> The deparser should reconstruct exactly the same statement from the parse tree. But without any change, it would forget the parenthesis.

Opted for just adding another path that allows a no parenthesis branch but with a single expression instead of messing around with a production of :
```
| "(" STUFF ")"
| STUFF { noParens=true; }
```


wumpz:
Please remove conflicts.

AnEmortalKid:
@wumpz conflicts removed!

Issue #898, opened 2019-11-19
JSqlParser - website discussion


sivaraam:
Let's create a website for the library! This issue is to track the discussion related to the website creation.

sivaraam:
Jekyll template suggested by @wumpz:  [Minimal Mistakes](https://mmistakes.github.io/minimal-mistakes/)

sivaraam:
@wumpz For the sake of completeness, can you list the information that you wish the website should provide to its visitors?

wumpz:
Let me think about it: 

- quickstart (maven, gradle)
- introduction visitor pattern
- introduction object tree of jsqlparser
- common tasks: parsing, replacing, analyzing
- examples

I think the biggest gap is explaining jsqlparsers object tree. Much of this should go into the javadocs as well.

sivaraam:
That looks like a good start! I think it would also be nice to add a link to the javadocs.

Anyways, I'll set up an initial webpage based on the  Minimal Mistakes template and then we'll add more content to it.

In the mean time, if you would like to host the site at https://jsqlparser.github.io/ instead of https://jsqlparser.github.io/jsqlparser you would have to create a repository in the JSQLParser organization with the name "jsqlparser.github.io" ([reference](https://help.github.com/en/github/working-with-github-pages/about-github-pages#types-of-github-pages-sites))

sivaraam:
Hi @wumpz,

Sorry I couldn't get the initial site past weekend. Anyways, I've started looking into it now.

Just wanted to ask you something. Though I initially told the [Minimal Mistakes](https://mmistakes.github.io/minimal-mistakes/) theme was nice. On second thought, it seems to give a dull kind of appearance to the website. So, I think its better to go with some other theme which gives a more appealing look. Let me know your opinion.

If you also agree on choosing an alternative theme, let me know your opinions about the one of these alternatives (or) suggest other ones.

* [Hyde](http://hyde.getpoole.com/about/)
* [Beautiful Jekyll](https://deanattali.com/beautiful-jekyll/)

wumpz:
Beautiful Jekyl looks nice as well. But Minimal Mistakes is kind of a framework. Beautiful Jekyl as well?

sivaraam:
> But Minimal Mistakes is kind of a framework.

[...]

PR #1095, opened 2021-01-07
javadoc fixes, remove doclint:none

gitmotte:
Removed the -Xdoclint:none configuration from pom.xml testing whether the javadoc is valid. 
Did some fixes - but not all errors could be fixed, because there are generated java-sources which are now excluded from javadoc.


wumpz:
So if I understand you right, you could not do a complete build due to unresolvable javadoc problems. Therefore I will not merge.

gitmotte:
The setting doclint:none ignores errors during the generation of javadoc, i.e. the build does **not** abort, in consequence the javadoc-html files are **created incorrectly**.

I fixed this by fixing the existing errors (there are still some warnings, i.e. missing descriptions for parameters, ... ) and by removing the setting -Xdoclint:none within pom.xml - by default, the -Xdoclint option is enabled.

See https://docs.oracle.com/javase/8/docs/technotes/tools/unix/javadoc.html for detailed documentation.

manticore-projects:
While I am no developer, it appears to me that we have two different issues here:

1) the is a lot of invalid JAVADOC in the manually written sources which deserves to get fixed, e.g.
```
Javadoc Warnings
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/parser/CCJSqlParserUtil.java:55: warning - @return tag has no arguments.
/home/are/Documents/src/JSqlParser/target/generated-sources/javacc/net/sf/jsqlparser/parser/Provider.java:26: warning - @param argument "length" is not a parameter name.
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/parser/feature/Feature.java:513: warning - Tag @see: reference not found: CreateSynonym
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java:126: warning - @return tag has no arguments.
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java:150: warning - @return tag has no arguments.
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java:160: warning - @return tag has no arguments.
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java:169: warning - @return tag has no arguments.
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java:179: warning - @return tag has no arguments.
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java:188: warning - @return tag has no arguments.
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java:198: warning - @return tag has no arguments.
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java:150: warning - Tag @link: can't find getOnDeleteReferentialAction() in net.sf.jsqlparser.statement.alter.AlterExpression
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java:169: warning - Tag @link: can't find getOnDeleteReferentialAction() in net.sf.jsqlparser.statement.alter.AlterExpression
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java:188: warning - Tag @link: can't find getOnDeleteReferentialAction() in net.sf.jsqlparser.statement.alter.AlterExpression
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java:160: warning - Missing closing '}' character for inline tag: "{@link #setOnDeleteReferentialAction(Action)"
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java:179: warning - Missing closing '}' character for inline tag: "{@link #setOnDeleteReferentialAction(Action)"
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java:198: warning - Missing closing '}' character for inline tag: "{@link #setOnDeleteReferentialAction(Action)"
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java:160: warning - @return tag cannot be used in method with void return type.
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java:179: warning - @return tag cannot be used in method with void return type.
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java:198: warning - @return tag cannot be used in method with void return type.
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/statement/create/table/ForeignKeyIndex.java:71: warning - @return tag has no arguments.
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/util/SelectUtils.java:57: warning - @return tag has no arguments.
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/util/SelectUtils.java:84: warning - @return tag has no arguments.
/home/are/Documents/src/JSqlParser/src/main/java/net/sf/jsqlparser/util/TablesNamesFinder.java:185: warning - @return tag has no arguments.
```

2) independent from that, there is a question if and how to activate DOCLINT.
In my own opinion, this did not yield in any merit because the automatic generated code will always break DOCLINT and when you look into the parser definition file then JavaDoc is your least concern.

So my suggestion was: why not focus on the brave task of fixiing the JavaDoc for the manually written code, but split off/remove the DOCLINT part. Would you like to amend your PR please, I really would love to see the JavaDoc fixed. Thank you already and cheers.

gitmotte:
The javadoc fix can of course also be carried out without changing the doclint setting. However, in this case javadoc will still not checked for correctness for manually written sources, so switching on doclint validation would only ensure code quality for future changes.

For generated sources, a valid javadoc is desirable, but not possible without fixing javacc, so these sources were excluded from the validation.

@wumpz can of course change the request according to his wishes.

manticore-projects:
> The javadoc fix can of course also be carried out without changing the doclint setting.
Then I would like to suggest to do so, because it looks to me like a low hanging fruit.

> However, in this case javadoc will still not checked for correctness for manually written sources, so switching on doclint validation would only ensure code quality for future changes.
I get your point, but right now the Java Doc is more alike a stub in anyway. It is largely empty and in fact the toString() methods yield the most information at the moment. Personally I would care more about adding flesh to the documentation itself than about its syntax or enforcement.

> For generated sources, a valid javadoc is desirable, but not possible without fixing javacc, so these sources were excluded from the validation.
Agreed.


> @wumpz can of course change the request according to his wishes.
I can only assume that he does not have the necessary time and simply expects any PR to at least survive the CI as a kind of minimum quality assurance. So how can we help to get the cow off the ice? I see three options:

a) you quickly remove the blocking doclint related change in order to salvage the more relevant javadoc fixes
b) you abandon the PR (that would be sad)
c) I will need to pull from your repository, remove the blocking ddoclint part and re-issue a new PR (that would be last stupid and should be the last resort)

Which one will your prefer, it is your work, it is good and I am keen to get it merged.

manticore-projects:
Pardon a maybe stupid question: Am I correct to assume, that the CI fails only because of Codaci Static Code Analysis, which complains about code that has been there before already?

image

manticore-projects:
I think there is just a misunderstanding: the rejection stems from a questionable Codacy Static Code Analysis, but not from DocLint.
(I have tested Coadcy once, laughed about it and deleted it.)

manticore-projects:
Please check PR #1163 
I added
```java
@SuppressWarnings({"PMD.MethodNamingConventions"})
```
to `net.sf.jsqlparser.parser.SimpleCharStream` which resolves 9 PMD issues.
If my understanding is correct, this would make your PR pass the CI without any ado.

@wumpz: Should we not add the PMD plugin to maven and check the same rules as Codacy does. It may annoy any contributor when cleanly compiling PRs get rejected just because the CI wrecks havoc afterwards. I spent almost 3 hours walking through this Codazy/PMD stuff when @gitmotte's PR itself seems to have no issue at all.

gitmotte:
> I think there is just a misunderstanding: the rejection stems from a questionable Codacy Static Code Analysis, but not from DocLint.

yes, these comments/fails from codacy are not new - i don't know how codacy works and how to fix it. it seems to be a code quality tool like sonarlint. 

> a) you quickly remove the blocking doclint related change in order to salvage the more relevant javadoc fixes

no, i won't. as i said, the pr can also be partially adopted - but i don't recommend it.

> b) you abandon the PR (that would be sad)

no, others may find it usefull 

> c) I will need to pull from your repository, remove the blocking ddoclint part and re-issue a new PR (that would be last stupid and should be the last resort)

You are welcome to do so, but the Doclint change does not block a build as you think.

> I get your point, but right now the Java Doc is more alike a stub in anyway. It is largely empty and in fact the toString() methods yield the most information at the moment. Personally I would care more about adding flesh to the documentation itself than about its syntax or enforcement.

I think syntax and its enforcement is the first step to good Javadoc documentation. If not enforced, the errors will cause the build to fail with default doclint - settings. Feel free to extend the documentation further - I'm sorry that I'm not doing this work for the whole library, but only for the parts i added / i have changed.

wumpz:
@manticore-projects You are right, it is an issue of time. 

@gitmotte If I change the code, to be honest, I have to force myself to correct documentation and stuff. That's why doclint:none check is there. To make it easier. 

@manticore-projects Codacy is there because it was sometimes advertised and somehow automatically integrated. You are right that it is a very lightweight checker. PMD should be integrated so the rules are in our hands. It would be a good addition to checkstyle which is already included. So the removal of Codacy is only a matter of time. 

I am a huge fan of self documenting code, to avoid javadocs whereever this is possible. I think like the most developers documentation is a daunting task to me. Thats why much of for instance the old javadocs of the original project from sourceforge are still present even if there is so much to improve. I know. 

```java
/**
     * Read a character.
     * @return the character read
     * @throws IOException
     */
    public char readChar() throws java.io.IOException {
´´´

For instance this javadoc is without additional meaning. It needs to be removed or this

```java
     * @param feature
     * @param namedObject
     * @param fqn - fully qualified name of named object
     * @param alias
     */
    protected void validateFeatureAndNameWithAlias(Feature feature, NamedObject namedObject, String fqn, String alias)  {
```

The better way to document this is to rename the parameter to fullyQualifiedName and again the javadoc is without any additional meaning. 

By the way there is another issue that wanted to start a separate documentation (#898), but the issue owner seemed to have lost its energy. 

So I would say:
1. improve javadocs, via writing a right documentation or replace it with refactoring 
2. let doclint checking switched off
3. include pmd with a starting rule that 
4. lets think about a new start website or wiki page again. somehow one understands why many projects have their own documentation writer and that is definitely not me :/

manticore-projects:
> @manticore-projects Codacy is there because it was sometimes advertised and somehow automatically integrated. You are right that it is a very lightweight checker. PMD should be integrated so the rules are in our hands. It would be a good addition to checkstyle which is already included. So the removal of Codacy is only a matter of time.

As far as I understand it, Codazy just uses PMD (beside a few less usefull checkers, the SQL checker for example got it very wrong and yielded n my harsh judgement). So I think, you should keep it because it creates such a nice badge. However, to make it useful, the PMD check must happen before commit and also the PMD ruleset should be customized and synchronised.

Especially the NPath complexity ( LOL the toString() methods) and the UnusedImports rules are not too helpful.

I have worked on that stuff, please see PR #1165

> I am a huge fan of self documenting code, to avoid javadocs whereever this is possible. 

Fully agreed, there will be ALWAYS a discrepancy between the latest code and any documentation and nothing is worse than wrong/old documentation. That said, I think any existing JavaDoc should be at least syntactical correct. Clean code is good from a Sales perspective too.

> By the way there is another issue that wanted to start a separate documentation (#898), but the issue owner seemed to have lost its energy.

Website would be nice though. What is your opinion on sharing a website for JSQLParser and JSQLFormatter (as soon as JSQLFormatter becomes a bit more matured). They belong together and I spend more effort on yours than on mine over the last 3 days :-D

>     1. improve javadocs, via writing a right documentation or replace it with refactoring

Done by PR #1095, but depends on the PMD related fixes by PR #1165

>     2. let doclint checking switched off

No objection, but DocLint does not even seem to be the issue. I will check that out when revising this PR @1095


>     3. include pmd with a starting rule that

Done, PR #1165 waits  for your approval.

>     4. lets think about a new start website or wiki page again. somehow one understands why many projects have their own documentation writer and that is definitely not me :/

I definitely won't write much documentation, but I could look into this website thing eventually.
Writing a website is easy, the main problem will be how to maintain it over the time.

gitmotte:
> @gitmotte If I change the code, to be honest, I have to force myself to correct documentation and stuff. That's why doclint:none check is there. To make it easier

Doclint none is guaranteed to lead to faulty documentation over time, which is why I like having it turned on. 
By the way, javadoc without comment but with parameter names is very valuable, because the compiler turns parameter names into arg0-argn. An IDE can use the information from the javadoc-jar to show me the parameter names correctly if the source is not available.
Furthermore, javadoc is for me a stylistic device for clarity (between each declaration a javadoc block), but maybe that's just my feeling.

wumpz:
> By the way, javadoc without comment but with parameter names is very valuable, because the compiler turns parameter names into arg0-argn. An IDE can use the information from the javadoc-jar to show me the parameter names correctly if the source is not available.

Which parameters do a compiler turn to arg0-argn? Every jar without javadoc holds the information of the original method names and parameter names. Only by some kind of obfuscation you are loosing those?

image

BTW nearly all IDEs are able to present now parameter names as inline hints so this is available without any additional javadoc opening.

manticore-projects:
I have merged @gitmotte's stuff as good as possible and fixed the DocLint entries.
DocLint is active and will stop when anything material is breaking, but will ignore any Missing Warnings.

See commit commit 36d79b6.

gitmotte:
> > By the way, javadoc without comment but with parameter names is very valuable, because the compiler turns parameter names into arg0-argn. An IDE can use the information from the javadoc-jar to show me the parameter names correctly if the source is not available.
> 
> Which parameters do a compiler turn to arg0-argn? Every jar without javadoc holds the information of the original method names and parameter names. Only by some kind of obfuscation you are loosing those?
> 

only if debug-information for vars is present in class files (By default, only line number and source file information is generated i.e. without -g option is being specified. )
most ide's have this turned on by default, but maven has only <debug>true</debug> by default, not <debuglevel>source,lines,vars</debuglevel>

manticore-projects:
Can we close this? I believe to have merged the relevant parts in 2021 already.

gitmotte:
merged by @manticore-projects

PR #1104, opened 2021-01-21
RESET statement, SET PostgreSQL compatibility

tvar:
RESET statement  (https://www.postgresql.org/docs/current/sql-reset.html)
  SET [LOCAL|SESSION]  (https://www.postgresql.org/docs/current/sql-set.html)
  SET search_path=my_schema, public ( https://www.postgresql.org/docs/current/sql-set.html
        **value** New value of parameter. Values can be specified as string constants, identifiers, numbers, or comma-separated lists of these)


wumpz:
please correct conflicts

tvar:
fixed

src/main/java/net/sf/jsqlparser/statement/SetStatement.java
```diff
@@ -80,33 +82,39 @@ public void setName(int idx, String name) {
         values.get(idx).name = name;
```
aimiebell, line 82: I detect that this code is problematic. According to the [Correctness (CORRECTNESS)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#correctness-correctness), [SA: Self assignment of value (SA_SELF_ASSIGNMENT)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#sa-self-assignment-of-field-sa-field-self-assignment).
This method contains a self assignment of a field; e.g.
Such assignments are useless, and may indicate a logic error or typo.

tvar: I've read code 10 times and can't see problem.
SetStatement.setName(int idx, String name)
here no name conflicts (SetStatement doen't contain fields named "idx" or "name")
then i get "values" (field of SetStatement)
and get "values" element (NameExpr) with index "idx" (param)
and set it(NameExpr) field name to parameter name

where is an expression **x=x** ? i can't see

PR #1109, opened 2021-01-21
subArray support arr[1:3]

tvar:
subArray support arr[1:3]

https://www.postgresql.org/docs/current/arrays.html

We can also access arbitrary rectangular slices of an array, or subarrays. An array slice is denoted by writing lower-bound:upper-bound


wumpz:
Please resolve conflicts first.

tvar:
Resolved

PR #1165, opened 2021-04-19
Pmd

manticore-projects:
I have worked on that PMD stuff and
1) have added the Maven PMD plugin (mandatory before build in order to fail fast)
2) fixed 63 (more or less useful) PMD exceptions
3) define a PMD ruleset (removing some ridiculous rules)

Todo: @wumpz: please share the ruleset defined for Codazy because we would need to synchronize with the project ruleset. Unfortunately I can't see that Codazy ruleset because it is running under your account only. 

Please fast track this PR because it touches many files and will so break any pending PR or change. Please do not let me solving conflicts on the material corrections for the rest of the week.

Viele Gruesse!


manticore-projects:
107 issues reported by Codazy have been fixed, this should elevate the Code Quality to Grade A.

Issue #1176, opened 2021-04-23
"group by ()" cause parsing exception


xsgao-github:
**Describe the bug**
Cannot parse this statement `select sum(c) from t group by ()`.

**To Reproduce**
1. `net.sf.jsqlparser.parser.CCJSqlParserUtil.parse("select sum(c) from t group by ()")`
2. Exception - `Caused by: net.sf.jsqlparser.parser.ParseException: Encountered " "(" "( "" at line 1, column 31.`

**Expected behavior**
Not sure if `group by ()` is SQL compliant, but tested on SQL/PostgreSQL, they both support it.

**System**
 - Database you are using - SQL/PostgreSQL
- Java Version - 1.8
- JSqlParser version - 4.0

manticore-projects:
Greetings.

While "supported by PostgreSQL" is a strong argument for implementing this "feature", I still wonder: Why would anyone want to write code like that? What does the `group by()` add to the statement? Am I right to assume, PostgreSQL would silently drop it and just executes
```sql
select sum(c)
from t
```

instead? If so, why not write the SQL exactly like this? Is any RDBMS expecying the empty `group by ()` clause?
Please correct me when I am wrong, but I kind of hesitate to spend my time on supporting 'dead code'.

xsgao-github:
I also mentioned that SQL server also support that. Didn’t test other db.

I don’t know why someone write it in this way, but Tableau does.

manticore-projects:
H2 supports it too, both as 
```sql
select id_instrument, count(*)
from cfe.instrument
group by (id_instrument);
```

as well as 
```sql
select count(*)
from cfe.instrument
group by ();
```

So be it.

manticore-projects:
Can we close this please, it has been solved, tested and merged.

PR #1178, opened 2021-04-24
Add Support for TSQL NEXT VALUE FOR

### LINKED ISSUE #1177: Parser fails on TSQL "NEXT VALUE FOR"?
Parser 4.1 can't parse:
```sql
-- SELECT
select /*+PARALLEL*/
	a.id_fee_type
	, concat('New fee ' , a.id_fee_type)
    , NEXT VALUE FOR cfe.id_fee_type_ref
	from (
	select distinct
		id_fee_type
	from
		cfe.SCHEDULE
	union select
		id_fee_type
	from
		cfe.ONE_TIME_PAYMENT
	) a left join cfe.FEE_TYPE b on a.id_fee_type=b.id_fee_type
where
	b.id_fee_type is null
	and not a.id_fee_type is null
;
```

I dug into the code and found that `NextValExpression` deparses into `NEXTVAL FOR` (and the query gets parsed with `NEXTVAL FOR`). 
However, this does not seem to be valid TSQL according to https://docs.microsoft.com/en-us/sql/t-sql/functions/next-value-for-transact-sql?view=sql-server-ver15. Is this a bug or a feature? (I am not MS SQL Server guy).

manticore-projects:
fixes #1177 

I do not think, that the previously implemented `NEXTVAL FOR` is correct, but I kept it in order to avoid breaking anything. Which RDBMS would support or request for `NEXTVAL FOR` please?

With this PR, both `NEXTVAL FOR` and `NEXT VALUE FOR` will be supported. Oracle/DB2 `foo.nextval` is not affected.
If we found out, that `NEXTVAL FOR` was obsolete, we should clean this up optionally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ Also I would like to know about needed examples or documentation stuff.
 
 * introduce a max depth to allow parsing complex expression lists without performance loss (thx to @manticore-projects)
 * allow all functions to have complex expressions as parameters (thx to @manticore-projects)
+** API change FunctionWithCondParams production removed
 * API change in ValuesStatement: the expression list is now hold as a ItemList and not as a List<Expression>
 * support for parser modification within **parseExpression** and **parseCondExpression**
 ' support for table schema for foreign keys


manticore-projects:
Pardon my stupid question: why does it keep referring to merged previous PRs and changes? How would I get rid of this please?
It seems like I was not using GIT correctly yet.

manticore-projects:
Add support for CTEs in `MERGE`, `UPDATE`, `DELETE` (but not yet for CTEs with VALUE LISTs)

manticore-projects:
Please please please: Deactivate the rules "PMD.CyclomaticComplexity", "PMD.ExcessiveMethodLength", "PMD.NPathComplexity" for Codazy! These do not make any sense for the `toString()` and `deParse()` methods and right now I need to activate these rules in PMD only to annotate every method with `@SuppressWarnings({"PMD.CyclomaticComplexity", "PMD.ExcessiveMethodLength", "PMD.NPathComplexity"})`.
I am not against Code quality validation at all, but this insane.

manticore-projects:
One more thing: Code Duplication. As far as I understand it, the `toString()` methods do more or less the same as the `Deparser.deparse()` methods. Would that not be the right time to overhaul this and to replace the code of the `toString()` methods with a static method `Deparser.deparse(Statement st)`?

manticore-projects:
fixes #1176

manticore-projects:
On Fri, 2021-04-30 at 15:04 -0700, Tobias wrote:
> So could it be, that here are more changes than you liked? It seems
> to be a big PR.

No, the PR is correct, please pull it.
I will use squash properly next time. Sorry for any incenvenience.

Cheers

wumpz:
Could you first remove the conflict? Then I will merge.

manticore-projects:
Done, cheers!

manticore-projects:
Fixes #1186

wumpz:
Now there are changes in 51 files??? Really?

manticore-projects:
Sorry, I know it is not ideal but what am I supposed to do?

the PR fixes 4 issues from the list
they all touch the JavaCC Parser files and ammend the Java Objects and
the Deparser
when submitting, your code checks complained so I hade to fix these too
(which affected 40 files alone)
which gave a lot of conflicting changes, making it really difficult to
split up

And last but not least, I work on JSQLParser in order to complete the
JSQLFormatter. Every change in JSQLParser affects the formatting and I
need a working JSQLParser
 immediately after I have fixed something.

So how would you want to proceed please? You pulled the #1189 first
(understandable, because its simple) and now the PR is conflicting
again. I will fix it again.

PR #1181, opened 2021-04-27
Row access support

tvar:
Support for accessing elements of row type.
Like this:
`SELECT (t.tup).id, (tup).name FROM t`
https://www.postgresql.org/docs/current/rowtypes.html#ROWTYPES-ACCESSING

Additionally remove in LEFT expression because it's just case of ROW-construct 
without this fix doesn't work queries like this:
`(t.tup).id IN (1, 2, 3)`


wumpz:
Wasn't this already solved in the actual snapshot 4.1?

tvar:
> Wasn't this already solved in the actual snapshot 4.1?

No, I've get current master and add my test case
    @Test
    public void testSelectRowElement() throws JSQLParserException {
        assertSqlCanBeParsedAndDeparsed("SELECT (t.tup).id, (tup).name FROM t WHERE (t.tup).id IN (1, 2, 3)");
    }

It failed. I can't see any code to support access to tuple elements.

wumpz:
Then sorry. I merged.

PR #1184, opened 2021-04-30
supporting/fixing unique inside sql function such as count 

sur5an:
eg - SELECT count(UNIQUE col2) FROM mytable


sur5an:
@wumpz can you please check the new PR --Thanks

PR #1189, opened 2021-05-09
Add support for casts in json expressions

tomershay:
Adding support for column type casts in JSON expressions. For example:
`SELECT id FROM tbl WHERE p.company::json->'info'->>'country' = 'test'`
These casts are supported in PostgreSQL.


wumpz:
Thx for the PR.

PR #1193, opened 2021-05-11
Add support for boolean 'XOR' operator

arh-eu:
The boolean operator `XOR` (exclusive or) is not part of the SQL standard, but many databases have support for them (ex. mySQL).
However, with the current parser these expressions cannot be parsed as the keyword and the expression for it is not part of the grammar.

This pull request adds support for the XOR Expression, that is treated as a Binary Conditional Operator.
The expression behaves the same as the usual `AND` or `OR` expressions, but its precedence is the lowest: `AND` > `OR` > `XOR`.

### DIFF
--- a/src/test/java/net/sf/jsqlparser/statement/builder/JSQLParserFluentModelTests.java
+++ b/src/test/java/net/sf/jsqlparser/statement/builder/JSQLParserFluentModelTests.java
@@ -9,20 +9,11 @@
  */
 package net.sf.jsqlparser.statement.builder;
 
-import static net.sf.jsqlparser.test.TestUtils.*;
-
-import java.util.List;
-
-import net.sf.jsqlparser.expression.operators.conditional.XorExpression;
-import org.junit.Test;
 import net.sf.jsqlparser.JSQLParserException;
-import net.sf.jsqlparser.expression.Alias;
-import net.sf.jsqlparser.expression.Expression;
-import net.sf.jsqlparser.expression.JdbcParameter;
-import net.sf.jsqlparser.expression.Parenthesis;
-import net.sf.jsqlparser.expression.StringValue;
+import net.sf.jsqlparser.expression.*;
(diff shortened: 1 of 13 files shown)


src/test/java/net/sf/jsqlparser/statement/builder/JSQLParserFluentModelTests.java
```diff
@@ -38,14 +40,15 @@
     @Test
     public void testParseAndBuild() throws JSQLParserException {
         String statement = "SELECT * FROM tab1 AS t1 " //
-                + "JOIN tab2 t2 ON t1.ref = t2.id WHERE (t1.col1 = ? OR t1.col2 = ?) AND t1.col3 IN ('A')";
+                + "JOIN tab2 t2 ON t1.ref = t2.id WHERE (t1.col1 = ? OR t1.col2 = ?) AND t1.col3 IN ('A') XOR t1.col4";
```
wumpz, line 43: You should not adapt tests but create new ones.

wumpz (changes requested):

src/test/java/net/sf/jsqlparser/statement/builder/JSQLParserFluentModelTests.java
arh-eu: I've fixed this by introducing a separate test case for the XOR case, restoring the original one to the previous state.

wumpz:
Are you sure you implemented the right precedence?

arh-eu:
The precedence for the `OR` over the `XOR` is implemented the same way as the `AND` is "stronger" than `OR`.

Many internal, dummy tests showed the precedence is right and parsed as intended, however, I can add more test cases that show that parsing goes how it's supposed to do with `OR`, `XOR` and ` AND` in the same condition.

arh-eu:
I've added new tests to showcase the precedence and associativity.

arh-eu:
@wumpz Could you review it again?

Issue #1115, opened 2021-02-01
Not support postgres function?


WenSteven:
**Describe the bug**
my sql is :
`SELECT count(*)
    FROM customers
    WHERE 1 =1 and (created_at at time zone 'Asia/Shanghai')::date = '2021-02-01'::date`
and would get exception:
`
Caused by: net.sf.jsqlparser.JSQLParserException: Encountered unexpected token: "at" <S_IDENTIFIER>
    at line 3, column 32.

Was expecting one of:

    "&"
    "&&"
    ")"
    "::"
    "<<"
    ">>"
    "AND"
    "COLLATE"
    "OR"
    "["
    "^"
    "|"

	at net.sf.jsqlparser.parser.CCJSqlParserUtil.parseStatements(CCJSqlParserUtil.java:188) ~[jsqlparser-4.0.jar:na]
	at net.sf.jsqlparser.parser.CCJSqlParserUtil.parseStatements(CCJSqlParserUtil.java:176) ~[jsqlparser-4.0.jar:na]
	at com.baomidou.mybatisplus.core.parser.AbstractJsqlParser.parser(AbstractJsqlParser.java:62) ~[mybatis-plus-core-3.3.2.jar:3.3.2]
	... 105 common frames omitted
Caused by: net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: "at" <S_IDENTIFIER>
    at line 3, column 32.

Was expecting one of:

    "&"
    "&&"
    ")"
    "::"
    "<<"
    ">>"
    "AND"
    "COLLATE"
    "OR"
    "["
    "^"
    "|"

	at net.sf.jsqlparser.parser.CCJSqlParser.generateParseException(CCJSqlParser.java:26538) ~[jsqlparser-4.0.jar:na]
	at net.sf.jsqlparser.parser.CCJSqlParser.jj_consume_token(CCJSqlParser.java:26377) ~[jsqlparser-4.0.jar:na]
	at net.sf.jsqlparser.parser.CCJSqlParser.AndExpression(CCJSqlParser.java:8531) ~[jsqlparser-4.0.jar:na]
	at net.sf.jsqlparser.parser.CCJSqlParser.OrExpression(CCJSqlParser.java:8400) ~[jsqlparser-4.0.jar:na]
	at net.sf.jsqlparser.parser.CCJSqlParser.Expression(CCJSqlParser.java:8371) ~[jsqlparser-4.0.jar:na]
[...]

Issue #214, opened 2016-01-28
Not parsing the sql with "AT TIME ZONE" operator


chandansh:
I am parsing a query with "AT TIME ZONE" operator and getting the error.

query: select systimestamp AT TIME ZONE 'US/Eastern' from dual;

error: net.sf.jsqlparser.JSQLParserException
    at net.sf.jsqlparser.parser.CCJSqlParserUtil.parse(CCJSqlParserUtil.java:51)
    at com.sabre.apd.as.cem.ixintegration.ChandanTest1.main(ChandanTest1.java:56)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:606)
    at com.intellij.rt.execution.application.AppMain.main(AppMain.java:134)
Caused by: net.sf.jsqlparser.parser.ParseException: Encountered " <S_IDENTIFIER> "TIME "" at line 1, column 111.
Was expecting one of:
    <EOF> 
    "INTO" ...
    "FROM" ...
    "WHERE" ...
    "FOR" ...
    "GROUP" ...
    "HAVING" ...
    "START" ...
    "CONNECT" ...

PR #1196, opened 2021-05-14
Add support for AT TIME ZONE expressions

tomershay:
Adding support for AT TIME ZONE expressions, used for example in PostgreSQL.
Sample usages:
`SELECT creation_date AT TIME ZONE 'UTC';`
`SELECT DATE(date1 AT TIME ZONE 'UTC' AT TIME ZONE 'australia/sydney') AS another_date;`

This PR doesn't cover all time zone related options that can be done on an expression, but is only a start.

Documentation and samples (some of the capabilities described aren't yet supported):
https://www.postgresql.org/docs/9.2/functions-datetime.html#FUNCTIONS-DATETIME-ZONECONVERT
https://www.enterprisedb.com/postgres-tutorials/postgres-time-zone-explained

This PR potentially fixes or makes progress with issues: #214, #1115


tomershay:
@wumpz, can you please share if there is a way to add further tests to increase coverage for the uncovered parts? Didn't see a way to do that as part of the existing testing suite. Thanks!

manticore-projects:
Greetings @tomershay, I ran into the same issue a few days ago.

1) The Coveralls checks if each line of the Java Code has been executed/touched by your tests and the Details Report will show you which Java Code lines have NOT been touched

2) You will need to write tests, which trigger these Java Code Lines.

An over simplified example:
```java
private int foo(int i) {
    if (i>0) {
         return 1;
    } else {
         return 0;
   }
}

@test 
private void fooTest() {
    assertEquals(1, foo(1));
}
```

The test above will execute the `foo()` method but will not run through all code lines and Coveralls would complain, that `else return(0)` is never touched.

You would need to add another test, that triggers that part.

wumpz:
@tomershay Thx for adding this.

Issue #1190, opened 2021-05-10
can't solve sql:count_if(xxx='')


kfdjsakfdjk:
sql: select  count_if(app_code='xxx') as aa from tablename

note : count_if     ->  presto grammar

manticore-projects:
Greetings.

Thank for reporting.

The issue is NOT about `COUNT_IF` but about Functions having logical Expressions as arguments.
This example works (because there is no function involved):

```sql
SELECT a = '3'
FROM dual
;
```

But this example fails, because there is a Function involved:
``` sql
SELECT to_char(a = '3')
FROM dual
;
```

We will need to look into this eventually.

manticore-projects:
@wumpz : Pardon my naive question (as I am just an accountant): are we not affected here by this particular problem about "nested syntactic lookahead" which have been [fixed in JavaCC 21](https://javacc.com/2020/07/15/nested-syntactic-lookahead-works/)

If so, should we not try to adopt JavaCC 21?

wumpz:
Since it has a massive performance impact, I splittet the functions productions in one with normal parameters and one with complex expressions allowed. Only a hand full of functions are configured here by name. Therefore to add count_if should not be a problem.

About the nested syntactic lookahead I should take a look again.

manticore-projects:
> Since it has a massive performance impact, I splittet the functions productions in one with normal parameters and one with complex expressions allowed. Only a hand full of functions are configured here by name. Therefore to add count_if should not be a problem.

I feel like this jumps too short, for example:
```sql
SELECT to_char(a = '3')
FROM dual
;
```

is a valid statement which I would even use in order to show "true" or "false" in a report. And exactly this `to_char` function is prone to recursion. Why would we not consider to allow complex arguments on the first level of occurrence and then switch  to simple expression when going deeper?

wumpz:
This split is some kind of a hack. With this at least the most common used methods like **iif** wored. Sure it would be better to have a general approach. I tried this already using a production like ComplexExpression (I think that was the name). 

Your proposed approach is used in some occurrences and made already a huge performance gain. To be more precise I first accept a SimpleExpression and then look for a more complex one, like a relation or a condition.

manticore-projects:
On Mon, 2021-05-17 at 01:14 -0700, Tobias wrote:
> This split is some kind of a hack. With this at least the most common
> used methods like iif wored. Sure it would be better to have a
> general approach. I tried this already using a production like
[...]

Issue #1103, opened 2021-01-19
Parser hangs on more complicated queries


serge-rider:
It is similar to the #1013

Our users found another queries which cause JSQLParser to hang (https://github.com/dbeaver/dbeaver/issues/10680)
For example:

```sql
SELECT
ROUND(ROUND(ROUND(ROUND(ROUND(ROUND(ROUND(ROUND(
ROUND(ROUND(ROUND(ROUND(ROUND(ROUND(ROUND(ROUND(
ROUND(ROUND(ROUND(ROUND(ROUND(ROUND(ROUND(ROUND(
ROUND(ROUND(ROUND(ROUND(ROUND(ROUND(ROUND(ROUND(0
,0),0),0),0),0),0),0),0)
,0),0),0),0),0),0),0),0)
,0),0),0),0),0),0),0),0)
,0),0),0),0),0),0),0),0)
```

It cycles infinitely somewhere around
```
	  at net.sf.jsqlparser.parser.CCJSqlParser.jj_3R_RelObjectNameWithoutValue_1322_5_415(CCJSqlParser.java:19368)
	  at net.sf.jsqlparser.parser.CCJSqlParser.jj_3R_RelObjectName_1350_6_295(CCJSqlParser.java:19173)
	  at net.sf.jsqlparser.parser.CCJSqlParser.jj_3R_RelObjectName_1350_5_133(CCJSqlParser.java:19181)
	  at net.sf.jsqlparser.parser.CCJSqlParser.jj_3R_RelObjectNameExt_1380_7_308(CCJSqlParser.java:18983)
	  at net.sf.jsqlparser.parser.CCJSqlParser.jj_3R_RelObjectNameExt_1380_5_147(CCJSqlParser.java:19013)
	  at net.sf.jsqlparser.parser.CCJSqlParser.jj_3R_RelObjectNameList_1294_5_226(CCJSqlParser.java:19715)
	  at net.sf.jsqlparser.parser.CCJSqlParser.jj_3R_InternalFunction_3754_5_437(CCJSqlParser.java:22938)
	  at net.sf.jsqlparser.parser.CCJSqlParser.jj_3R_Function_3731_11_354(CCJSqlParser.java:22969)
	  at net.sf.jsqlparser.parser.CCJSqlParser.jj_3R_Function_3729_5_171(CCJSqlParser.java:22999)
[...]

PR #1200, opened 2021-05-20
Allow Complex Parsing of Functions

manticore-projects:
This PR will determine the nested brackets "(" ")" and configure the Parser in order to parse Complex Function Parameters like `to_char( a = b)` and Named Function Parameters like `trim(BOTH ' ' from 'foo bar ')` ONLY WHEN the level of nesting functions is below a threshold = 7. Although this is a configurable Parser Feature.

This massively speeds up the parser. The whole `NestedBracketsPerformanceTest` Suite takes less than 3 seconds now, with 50! iterations for the concat test.

At the same time we can parse Complex  Expression Parameters for all Functions (but only when there is no deep recursion of more than 9 levels although this detail is mood because it would not work in anyway because of the slowdown).

Fixes issues #1190 #1103

This is a **CLEAN** PR handcrafted against the latest Upstream MASTER, containing only the minimal changes.


manticore-projects:
LOL! The CI server is too slow for the performance test timeouts.

wumpz:
Building an upper boundary is quite a good idea. Could you please resolve the conflicts. I already merged your case when pr. My first changes for adaption the ValuesStatement are included as well. Maybe we could skip those MultiExpressionLists, because they are handles now by SimpleExpressonLists as well.

manticore-projects:
Merged the latest Master Branch.

wumpz:
So could you somehow better document, how the nesting depth is calculated or modified?

wumpz:
BTW I check your modifications. Since complex parsing is now allowed for functions I removed FunctionWithCondParams. But now your depth of seven is not enough for multiple tests. Unfortunatly.

manticore-projects:
> So could you somehow better document, how the nesting depth is calculated or modified?

It is a simple count of the maximum Open Bracket `(`. If there are more than 7 Open Brackets we assume deep nesting and do not allow complex parsing.

> BTW I check your modifications. Since complex parsing is now allowed for functions I removed FunctionWithCondParams. But now your depth of seven is not enough for multiple tests. Unfortunately I have problems for accepting one test.

Out of my head, I foresee a situation where you have a deep nesting on a function (without complex expressions) and one complex expression somewhere else (without deep nesting). That would fail because Complex Parsing and Deep Nesting are now mutual exclusive.

However, I kept experimenting when addressing the `CASE WHEN` expressions.
Instead of counting the Open Brackets, we can define a Nested Expression Counter in the Grammar File and increment it when ever we enter a `Primary Expression` and decrement it whenever we leave a `Primary Expression`. If this worked we would have much better granularity on when to give up `Complex Parsing`.

Please let me know your particular failing test case and I will have a look at it promptly.

manticore-projects:
@wumpz: You have modified the Brackets Threshold for Complex Parsing from 7 to 10. Unfortunately, the 7 has been selected for a purpose:
```
old duration 1565 new duration time 6873 for SELECT concat(concat(concat(concat(concat(concat(concat(concat(concat('A','B'),'B'),'B'),'B'),'B'),'B'),'B'),'B'),'B') FROM mytbl
```
It is the point when the performance impact LOOKAHEAD amplifies too much. This was also the reason why I have had defined `timeout` for the unit tests and I would like to recommend to leave it at 2000 and not to accept any change that breaks these constraints again. (2 seconds for parsing 1 single statement is too much already).

So, can we set the Threshold back to 7 again please? I am still working on a more fine granular solution, where the nesting is determined as per expression.

wumpz:
Understand. The problem is, you are calculating the "complexity" of a statement via running through the parenthesis. So you are taking parenthesises into account that have nothing to do with complex expressions. Somehow we should count the deepness within the productions. I don't know, if that is even possible and limit there to a level of 7 or 10.

manticore-projects:
It is not perfect, but it solves maybe 90% of the "real life" problems.
When my PRs are merged then I would like to count the real nesting in
the Primary Expression, same as I did for the CASE statement.

Problem is really that I have so many loose ends right now because the
acceptance is a bit slow.

wumpz:
Critics acknowledged. I try to improve.

wumpz:
Does this case counting work for lookaheads as well? If I remember right, this case deepness was never checked, or am I wrong?

manticore-projects:
On Wed, 2021-06-02 at 05:22 -0700, Tobias wrote:
> Does this case counting work for lookaheads as well?

Yes.

>  If I remember right, this case deepness was never checked, or am I
> wrong?
It was not needed for the cases but you see a similar preparation:

1) we have a global atomic LEVEL COUNT variable (this works well,
because the parser works strictly LL)
2) when entering a Primary Expression we count up
3) when leaving the Primary Expression we count down

This will give us the true nesting level.
We will stop for looking for Complex Expressions below a certain
threshold.

I am keen to implement and test and benchmark this better logic as soon
as everything else has been tied up.
After that we can also look again if JavaCC21 can help anything. or
not.

Cheers!

PR #1215, opened 2021-05-28
Pmd clean up

manticore-projects:
This is a CLEAN PR which containts only Annotations to switch off some of the Codazy/PMD rules regarding the Deparsers and Visitors.
Also added some Java Code Formatter Configuration.

No code changes.

### DIFF
--- a/nb-configuration.xml
+++ b/nb-configuration.xml
@@ -20,5 +20,14 @@
         <com-junichi11-netbeans-changelf.use-global>false</com-junichi11-netbeans-changelf.use-global>
         <org-netbeans-modules-javascript2-requirejs.enabled>true</org-netbeans-modules-javascript2-requirejs.enabled>
         <netbeans.hint.jdkPlatform>JDK_1.8</netbeans.hint.jdkPlatform>
+        <netbeans.checkstyle.format>true</netbeans.checkstyle.format>
+        <org-netbeans-modules-editor-indent.text.xml.CodeStyle.project.expand-tabs>false</org-netbeans-modules-editor-indent.text.xml.CodeStyle.project.expand-tabs>
+        <org-netbeans-modules-editor-indent.CodeStyle.project.text-line-wrap>none</org-netbeans-modules-editor-indent.CodeStyle.project.text-line-wrap>
+        <org-netbeans-modules-editor-indent.CodeStyle.project.indent-shift-width>4</org-netbeans-modules-editor-indent.CodeStyle.project.indent-shift-width>
+        <org-netbeans-modules-editor-indent.CodeStyle.project.spaces-per-tab>4</org-netbeans-modules-editor-indent.CodeStyle.project.spaces-per-tab>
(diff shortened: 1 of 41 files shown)


manticore-projects:
Summary of that PR in one chart:
image

nb-configuration.xml
```diff
@@ -20,5 +20,14 @@
         <com-junichi11-netbeans-changelf.use-global>false</com-junichi11-netbeans-changelf.use-global>
         <org-netbeans-modules-javascript2-requirejs.enabled>true</org-netbeans-modules-javascript2-requirejs.enabled>
         <netbeans.hint.jdkPlatform>JDK_1.8</netbeans.hint.jdkPlatform>
+        <netbeans.checkstyle.format>true</netbeans.checkstyle.format>
```
wumpz, line 23: Was necessary to introduce a project level formatting section? Why?

src/main/java/net/sf/jsqlparser/util/TablesNamesFinder.java
```diff
@@ -244,6 +244,7 @@ public void visit(IntegerDivision division) {
 
     @Override
     public void visit(DoubleValue doubleValue) {
+    // intentional (commented) empty methods.
```
wumpz, line 247: Which PMD rule forces this?

wumpz: I would prefer it without any comment. The name itself is documentation enough.

wumpz (changes requested):

src/main/java/net/sf/jsqlparser/util/TablesNamesFinder.java
manticore-projects: The rule `<rule ref="category/java/documentation.xml/UncommentedEmptyMethodBody" />` is applied by Codacy and will reject any commit affecting this class.
Although I can replace it by a Class wide Annotation.

nb-configuration.xml
manticore-projects: Removed.

src/main/java/net/sf/jsqlparser/util/TablesNamesFinder.java
wumpz: Damned. You are right. However, I will remove it from Codacy.

manticore-projects:
On Wed, 2021-06-02 at 05:14 -0700, Tobias wrote:
> Damned. You are right. However, I will remove it from Codacy.

Have a look at the project's PMD config. I marked the rules in palce
fro Codazy only and would like to suggest to remove these from Codazy.
It is not a bad thing at all, but some of the rules really create more
headache then helping.

src/main/java/net/sf/jsqlparser/util/TablesNamesFinder.java
wumpz: Removed.

manticore-projects: Yes, but I have solved it before already.
I have removed these comments already and you just need to accept the PR pls.

PR #1218, opened 2021-05-28
Implement GROUP BY () without columns

manticore-projects:
CLEAN PR against master with minimal change set.


wumpz:
Which database does support this? Whats the meaning of this? A **group by** that does nothing?

manticore-projects:
Fixes Issue #1176.

I did not want to believe it myself, but it is supported by Postgres and H2. I published the valid samples.
To me, H2 and Postgres are the Holy Grail, so please accept.

src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
```diff
@@ -2326,9 +2326,15 @@ GroupByElement GroupByColumnReferences():
 }
 {
     <K_GROUP> <K_BY>
-    (
-        columnReference=SimpleExpression() {groupBy.addGroupByExpression(columnReference); }
+    (     LOOKAHEAD(3) (
```
wumpz, line 2329: I changed a bit of logic in my latest commits. There is now a production SimpleExpressionList with and without brackets. Maybe we should change the group by expressions to an Expression List and simply put a SimpleExpressionList into it. This one should support both cases, but I am not sure about empty brackets.

wumpz (changes requested):
Only one question.

src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
manticore-projects: No problem, will look into that tonight. You will have it before you wake up.

manticore-projects: Delivered as requested:
```
<K_GROUP> <K_BY>
    (     LOOKAHEAD(2) (
              "(" ")" { groupBy.withUsingBrackets(true); }
          )
          | LOOKAHEAD(3) (
              "(" list = SimpleExpressionList(true)  ")" { groupBy.setGroupByExpressionList(list); }
          )
          |
          ( list = SimpleExpressionList(false) { groupBy.setGroupByExpressionList(list); }
          )
    |
```

Isn't it a beauty?

manticore-projects: Btw, SimpleExpressionLists can't be empty and always expects at least 1 element in the Grammar Definition. I did not want to refactor this right now, but I think we should look for making it a normal list which could be empty.

PR #1224, opened 2021-06-10
Delete queries without from, with a schema identifier fails

fanchuo:
Hello,

It is currently possible to analyze this query:

delete a where z=1

However, it fails if the schema is determined:

delete a.b where z=1

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/delete/Delete.java
+++ b/src/main/java/net/sf/jsqlparser/statement/delete/Delete.java
@@ -37,6 +37,7 @@ public class Delete implements Statement {
     private Expression where;
     private Limit limit;
     private List<OrderByElement> orderByElements;
+    private boolean hasFrom = true;
     public List<WithItem> getWithItemsList() {
         return withItemsList;
     }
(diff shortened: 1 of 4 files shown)


manticore-projects (commented):
It looks to me, like the `DELETE WITHOUT FROM` deparses into a `DELETE WITH FROM`?
Should we not deparse into exactle the provided statement?

Which RDBMS allows `DELETE WITHOUT FROM`?

francois-secherre:
Hello,

I ran into such a query yesterday while working in AWS Redshift. Though I remember it works on Oracle, by the days I used it. It also appears to run OK for T-SQL:

https://docs.oracle.com/cd/B19306_01/server.102/b14200/statements_8005.htm
https://docs.microsoft.com/fr-fr/sql/t-sql/statements/delete-transact-sql?view=sql-server-ver15
https://stackoverflow.com/questions/4484250/in-sql-syntax-is-from-in-delete-from-optional-if-you-plan-to-use-where

In both case, FROM is considered optional, and delete table where ... is equivalent to delete from table where ...

I am not sure it answers your questions.

I am perfectly aware this is not standard SQL, but as as I check your grammar, the FROM keyword have been optional since the first release. It accepts this syntax as long as the table schema is not specified (lookahead(2) checks only next literal), and it deparses with a FROM keyword.

So tell me:
- I can make the FROM mandatory to the parser
- I can make it deparse without the FROM keyword

manticore-projects:
All fine, I was not aware that Oracle and MS SQL Server support `DELETE WITHHOUT FROM` although the Grammar treated it as optional already.
However, in my opinion, the deparser should reflect that accordingly,

manticore-projects:
Just looked it up fro Oracle:
image

wumpz:
Thx for improving JSqlParser.

PR #1228, opened 2021-06-14
Support DELETE FROM T1 USING T2 WHERE ...

francois-secherre:
Support for DELETE FROM USING queries, from postgres

https://www.postgresql.org/docs/11/sql-delete.html

Example:

DELETE FROM films USING producers
  WHERE producer_id = producers.id AND producers.name = 'foo';


wumpz:
Thx for improving JSqlParser.

PR #1241, opened 2021-06-25
CROSS JOIN is used without conditions

LonwoLonwo:
Thanks for the great library!

I found a small issue, which prevents me from generating statements with CROSS JOINs. 
This happens when I first have any other JOIN with conditions, but then I want to change it on Cross JOIN.

I know that the tests are welcome here, but the CROSS JOIN test is already in the code, so I didn't add it.

Thanks for your attention.


LonwoLonwo:
Sorry. Change my mind. It turns out, SQLite support cross joins with conditions.

PR #1271, opened 2021-07-17
Implement Gradle Buildsystem

manticore-projects:
Advantages of the Gradle Build:
1) well maintained JavaCC plugin which allows to set DEBUG during the build (instead of changing the Grammar JJT file)
2) much simpler build file (also easier to read)
3) Railroad diagrams per RR
4) better checks via JaCoCo, PMD and SpotBugs
    a) JaCoCo reports now Test Coverage before Push and also illustrates the uncovered parts directly in the Netbeans IDE. Reports are activated.
    b) SpotBugs revealed one Prio 1 issue, I have fixed that already. I deactived any Prio 2 issue for the time being, but we should look at these later. Reports are activated.
    c) PMD also revealed 6 Prio 1 issues (not covered by our custom rules). I deactivated these for the time being, but we should look at these later. Reports are activated.
5) Gradle builds much faster and provides better output

I kept the Maven Build in place and right  now you can choose if you build per Maven or per Gradle.
The "Deploy to Maven Central/Sonartype" is prepared but I have not been able to test it yet (because I do not have the credentials/access)

I have not  provided a Website Build yet, but have prepared a Sphinx Skeleton which would show:
1) the Readme.md
2) Change Log
3) the Railroad Diagrams
4) the Java API
5) the Code Quality Reports
I would like to conclude it when you have digested/agree on the new infrastructure.


manticore-projects:
Netbeans will prefer the Gradle Build over the Mavens build.

Below is the JaCoCo test coverage within Netbeans, I love this stuff!
image

manticore-projects:
Ups! I just noticed that the TRAVIS CI will automatically prefer the Gradle as well, I have not been expecting that.

wumpz:
image

image

Same with the maven build.  I just forgot to include the **jacoco** plugin again. 


What really is compelling is the debug setting of the grammar. 

However, I will give it a try. Gradle is not my turf you know. 

But why are 285 files changed?

manticore-projects:
> Same with the maven build. I just forgot to include the **jacoco** plugin again.

Good, then we we will have it one way or the other.

> What really is compelling is the debug setting of the grammar.
Yes, this one and the fact that the JavaCC plugin is maintained and consists of simple Groovy code.
Also, the RR integrations shows, how easy and simple it was to executed other Jar files (e. g. if we ever use a JavaCC without an available plugin).

When I tried to activate JJDoc with the Maven Plugin it was a real struggle. All documentation dead or stale.
Lets face it: on the long run we will need something more future proof.
 
> However, I will give it a try. Gradle is not my turf you know.

Same here. I never touched Gradle before and was actually scared because I thought it might be complex.
But not at all. It feels very natural, almost Java like. Please just give it a try.

> 
> But why are 285 files changed?

I moved all the Special Oracle Test SQL Files into the correct folder according to the package of the test class, when I had to adjust the hard coded location in the Test Class (from "target/..." to "build/..." for Gradle).

manticore-projects:
I'd like to add a personal note: My own programs are still running on ANT and I am new to Maven and Gradle. I touched Maven before Gradle for the sake of working on JSQLParser and POI and H2.

That said, I liked Gradle instantly and will migrate my own projects now.
I never really got warm with Maven though.

wumpz:
I looked into it and indeed the maven javacc plugin we are using is quite old. But as those javacc people introduced a new gradle plugin, they did with the maven plugin with the advantages you mentioned (IMHO). 

```xml
<!-- https://mvnrepository.com/artifact/org.javacc.plugin/javacc-maven-plugin -->
<dependency>
    <groupId>org.javacc.plugin</groupId>
    <artifactId>javacc-maven-plugin</artifactId>
    <version>3.0.3</version>
</dependency>
```

I will integrate this one, so we have a direct comparision.

I have my problems with gradle to be honest. But my experiences were some years old. So I will give it a try.

manticore-projects:
On Sun, 2021-07-18 at 23:25 -0700, Tobias wrote:
> I will integrate this one, so we have a direct comparision.
I think it boils down to 3 considerations

1) JavaCC support (and the JaCoCo Reports integration) <-- You have
solved that, they are on par.
2) Integration of any (Java-) programs without plugins (the RR
diagrams) <-- Certainly works in both, but I found Gradle much simpler
3) Simplicity and Build Speed <-- I found Gradle better in that regard,
especially for repeating builds
> I have my problems with gradle to be honest. But my experiences were
> some years old. So I will give it a try.
Fair enough, I was really not expecting you to be too excited about
replacing the build system.
And again, it is not the hill I want to die on as long as we get
everything working with Maven. The main reason I started to work on
that was 2) the RR diagrams.

Thank you for looking at it, I appreciate.
Cheers!

PR #1279, opened 2021-07-21
Special oracle tests

### LINKED ISSUE #1201: SpecialOracleTest currently creates a lot of noise
During Build/Testing some Oracle related tests are executed and are expected/accepted to fail. For example
```
May 20, 2021 6:22:12 AM net.sf.jsqlparser.statement.select.SpecialOracleTest testAllSqlsParseDeparse
INFO: testing for_update07.sql
net.sf.jsqlparser.JSQLParserException: Encountered unexpected token: "," ","
    at line 11, column 19.

Was expecting one of:

    ";"
    "FOR"
    "NOWAIT"
    "ORDER"
    <EOF>
```

Would it not be better to quieten or deactivate these tests, until we have time to look into these?

manticore-projects:
The SpecialOracle Test succeeded  when a new feature fixed more samples than it broke.
We lost `GROUPING()`, `LINK` this way. 

Furthermore it was very difficult to read a particular Test Case and to understand, what was actually expected and working or not.
Last but not least, this test was very noisy.

Implemented Changes:
1) Implemented `GROUPING()`, `LINK` keywords correctly (this fixed a bunch of tests)
2) Implemented `ParenthesisFromItem` `Pivot` and `UnPivot` correctly (this fixed a few more tests)
3) Write the expected Test Result directly into the Test SQL source file
4) Updated the List of Tests expected to succeed and will fail when any of these suddenly fails (no sacrifices please!)
5) Warn only when anything unexpected happens (e. g. Newly achieved Success)

In the result, now 186 out of 272 tests succeed (almost 10 more than before).
Fixes #1201.


manticore-projects:
Something is still wrong with the `SelectDeparser` and `ParenthesisFromItem` `Pivots` or `UnPivots`:
- `Pivot` seems to be deparsed already somewhere else
- `UnPivot` is not deparsed somewhere else and so needed to be added (although I likely added them at the wrong place)

Right now, I got it working *somehow* but someone else should have an eye onto it.
I do not work with the `Visitor` patterns, they are confusing me.

manticore-projects:
The change works beautifully: When synchronizing it against the merged `PURGE` PR, suddenly an exception about `bindvar05.sql` has been thrown and pointed me on declaring `RECEYCLEBIN` as valid Object Name.

This has not been detected in `PURGE` PR itself, where it belonged to.

wumpz:
Is this PR still valid with all these conflicts?

manticore-projects:
> Is this PR still valid with all these conflicts?

Very valid and beautiful :-D
We could add similar SQL Batch tests for other RDBMS as well with this improved infrastructure.

PR #1303, opened 2021-08-14
Implement Joins with multiple trailing ON Expressions

### LINKED ISSUE #1229: join missing the 'on' filter
original SQL : select t1.column1,t1.column2,t2.field1,t2.field2 from T_DT_ytb_01 t1 , T_DT_ytb_02 t2 on t1.column1 = t2.field1
after parse SQL:   SELECT t1.column1, t1.column2, t2.field1, t2.field2 FROM T_DT_ytb_01 t1, T_DT_ytb_02 t2

What is simple join?

### LINKED ISSUE #1302: Can't parse with query inner join
I have statement :
SELECT * FROM TABLE1 tb1 
	INNER JOIN TABLE2 tb2
		INNER JOIN TABLE3 tb3
			INNER JOIN TABLE4 tb4
			ON (tb3.aaa = tb4.aaa)
		ON (tb2.aaa = tb3.aaa)
	ON (tb1.aaa = tb2.aaa)

I try :
Select select = (Select) CCJSqlParserUtil.parse("SELECT * FROM TABLE1 tb1 "
				+ "INNER JOIN TABLE2 tb2 "
				+ "INNER JOIN TABLE3 tb3 "
				+ "INNER JOIN TABLE4 tb4 "
				+ "ON (tb3.aaa = tb4.aaa) "
				+ "ON (tb2.aaa = tb3.aaa) "
				+ "ON (tb1.aaa = tb2.aaa)");

**Expected behavior**
Exception in thread "main" net.sf.jsqlparser.JSQLParserException: Encountered unexpected token: "ON" "ON"
    at line 1, column 115.
**System**
 - Database you are using : Oracle
- Java Version : 1.8
- JSqlParser version : 4.1

manticore-projects:
Fixes #1302
Fixes #1229
Fixes SpecialOracleTest JOIN17, now 190/273 tests pass


src/main/java/net/sf/jsqlparser/statement/select/Join.java
```diff
@@ -278,49 +294,58 @@ public void setJoinWindow(KSQLJoinWindow joinWindow) {
...
+        if (usingColumns.size()>0) {
+            builder.append(PlainSelect.getFormatedList(usingColumns, "USING", true, true));
         }
 
+        return builder;
+    }
+
+    @Override
+    public String toString() {
+        return appendTo(new StringBuilder()).toString();
```
wumpz, line 348: Again. Whats the improvement of this type of generating a StringBuilder and populating it? IMHO this is no improvement over this older toString approach.

wumpz (commented):
I am not convinced that this StringBuilder approach is the way to go. What's the point?

src/main/java/net/sf/jsqlparser/statement/select/Join.java
manticore-projects: Good Morning. Thank you for your feedback.
My thought process was like this:

1. I have had to touch `toString()` because of the new Loop through the ON-Expressions
2. String Addition `+=` in Loops is still considered bad style and will be called out by many style checkers, because the Compiler **won't** substitute the concatenation by a StringBuilder
3. providing an `appendTo()` method consistently for all Statements and Expressions would eventually allow the Deparser to work with one single StringBuilder only (imitated at the top). Right many Statements and Expressions initiate and dispose their own StringBuilder putting useless load onto the GC (think about deep nested expressions especially).
4. I do not see any disadvantages of this pattern. `toString()` still works without any caveats. We just delegate the logic into a special method. You did the same for many other Statements and Expressions already -- just not calling it `appendTo()` and not applying a systematic pattern.

What counter arguments do you see? Why was the String concatenation better, especially with the newly needed Loops?

Cheers.

manticore-projects: I have removed the `appendTo()` method and re-implemented the traditional `toString()` as you obviously do not like this approach.

PR #1309, opened 2021-08-18
CREATE TABLE AS (...) UNION (...) fails

fanchuo:
This test query is failing so far:
CREATE TABLE temp.abc AS (SELECT c FROM t1) UNION (SELECT c FROM t2)


manticore-projects:
Thank you so much, I have incorporated your suggestions into PR #1329 -- but attributed all credits to you.
I chose this path when the Travis CI Test kept failing.

PR #1334, opened 2021-09-09
Postgres compliant ALTER TABLE ... RENAME TO ...

### LINKED ISSUE #1323: Validation of a statement for renaming a table fails with ParseException: Cannot parse statement: Encountered unexpected token: "to" "TO"
My code:
```
	@Test
	public void testJSQParser()
	{
		String sql = "alter table public.table2 rename to table3;";
		Validation validation = new Validation(List.of(DatabaseType.POSTGRESQL), sql);
		List<ValidationError> errors = validation.validate();
		Assert.assertTrue(errors.toString(), errors.isEmpty());
	}
```
Result:
```
java.lang.AssertionError: [ValidationError [
statement=alter table public.table2 rename to table3;
capability=parsing
errors=[ParseException: Cannot parse statement: Encountered unexpected token: "to" "TO"
    at line 1, column 34.

Was expecting:

    "COLUMN"
]
]]
```
Same with Oracle. This statement can be executed in both.

### LINKED ISSUE #1333: No support for alter expressions of renaming tables
In PostgreSQL, the only supported table renaming syntax is:
ALTER TABLE  [IF EXISTS] [<schema_name>.]<old_table_name> RENAME TO <new_table_name>

Same for Oracle, when renaming tables across schemas, excluding the IF EXISTS option.

When trying to validate such expression, you get the error detailed in #1323

manticore-projects:
Postgres compliant `ALTER TABLE schema.oldTableName RENAME TO newTableName`
Fixes #1333
Fixes #1323

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java
+++ b/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java
@@ -32,6 +32,7 @@ public class AlterExpression {
 
   private AlterOperation operation;
   private String optionalSpecifier;
+  private String newTableName;
   private String columnName;
   private String columnOldName;
   // private ColDataType dataType;
@@ -246,6 +247,14 @@ public class AlterExpression {
     this.fkSourceColumns = fkSourceColumns;
   }
 
+  public String getNewTableName() {
+    return newTableName;
+  }
+
+  public void setNewTableName(String newTableName) {
+    this.newTableName = newTableName;
+  }
+
   public String getColumnName() {
     return columnName;
   }
@@ -367,6 +376,8 @@ public class AlterExpression {
     
     if (operation== AlterOperation.UNSPECIFIC) {
         b.append(optionalSpecifier);
+    } else if (operation== AlterOperation.RENAME_TABLE) {
+      b.append("RENAME TO ").append(newTableName);
     } else {
         b.append(operation).append(" ");
 
--- a/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java
+++ b/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java
@@ -45,7 +45,7 @@ public class AlterExpression {
   private String ukName;
   private Index index = null;
   private String constraintName;
-  private boolean constraintIfExists;
+  private boolean usingIfExists;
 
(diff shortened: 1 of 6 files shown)


src/test/java/net/sf/jsqlparser/statement/alter/RenameTableStatementTest.java
```diff
@@ -88,8 +88,22 @@ public void testTableNamesFinder() throws JSQLParserException {
...
+      // this needs to succeed
+      ValidationTestAsserts.validateNoErrors(sqlStr, 1, DatabaseType.ORACLE);
+
+      // this should fail when not supported in Postgres
+      ValidationTestAsserts.validateNoErrors(sqlStr, 1, DatabaseType.POSTGRESQL);
+
+      sqlStr = "ALTER TABLE public.oldTableName RENAME TO newTableName";
+      TestUtils.assertSqlCanBeParsedAndDeparsed(sqlStr, true);
+
+      //  should fail when not supported in Oracle
```
AmitAmely, line 102: Both syntaxes work in Oracle, and specifically, when trying to rename a table on a different schema, only the ALTER syntax works.

manticore-projects: Thanks, I have reworked the comments although the works remains on the Validator.

src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java
```diff
@@ -44,7 +45,7 @@
   private String ukName;
   private Index index = null;
   private String constraintName;
-  private boolean constraintIfExists;
+  private boolean usingIfExists;
```
AmitAmely, line 48: Did you consider adding this field to the Alter class instead of merging it with constraintIfExist? that would also make the deparsing much more simple

manticore-projects: Thank you for your feedback, you are right of course: The whole `ALTER TABLE ...` statement needs to get overhauled -- but I did not have a time for this yesterday and just rushed a fix to make it work for you.

At the same time you are very knowledgeable and would make a great contributor. I stand ready to provide you with any assistance for a first PR.

PR #1370, opened 2021-10-13
Fixes #1369

### LINKED ISSUE #1369: Unable to get subselect from WithItem when it is null
**Describe the bug**
WithItem currently tries to set brackets to false when getting the subselect. If the subselect is null, this call fails as an NPE

**To Reproduce**
1. Example SQL
```
WITH data AS (VALUES ('1'), ('2')) SELECT * FROM data
```
fails
3. Exception
```
java.lang.NullPointerException: Cannot invoke "net.sf.jsqlparser.statement.select.SubSelect.withUseBrackets(boolean)" because "this.subSelect" is null
        at net.sf.jsqlparser.statement.select.WithItem.getSubSelect(WithItem.java:92)
```

**Expected behavior**
For the call to not fail and return null if it is not present

**System**
 - Database you are using Postgresql
- Java Version 11
- JSqlParser version 4.1

SerialVelocity:
The `get` method should return null instead of throwing a NullPointerException

Fixes #1369

### DIFF
--- a/src/test/java/net/sf/jsqlparser/statement/select/SelectTest.java
+++ b/src/test/java/net/sf/jsqlparser/statement/select/SelectTest.java
@@ -4830,7 +4830,7 @@ public class SelectTest {
         assertSqlCanBeParsedAndDeparsed(
                 "SELECT fulltext from fulltext.fulltext", true);
     }
-  
+
     @Test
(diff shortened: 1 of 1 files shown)


manticore-projects:
Good first contribution, thank you a lot. Kudos!

Issue #1254, opened 2021-07-05
Update ignore and delete ignore not parseable


Todd102030:
**Describe the bug**
Update ignore and delete ignore statements throw exception when parsing.

**To Reproduce**
1. Example SQL
update ignore tRequirement set VendorItem=? where cn=@cn and VendorItem=?

2. Parsing this SQL using JSqlParser with this statements
SimpleNode node = (SimpleNode) CCJSqlParserUtil.parseAST(sqlInput);

3. Exception
Encountered unexpected token: "ignore" "IGNORE"
    at line 1, column 8.

**Expected behavior**
Expected to parse similarly to statements without the "ignore" keyword

**System**
 - MySQL 8.0
- Java Version - JDK 8
- JSqlParser version - 4.1

Thanks for all your work on this library, it's really great!

wumpz:
So this is only about *update* statements? PRs are welcome. Grammar line 950 should be the start.

PR #1382, opened 2021-10-18
Enhanced Keywords

### LINKED ISSUE #1148: Unexpected token: apply
**Describe the bug**
There is a sql which i can run with mysql success, but if with jsqlparser, got an exception

**To Reproduce**
1. Example SQL
```sql
 SELECT  remark,apply FROM oms_data
```

2. Exception Log
```
2021-04-07 09:55:02.471  WARN 41488 --- [nio-8080-exec-1] c.b.m.e.p.i.PaginationInnerInterceptor   : optimize this sql to a count sql has exception, sql:"SELECT  
remark,apply
  FROM oms_data", exception:
net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: "apply" "APPLY"
    at line 2, column 8.

Was expecting one of:

    "!"
    "("
    "*"
    "+"
    "-"
    ":"
    "?"
    "@"
    "@@"
    "ACTION"
    "ANY"
    "BYTE"
    "CASCADE"
    "CASE"
    "CAST"
    "CHANGE"
    "CHAR"
    "CHARACTER"
    "COLUMN"
    "COLUMNS"
    "COMMENT"
    "COMMIT"
    "CREATE"
    "CURRENT"
    "CYCLE"
    "DESCRIBE"
    "DISABLE"
    "DO"
    "DOUBLE"
    "DUPLICATE"
    "ENABLE"
    "END"
    "EXTRACT"
    "FALSE"
    "FIRST"
    "FN"
    "FOLLOWING"
    "GROUP"
    "GROUP_CONCAT"
    "IF"
    "IIF"
    "INDEX"
    "INSERT"
    "INTERVAL"
    "ISNULL"
    "KEY"
    "LAST"
    "LEFT"
    "LIMIT"
    "MATCH"
    "MATERIALIZED"
    "NEXTVAL"
    "NO"
    "NOLOCK"
    "NOT"
    "NULL"
    "NULLS"
    "OFFSET"
    "ON"
    "OPEN"
    "OPTIMIZE"
    "ORDER"
    "OVER"
    "PARTITION"
    "PATH"
    "PERCENT"
    "PRECISION"
    "PRIMARY"
    "PRIOR"
    "PROCEDURE"
    "RANGE"
    "READ"
    "REPLACE"
    "RIGHT"
    "ROW"
    "ROWS"
    "SCHEMA"
    "SEPARATOR"
    "SESSION"
    "SET"
    "SIBLINGS"
    "SIZE"
    "START"
    "TABLE"
    "TEMP"
    "TEMPORARY"
    "TOP"
    "TRUE"
    "TRUNCATE"
    "TYPE"
    "UNSIGNED"
    "VALIDATE"
    "VALUE"
    "VALUES"
    "VIEW"
    "XML"
    "ZONE"
    "{"
    "{d"
    "{t"
    "{ts"
    "~"
    <K_DATETIMELITERAL>
    <K_DATE_LITERAL>
    <K_TIME_KEY_EXPR>
    <S_CHAR_LITERAL>
    <S_DOUBLE>
    <S_HEX>
    <S_IDENTIFIER>
    <S_LONG>
    <S_QUOTED_IDENTIFIER>
```

**Expected behavior**

**System**
 - Database you are using: `MySQL 5.6.16`
- Java Version: `Oracle 1.8.0_251`
- JSqlParser version: `3.2`

### LINKED ISSUE #1443: 4.3Release：Error SQL:select database() from dual
**DB : mysql5.7**
**JSqlParser : 4.3**

```
        String sql = "select database() from dual";
        try {
            Statement statement = CCJSqlParserUtil.parse(sql);
            System.out.println(statement.toString());
        } catch (JSQLParserException e) {
            throw new RuntimeException(e.getCause() + "Failed to process, Error SQL:" + sql);
        }
    }
```

run result:
```
Exception in thread "main" java.lang.RuntimeException: net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: "database" "DATABASE"
    at line 1, column 8.

Was expecting one of:

    "*"
    "STRAIGHT_JOIN"
Failed to process, Error SQL:select database() from dual
	at com.ruoyi.common.Test.main(Test.java:15)
```

Greetings.

**this bug not fixed on release-4.3, but Why does it work normally with 4.3-Snapshot ?**

~~~~~~~~~

### LINKED ISSUE #1450: select database()
**Describe the bug**
When I want to get the currently used database, I find that it cannot be parsed

**To Reproduce**
1. Example SQL
`String sql = "select database()";`
`CCJSqlParserUtil.parse(sql);`
3. Exception
```
Encountered unexpected token: "database" "DATABASE"
    at line 1, column 8.

Was expecting one of:

    "*"
    "STRAIGHT_JOIN"
```

**Expected behavior**

**System**
 - mysql 8.0.27
- jdk-liberica-11.0.13
- 4.2 or 4.3

### LINKED ISSUE #1462: CCJSqlParserUtil fails on different variable names
**Describe the bug**
When I want to parse the following query 
```sql
select p from Customer c join c.productOrder p where p.delayed = true
```
using `(Select) CCJSqlParserUtil.parse(query);` I get the following error:

```
net.sf.jsqlparser.JSQLParserException: Encountered unexpected token: "." "."
    at line 1, column 55.

Was expecting one of:

    "&"
    "::"
    ";"
    "<<"
    ">>"
    "COLLATE"
    "CONNECT"
    "EMIT"
    "GROUP"
    "HAVING"
    "START"
    "["
    "^"
    "|"
    <EOF>

	at net.sf.jsqlparser.parser.CCJSqlParserUtil.parseStatement(CCJSqlParserUtil.java:190)
	at net.sf.jsqlparser.parser.CCJSqlParserUtil.parse(CCJSqlParserUtil.java:63)
	at net.sf.jsqlparser.parser.CCJSqlParserUtil.parse(CCJSqlParserUtil.java:38)
	at org.springframework.data.jpa.repository.query.JSqlParserQueryUtils.parseSelectStatement(JSqlParserQueryUtils.java:111)
	... 71 more
```

Although when I try to parse the following query 
```sql
select p from Customer c join c.productOrder p where p.delaye = true
```
everything works as expected.

It is quite stragne that only the difference between `delayed` and `delaye` makes the parsing to fail. Maybe I am getting something wrong but I think an identifier should not make such a huge difference.

**To Reproduce**
See above

**Expected behavior**
Both queries should be parse without failing.

**System**
- Database you are using: No DB just calling `CCJSqlParserUtil.parse(query)`
- Java Version: `Azul Zulu 1.8.0_322`
- JSqlParser version: 4.3.0

### LINKED ISSUE #1508: where table_name = select database() has exception
**Describe the bug**
paser `where table_name = select database()` has exception

**To Reproduce Example**
```
       public static void main(String[] args) throws JSQLParserException {
        String sql = " select table_name, table_comment, create_time, update_time from information_schema.tables " +
                " where table_schema = (select database()) ";
        Statement statement = CCJSqlParserUtil.parse(sql);
        System.out.println(statement.toString());
    }

```

**Exception out**
```
Exception in thread "main" net.sf.jsqlparser.JSQLParserException: Encountered unexpected token: "=" "="
    at line 1, column 112.

Was expecting one of:

    "&"
    "::"
    ";"
    "<<"
    ">>"
    "COLLATE"
    "CONNECT"
    "EMIT"
    "GROUP"
    "HAVING"
    "START"
    "["
    "^"
    "|"
    <EOF>

	at net.sf.jsqlparser.parser.CCJSqlParserUtil.parseStatement(CCJSqlParserUtil.java:190)
	at net.sf.jsqlparser.parser.CCJSqlParserUtil.parse(CCJSqlParserUtil.java:63)
	at net.sf.jsqlparser.parser.CCJSqlParserUtil.parse(CCJSqlParserUtil.java:38)
	at com.seerbigdata.tourism.common.config.MybatisPlusConfig.main(MybatisPlusConfig.java:158)
Caused by: net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: "=" "="
    at line 1, column 112.

Was expecting one of:

    "&"
    "::"
    ";"
    "<<"
    ">>"
    "COLLATE"
    "CONNECT"
    "EMIT"
    "GROUP"
    "HAVING"
    "START"
    "["
    "^"
    "|"
    <EOF>

	at net.sf.jsqlparser.parser.CCJSqlParser.generateParseException(CCJSqlParser.java:31468)
	at net.sf.jsqlparser.parser.CCJSqlParser.jj_consume_token(CCJSqlParser.java:31301)
	at net.sf.jsqlparser.parser.CCJSqlParser.Statement(CCJSqlParser.java:163)
	at net.sf.jsqlparser.parser.CCJSqlParserUtil.parseStatement(CCJSqlParserUtil.java:188)
	... 3 more
```

**System**
- MySQL-v8.0.27
- java version "1.8.0_31"
- JSqlParser 4.4

### LINKED ISSUE #1538: `select tbl.increment from table tbl` throw JSqlParserException
* Database: Mysql 5.7
* JsqlParser: 3.2

Below is the exception:
```
net.sf.jsqlparser.JSQLParserException
	at net.sf.jsqlparser.parser.CCJSqlParserUtil.parse(CCJSqlParserUtil.java:65)
	at net.sf.jsqlparser.parser.CCJSqlParserUtil.parse(CCJSqlParserUtil.java:40)
	at com.zuora.poc.dac.SqlParserPerfTest.testParseFrom(SqlParserPerfTest.java:62)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:69)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater$1.execute(IdeaTestRunner.java:38)
	at com.intellij.rt.execution.junit.TestsRepeater.repeat(TestsRepeater.java:11)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:35)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:235)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:54)
Caused by: net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: "." "."
    at line 1, column 11.

Was expecting one of:

    "&"
    ","
    "::"
    ";"
    "<<"
    ">>"
    "ACTION"
    "ANY"
    "AS"
    "BYTE"
    "CASCADE"
    "CAST"
    "CHANGE"
    "CHAR"
    "CHARACTER"
    "COLLATE"
    "COLUMN"
    "COLUMNS"
    "COMMENT"
    "COMMIT"
    "CONNECT"
    "CYCLE"
    "DESCRIBE"
    "DISABLE"
    "DO"
    "DUPLICATE"
    "ENABLE"
    "END"
    "EXCEPT"
    "EXTRACT"
    "FALSE"
    "FIRST"
    "FN"
    "FOLLOWING"
    "FOR"
    "FROM"
    "GROUP"
    "HAVING"
    "INDEX"
    "INSERT"
    "INTERSECT"
    "INTERVAL"
    "INTO"
    "ISNULL"
    "KEY"
    "LAST"
    "MATERIALIZED"
    "MINUS"
    "NEXTVAL"
    "NO"
    "NOLOCK"
    "NULLS"
    "OPEN"
    "ORDER"
    "OVER"
    "PARTITION"
    "PATH"
    "PERCENT"
    "PRECISION"
    "PRIMARY"
    "PRIOR"
    "RANGE"
    "READ"
    "REPLACE"
    "ROW"
    "ROWS"
    "SCHEMA"
    "SEPARATOR"
    "SESSION"
    "SIBLINGS"
    "SIZE"
    "START"
    "TABLE"
    "TEMP"
    "TEMPORARY"
    "TOP"
    "TRUE"
    "TRUNCATE"
    "TYPE"
    "UNION"
    "UNSIGNED"
    "VALIDATE"
    "VALUE"
    "VALUES"
    "VIEW"
    "WHERE"
    "WINDOW"
    "XML"
    "ZONE"
    "["
    "^"
    "|"
    <EOF>
    <K_DATETIMELITERAL>
    <K_DATE_LITERAL>
    <S_CHAR_LITERAL>
    <S_IDENTIFIER>
    <S_QUOTED_IDENTIFIER>

	at net.sf.jsqlparser.parser.CCJSqlParser.generateParseException(CCJSqlParser.java:25031)
	at net.sf.jsqlparser.parser.CCJSqlParser.jj_consume_token(CCJSqlParser.java:24875)
	at net.sf.jsqlparser.parser.CCJSqlParser.Statement(CCJSqlParser.java:91)
	at net.sf.jsqlparser.parser.CCJSqlParserUtil.parse(CCJSqlParserUtil.java:63)
	... 26 more
```

I guess the `increment` is treated as a keyword?

### LINKED ISSUE #1650: JSqlParser can't parse DATABASE function in mysql.
```sql
        select count(*)
        from information_schema.columns
        where table_name = 'extend_value_info'
          and column_name = 'extend_value_str'
          and table_schema = DATABASE();
```
throws:
```
Caused by: net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: "=" "="
    at line 5, column 28.

Was expecting one of:

    "&"
    "::"
    ";"
    "<<"
    ">>"
    "COLLATE"
    "CONNECT"
    "EMIT"
    "GROUP"
    "HAVING"
    "START"
    "["
    "^"
    "|"
    <EOF>

	at net.sf.jsqlparser.parser.CCJSqlParser.generateParseException(CCJSqlParser.java:33398) ~[jsqlparser-4.5.jar:na]
	at net.sf.jsqlparser.parser.CCJSqlParser.jj_consume_token(CCJSqlParser.java:33231) ~[jsqlparser-4.5.jar:na]
	at net.sf.jsqlparser.parser.CCJSqlParser.Statement(CCJSqlParser.java:167) ~[jsqlparser-4.5.jar:na]
	at net.sf.jsqlparser.parser.CCJSqlParserManager.parse(CCJSqlParserManager.java:23) ~[jsqlparser-4.5.jar:na]
	... 54 common frames omitted
```
Is there any workround?
I try to around the DATABASE with "``" like this:
```sql
        select count(*)
        from information_schema.columns
        where table_name = 'extend_value_info'
          and column_name = 'extend_value_str'
          and table_schema = `DATABASE()`;
```
Then mysql can't parse it.

manticore-projects:
Rewrite the Reserved Keywords Logic:
1) Automatically Derive all Keyword Tokens from the Parser.
2) Explicitly add Reserved Keywords to the Grammar and classify, why/when it is reserved.
3) Auto-generate the list of Allowed Keywords as Difference of 1) and 2) above
4) Provide Gradle task to  semi-auto generate `RelObjectNameWithoutValue()` based on 3)
5) Parametrized Tests for Keywords, testing all keywords

Advantage: 
a) we have now a clear documentation which Keywords are reserved for what reason, with proper tests
b) allowed keywords are generated (semi-)automatically, especially when more Tokens are added to the Grammar. New Tokens won't break SQL Statements, which have been parsed fine before.

To-Dos:
c) Composite Tokens do not work well and still need to be added manually to ALL_KEYWORDS (we would need to refactor the GRAMMAR in order to avoid such composite tokens)
d) @wumpz: Clarify the meaning/purpose of the various `RelObjectNamexxxx()` Productions, so we can auto generate them too.
e) Use the Gradle task to fully inject the `RelObjectNamexxxx()` Productions (instead of manually updating the code)

Resolves one more Special Oracle Test.
Fixes #1148
Fixes #1450
Fixes #1443
Fixes #1462
Fixes #1508
Fixes #1538
Fixes #1650


manticore-projects:
@Wumpz: Please check out the great work on PR #1254. It adds a new token `QUICK` and which of course should not become a Reserved Keyword.

Of course the author @OlivierCavadenti  has no chance to know, that he would need to add `QUICK` to the productions `RelObjectNamexxxx()` in order to enable this token as Object Identifier. (And how should he, I figured that out only 6 month after my first contribution.)

This PR would solve this challenge reliably, preventing new tokens from breaking valid statements and easy the entry for new contributors.

build.gradle
```diff
@@ -227,6 +242,13 @@ task renderRR() {
         } 
     }
 }
+
+task updateKeywords(type: JavaExec) {
```
wumpz, line 246: Since I do not use gradle, what does this do?

Maven is the main build engine. You changed some gradle build options. Does the maven build still run?

wumpz:
The naming of this RelObjectName ... methods is sure an issue that should be adressed in futher improvements. So sorry for that: its grown ... ;)

wumpz:
I am not a fan of injecting those keywords, since not every keyword could be introduced this way and needs other parts of the grammar to be modified. However, building this keyword testing using JUnit 5 is cool.

build.gradle
manticore-projects: It generates and prints the Source Code Text of the method `RelObjectName()` which you would manually insert/replace in the JSQL Grammar file.

This step is optional and does not affect the Maven build. It is executed manually and on demand only.

Although the long term goal was to have a mechanism inside the build tool (Gradle and/or Maven), which during the build:

1) analyses the KeyWords automatically
2) modifies the Grammar file (methods `RelObjectName()`  and friends)
3) builds the Parser with the modified Grammar
4) runs the Keyword Tests

manticore-projects:
> The naming of this RelObjectName ... methods is sure an issue that should be adressed in futher improvements. So sorry for that: its grown ... ;)

You would do me a great favor by documenting the purpose of those 5 methods verbosely. I tried my best to figure it out by trial'n error but I am still not exactly sure why we ended up with 5 methods. (And no worries about grown code.)

manticore-projects:
> I am not a fan of injecting those keywords, since not every keyword could be introduced this way and needs other parts of the grammar to be modified. However, building this keyword testing using JUnit 5 is cool.

Honestly, I am not a big fan of that PR either -- but after spending many days on that matter trying various things it was the best approach I found and in my opinion it is still much better than the current state:

1) right now, Keywords are just a mess and we get a lot of issues about keywords only.
Are you able to document right now, what keywords are reserved and for what reason? I do not think so and the PR would solve this documentation problem.

2) right now, even when you have a opinion on keywords, we have no parametrized tests.
You admit, that these tests are kind of cool and I appreciate your feedback. Those parametrized tests however depend on a well defined list of keywords though which the PR provides. (Although we can argue of course, how/where exactly this list should be defined.)

>   not every keyword could be introduced this way and needs other parts of the grammar to be modified

3) That is true for the PR, but it is true by now already. Also we could easily enforce a policy of "Keywords are defines by simple Tokens" (not allowing complex Tokens for Keywords).

The work flow is to: a) define the Tokens and Productions in the Grammar first and then b) document any new RESERVED Keywords only in the List (emphasis on RESERVED).

The PR has one big advantaged: By defining RESERVED keywords only, it will determine automatically any allowed Keywords and modify the Grammar semi-automatically.

manticore-projects:
Allow me to ask please:

1) what was your preferred way to manage and document RESERVED KEYWORDs

2) what was your preferred way to maintain any allowed Token/Keyword in `RelObjectName()`

3) what was your preferred way to test all the possible Keywords (automatically)

And can we have a Skype chat about this :-)

wumpz:
The reserved keywords are introduced for syntactic rules. I do not get what you mean by documenting the keywords. The documentation or meaning of a keyword comes from its usage in syntactic rules. The introduction into RelOjectName or the other methods came by request.  
The split of this method came from the need to avoid LOOKAHEADS. Processing those changes the parsing order. Introducing it automatically or in large numbers changes the parse order in a not predictable way. So I avoided introducing new LOOKAHEADs wherever possible. 

That's why I am so reluctant to introduce those automatically.

manticore-projects:
> The reserved keywords are introduced for syntactic rules. I do not get what you mean by documenting the keywords.

How exactly does a user know what keywords are reserved and which ones he has to quote?
In H2, in gets the list from the website. In Oracle, he get the list from the website. And in JSQLParser?

How do we set up a consistent Use Case Test which automatically tests all keywords? And how do we ensure, that it is maintained properly after every change and adding new keywords?

So what I have done is do extract all tokens and to allow them, as long as those are not reserved.
It answers reliably both problems mentioned above.

> The documentation or meaning of a keyword comes from its usage in syntactic rules. The introduction into RelOjectName or the other methods came by request. The split of this method came from the need to avoid LOOKAHEADS. Processing those changes the parsing order. Introducing it automatically or in large numbers changes the parse order in a not predictable way. So I avoided introducing new LOOKAHEADs wherever possible.
> 
> That's why I am so reluctant to introduce those automatically.

We agree that LOOKAHEADs are bad and need to be avoided.
However, we are arguing if avoiding LOOKAHEADs justifies false parser errors (from keywords artificially not allowed).

The current approach of "Enabling on request" is half baken and "nicht Fisch or Fleisch", it just dodged the discussion.
Over the last 18 month most of the issues reported were about "keyword not supported, why?".

So, if you do not like the suggested solution, then please decide How exactly Keywords are to be maintained and documented.

wumpz:
> > The reserved keywords are introduced for syntactic rules. I do not get what you mean by documenting the keywords.
> 
> How exactly does a user know what keywords are reserved and which ones he has to quote? In H2, in gets the list from the website. In Oracle, he get the list from the website. And in JSQLParser?

JSqlParser does not provide it. That's right, but using JSqlParser does often not allow to change the underlying SQLs. Therefore the feature requests to allow this keyword as an identifier. But since we do not stick to one DBMS and could not build a "clean" grammar we have to go the other way around. Would'nt it more hurt, to give a list of reserved keywords because each could be discussed and even allowed. 

> 
> How do we set up a consistent Use Case Test which automatically tests all keywords? And how do we ensure, that it is maintained properly after every change and adding new keywords?

Since reserved keywords are transformed to allowed keywords by feature request, til now test cases were the result that is
still and will be part of JSqlParser. So you are right, that this could be forgotten. So maybe I did not get what you are trying to achive. I will read on ... :) 

> 
> So what I have done is do extract all tokens and to allow them, as long as those are not reserved. It answers reliably both problems mentioned above.
> 
> > The documentation or meaning of a keyword comes from its usage in syntactic rules. The introduction into RelOjectName or the other methods came by request. The split of this method came from the need to avoid LOOKAHEADS. Processing those changes the parsing order. Introducing it automatically or in large numbers changes the parse order in a not predictable way. So I avoided introducing new LOOKAHEADs wherever possible.
> > That's why I am so reluctant to introduce those automatically.
> 
> We agree that LOOKAHEADs are bad and need to be avoided. However, we are arguing if avoiding LOOKAHEADs justifies false parser errors (from keywords artificially not allowed).

What do you mean by false parse errors? Those errors occur if a reserved keyword is used. JSqlParser needs to disallow some
keywords to allow a multi database grammar. If you mean by false parse errors, that JSqlParser fails to parse a statement a specific database could process, then those exists. 

> 
> The current approach of "Enabling on request" is half baken and "nicht Fisch or Fleisch", it just dodged the discussion. Over the last 18 month most of the issues reported were about "keyword not supported, why?".
> 
> So, if you do not like the suggested solution, then please decide How exactly Keywords are to be maintained and documented.

Just wait with your response, I will recheck your implementation. Still, I think, I missed something.

wumpz:
So if I understand this right, that is the workflow:

1. build the parser using javacc 
2. compile it
3. run to get the derived RelObjectNames rule (and the others)
```java
System.out.println( buildGrammarForRelObjectNameWithoutValue() );
System.out.println( buildGrammarForRelObjectName() );
``` 
4. I assume you copy and paste those into the grammar file. 
5. build the parser using javacc
6. compile it
7. use it

However, my misunderstanding was from were you are deriving those keywords.  So the list of keywords is somehow fixed (at the moment in **CCJSqlParser.getDefinedKeywords**). I follow your idea to have a fixed list of those keywords including there lets call it **visibility type** and to build a complete test for those. 

What I dislike is this roundtrip. Why do we need to process JavaCC to be able to process JavaCC. So the list of keywords should be grammar externalized so you could do something like:

1. process a list of keywords (text file, ...)
2. change the specific parts in the grammar: **CCJSqlParser.getDefinedKeywords** and production rules. (this one should be done in maven / gradle alone without a compile)
3. build the parser
4. running tests using **CCJSqlParser.getDefinedKeywords**
5.  
6.

manticore-projects:
Thank you so much for this discussion, I do appreciate looking into this topic with you.

> So if I understand this right, that is the workflow:
>     1. build the parser using javacc
>     2. compile it
>     3. run to get the derived RelObjectNames rule (and the others)
> 
> ```java
> System.out.println( buildGrammarForRelObjectNameWithoutValue() );
> System.out.println( buildGrammarForRelObjectName() );
> ```
>     4. I assume you copy and paste those into the grammar file.
>     5. build the parser using javacc
>     6. compile it
>     7. use it

Yes, this is correct although we both agree, that it can't stay like that.
As soon as we reach an agreement, on what to do I would like to implement a `Gradle` task fully automating this. (`Gradle` has the great advantage, that you can write Java/Groovy code inside the build-script.)


> However, my misunderstanding was from were you are deriving those keywords. So the list of keywords is somehow fixed (at the moment in **CCJSqlParser.getDefinedKeywords**). I follow your idea to have a fixed list of those keywords including there lets call it **visibility type** and to build a complete test for those.

Yes, the Keywords have to be define explicitly, including the reason for being a keyword (either SQL Standard or JSQLParser Grammar). From this list we can derive:

1) The list of keywords for the documentation
2) The list of allowed tokens as difference between All Tokens and Keyword Token
3) The parametrized tests for allowed tokens


> What I dislike is this roundtrip. 

I do not like it either, but even after thinking about it I did not find any better solution.
Any idea was most welcome of course.


> Why do we need to process JavaCC to be able to process JavaCC. 
Because we need to get ALL TOKENS first, which are available only after `JTree` ran over the grammar. Only when we have ALL TOKENS, we can subtract the Keywords and define `RelObjectNames` and friends.

(Only alternative was to write/use a Parser for the Grammar, but then you replaced running JTree over the grammar with running the other parser over the grammar.)


So the list of keywords should be grammar externalized so you could do something like:
> 
>     1. process a list of keywords (text file, ...)
> 
>     2. change the specific parts in the grammar: **CCJSqlParser.getDefinedKeywords** and production rules. (this one should be done in maven / gradle alone without a compile)
> 
>     3. build the parser
> 
>     4. running tests using **CCJSqlParser.getDefinedKeywords**

This was not impossible, however I see some challenges:

1) How do you want to get ALL TOKENS?
2) Maintaining the Keywords outside the Grammar again means again a disconnect between Grammar and Configuration. Howe will this work, when another Contributor adds new Tokens?!
3) It is still a kind of a roundtrip, you just hid it by saying "should be done in maven/gradle". We fully agree that the roundtrip needs to be automated and I am fully committed to implement that -- but only after we agree on the general solution (because its a lot of work).

manticore-projects:
BTW, rebuilding `RelObjectNames` and friends is **only necessary, when new `Token` have been added**.
The situation I have in mind is: User submits a PR and introduces new `Token`, but is not aware that those must be allowed or else would become restricted keywords. (Like it happened to me when starting work on JSQLParser, I broke quite a few working statements during 4.2 and 4.3)

Nevertheless, I will wrap it into proper `Gradle` tasks of course.

manticore-projects:
The Gradle Task `updateKeywords` now updates and replaces ` RelObjectNameWithoutValue()` automatically without manual interaction.

It is still an optional call, so a full build would look alike:
```
gradle updateKeywords
gradle build
```

wumpz:
I took some days to think about this again. The goal is to automate all allowed keywords tests. Right? Maybe the documentation.

So this externalization of keywords in `getReservedKeywords` is a helper method to build parameterized tests for all this stuff. But this results in this weird workflow regenerating the grammer. 

Wouldn't it be better to parse the grammar file and simply extract those list the `getReservedKeywords` provides? After that all automated tests could be done like you propose and no regenerating the grammar is needed.

manticore-projects:
> I took some days to think about this again. The goal is to automate all allowed keywords tests. Right? Maybe the documentation.

Yes, and also: to automatically allow all Tokens, that are not reserved keywords (especially after adding new tokens).

> 
> So this externalization of keywords in `getReservedKeywords` is a helper method to build parameterized tests for all this stuff. But this results in this weird workflow regenerating the grammer.

No. `getReservedKeywords` is needed for knowing, which tokens to allow in `RelObjectName()` and friends. 
 
> Wouldn't it be better to parse the grammar file and simply extract those list the `getReservedKeywords` provides? After that all automated tests could be done like you propose and no regenerating the grammar is needed.

I think, you are still miss the main point: Tests and Documentation are nice side products.
The main goal though is to Allow the permitted Tokens (as difference between All Tokens and the Reserved Keywords), whenever new Tokens are added.


> By the way, renaming all RelObjectNames methods is a must.
Fully agreed.

Can we have a call on this topic maybe? I feel we could clarify the approach within a few minutes when we just talk about it.

manticore-projects:
> So this externalization of keywords in `getReservedKeywords` is a helper method to build parameterized tests for all this stuff. But this results in this weird workflow regenerating the grammer.

Actually, not much has changed: in 99% of all cases, you just build as before. Only when a new Token has been defined, you would run `gradle buildKeywords` first and even then it is optional (at the cost of Allowed Tokens and Documentation).

wumpz:
You are right, but if we read out those keywords directly from the grammar file we never need to run this `buildKeywords` because always all tests use the actual list. As always the documentation should be much more improved as well as the source code documentation (if I may make this difference).

manticore-projects:
Greetings.

I have modified the approach and read the Tokens directly from the Grammar file without invoking JTREE.
Thus all the Keywords have been moved our of the Grammar into ParserKeywordUtils and we spare the redundant run.

Cheers.

wumpz:
I did not look into your modifications, but I think I managed to parse the grammar using JavaCCs classes. (https://github.com/javacc/javacc/issues/221)

wumpz:
So if I understand your modifications right, you are scanning for all tokens defined in the token part. These different `RelObjectName` productions are used for different purposes. So depending on in which a token is used the tests should be different right?

manticore-projects:
> I did not look into your modifications, but I think I managed to parse the grammar using JavaCCs classes. ([javacc/javacc#221](https://github.com/javacc/javacc/issues/221))

Yes, I saw that. Thank you for your effort.
I am not sure, if this is better or worse than Parsing the Grammar per Regex because again, your engage the JavaCC Parser (now only from inside Java instead of from Gradle/Maven Plugin before).

At the same time i really do not care much as long as we get the Content of all tokens reliably (which seems to work both ways).

> So if I understand your modifications right, you are scanning for all tokens defined in the token part. 
Yes, but as per REGEX (because I do not want to rely on the `K_` or `S_` prefixes and also want to get the keywords of the compound tokens.

These different RelObjectName productions are used for different purposes. So depending on in which a token is used the tests should be different right?
> Yes, thats why there are two tests sofar.

wumpz:
So these two tests are enough? So we could merge some of those RelObjectName productions?

How we extract the keywords does not matter that much. What matters to me is the direct use of keywords without rebuilding a list, rebuilding the grammar building the parser, and so on. 

However, to read out all keywords correctly one has to rebuild this part of JavaCCs grammar parser, so why not reuse the original and not having troubles with some incomplete regular expressions?

manticore-projects:
> So these two tests are enough? So we could merge some of those RelObjectName productions?

I do see this PR as a first step to move us into a better direction. Certainly more fine tuning will need to follow. Although I feel like this PR is quite massive already and I would like to see a Merge first, before touching anything else.

This particular development step marries the best out of both worlds:
1) it demonstrates the idea and the benefits
2) it is still fully compatible with what has been there before and does not break anything

Please understand, that maintaining multiple PRs in parallel is quite a nightmare for any contributor and a lot of effort goes into keeping up with origin/main, depending on your Merge orrder.

> 
> How we extract the keywords does not matter that much. What matters to me is the direct use of keywords without rebuilding a list, rebuilding the grammar building the parser, and so on.
> 
> However, to read out all keywords correctly one has to rebuild this part of JavaCCs grammar parser, so why not reuse the original and not having troubles with some incomplete regular expressions?

Fully agreed, I would provide a similar method using JavaCC instead of the RegEx (so we could switch on/off later on demand).
JavaCC would be the default, but we could still use the RegEx whenever we would hit a snag.

manticore-projects:
I implemented they `getAllKeywords()` method based on JTree and got it working.
Although there are pros and cons:

1) We really get ALL keywords even from when not declared as a Token -- that's awesome!
2) The implementation is quite complex and verbose
3) We carry now JavaCC as a dependency, although we needed it only for the compilation -- I do not like that at all, but have no good idea what to do about that.

src/main/java/net/sf/jsqlparser/parser/ParserKeywordsUtils.java
```diff
@@ -0,0 +1,429 @@
...
+            | RESTRICTED_COLUMN
+            | RESTRICTED_EXPRESSION
+            | RESTRICTED_ALIAS
+            | RESTRICTED_SQL2016;
+
+
+    @SuppressWarnings({"PMD.ExcessiveMethodLength"})
+    public static List<String> getReservedKeywords(int restriction) {
+        // Classification follows http://www.h2database.com/html/advanced.html#keywords
+        Object[][] ALL_RESERVED_KEYWORDS = {
```
wumpz, line 64: Couldn't thoses lists directly derived from the grammar? At least RESTRICTED_ALIAS?

wumpz (commented):
I still do not get, why you want to hold this keyword list. Since we could extract all keywords, like you mentioned, and read all RelObj .. methods, we have all lists we need to get the reserved keyword list directly from the grammar without rewriting it. 

The dependency of JavaCC is only needed for tests. So this ParserKeywordsUtils list should be a test class only. 

Sure as I already mentioned, the RelObj - methods needs to be renamed to better reflect there meaning and use.

manticore-projects:
> I still do not get, why you want to hold this keyword list. 

We want to:

1) Create/Rewrite `RelObject....` automatically during the build process (at least whenever new tokens/productions have been added to the Grammer)

We will use the difference between ALL keywords and RESERVED keywords for this rewrite.
ALL keywords parsed from the Grammer, RESERVED keywords from the manually defined list.

> Since we could extract all keywords, like you mentioned, and read all RelObj .. methods, we have all lists we need to get the reserved keyword list directly from the grammar without rewriting it.

It is the other way around: we get the ALLOWED keywords and rewrite `RelObjectNames...` accordingly.
 
> Sure as I already mentioned, the RelObj - methods needs to be renamed to better reflect there meaning and use.

Certainly, I have asked you above for an explanation, what you had in mind when introducing the 4 methods. (The first 2 are clear to me, but I struggle to understand the methodology behind `...Ext` and `...Ext2`).

src/main/java/net/sf/jsqlparser/parser/ParserKeywordsUtils.java
manticore-projects: It is the other way around: This List will define, how we rewrite `RelObjectName...`. This List will be defined manually and everything else follows it.

src/test/java/net/sf/jsqlparser/statement/ConditionalKeywordsTest.java
```diff
@@ -37,7 +37,7 @@ public static Stream<String> keyWords() {
         List<String> keywords = new ArrayList<>();
         try {
             try {
-                keywords.addAll( ParserKeywordsUtils.getAllKeywords(file) );
+                keywords.addAll( ParserKeywordsUtils.getAllKeywordsUsingRegex(file) );
```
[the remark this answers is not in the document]
manticore-projects: It took me a bit of effort, to extract all keyword tokens based on your initial stub. I used the previous Regex generated List as reference. When both lists returned the same, I got confidence, that I got it working.

However, the JavaCC approach was based on trial an error and is not documented. So having a Test in the future might prove useful.

See it as the equivalent of you `toString()` vs `DeParser` -- it is redundant, but provides a great way to test the correctness.

build.gradle
```diff
@@ -30,22 +30,25 @@ repositories {
...
-    testImplementation 'org.mockito:mockito-core:4.+'
-    testImplementation 'org.assertj:assertj-core:3.+'
-    testImplementation 'org.hamcrest:hamcrest-core:2.+'
-    testImplementation 'org.apache.commons:commons-lang3:3.+'
-    testImplementation 'com.h2database:h2:2.+'
-    
+    testImplementation 'commons-io:commons-io:2.11.0'
+    testImplementation 'junit:junit:4.13.2'
+    testImplementation 'org.mockito:mockito-core:4.5.1'
+    testImplementation 'org.assertj:assertj-core:3.22.0'
```
d2a-raudenaerde, line 36: Is there a reason to explicitly list these here (compared to the 'older' on the left, where only the major version is fixed?

build.gradle
d2a-raudenaerde: I think this might also be implemented in a mojo in Maven, but that might be not worth the effort.

pom.xml
```diff
@@ -202,15 +211,6 @@
                             <goal>jjtree-javacc</goal>
                         </goals>
                     </execution>
-                    
-                    <!-- execute JJTree explicitely in order to generate the *.jj file needed for JJDoc -->
```
d2a-raudenaerde, line 206: Will removing this cause maven not to generate the source anymore when I just changes simple grammar rules?

src/main/java/net/sf/jsqlparser/parser/ParserKeywordsUtils.java
```diff
@@ -0,0 +1,429 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class ParserKeywordsUtils {
+    public final static CharsetEncoder CHARSET_ENCODER = StandardCharsets.US_ASCII.newEncoder();
+
+    public final static int RESTRICTED_FUNCTION = 1;
```
d2a-raudenaerde, line 43: While this works, Java's `enum` combined with `EnumSet.of()` combined is less 'Hacky'?

d2a-raudenaerde: for example `enum Restricted { FUNCTION, SCHEMA, TABLE, COLUMN ... }`

src/main/java/net/sf/jsqlparser/parser/ParserKeywordsUtils.java
```diff
@@ -0,0 +1,429 @@
...
+        if (file.exists() && file.canRead()) {
+            buildGrammarForRelObjectName(file);
+            buildGrammarForRelObjectNameWithoutValue(file);
+        } else {
+            throw new FileNotFoundException("Can't read file " + args[0]);
+        }
+    }
+
+    public static TreeSet<String> getAllKeywordsUsingRegex(File file) throws IOException {
+        Pattern tokenBlockPattern = Pattern.compile("TOKEN\\s*:\\s*(?:/\\*.*\\*/*)\\n\\{(?:[^\\}\\{]+|\\{(?:[^\\}\\{]+|\\{[^\\}\\{]*\\})*\\})*\\}", Pattern.MULTILINE);
```
d2a-raudenaerde, line 224: Cool regex. Maybe insert a comment to indicate/example what matches? I cannot read it from the regex directly.

src/main/java/net/sf/jsqlparser/parser/ParserKeywordsUtils.java
```diff
@@ -0,0 +1,429 @@
...
+
+    private static void addTokenImage(TreeSet<String> allKeywords, RStringLiteral literal) {
+        if (CHARSET_ENCODER.canEncode(literal.image) && literal.image.matches("[A-Za-z]+")) {
+            allKeywords.add(literal.image);
+        }
+    }
+
+    @SuppressWarnings({"PMD.EmptyIfStmt", "PMD.CyclomaticComplexity"})
+    private static void addTokenImage(TreeSet<String> allKeywords, Object o) throws Exception {
+        if (o instanceof RStringLiteral) {
```
d2a-raudenaerde, line 256: This would benefit from:  https://www.baeldung.com/java-pattern-matching-instanceof. But I think it is limited to JDK8 for now.

src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
```diff
@@ -1751,14 +1725,15 @@ String RelObjectNameWithoutStart() :
...
-		return result;
-	}
+    { return tk!=null ? tk.image : result; }
 }
 
 /*
 Extended version of object names.
+
+These tokens can be used as names for Schema and Tables and Columns
+BUT NOT for Aliases (without quoting)
```
d2a-raudenaerde, line 1735: :+1:  useful comment!

d2a-raudenaerde: Maybe a more descriptive name of the `RelObjNameExt` would be:
`SchemaTableColumnName`

src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
```diff
@@ -1768,26 +1743,24 @@ String RelObjectNameExt():
...
-        return result;
-    }
+        | tk=<K_CASEWHEN> | tk=<K_IN> | tk=<K_GROUPING> | tk=<K_ORDER>  )
+    { return tk!=null ? tk.image : result; }
 }
 
 /*
 Extended usage of object names - part 2. Using within multipart names as following parts.
+
+These tokens can be used as names for Tables and Columns
```
d2a-raudenaerde, line 1753: And there `TableColumnName`

src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
```diff
@@ -1733,7 +2007,7 @@ Table TableWithAlias():
     Alias alias = null;
 }
 {
-    table=Table() [alias=Alias() { table.setAlias(alias); }]
+    table=Table() [ LOOKAHEAD(2) alias=Alias() { table.setAlias(alias); }]
```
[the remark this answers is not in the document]
d2a-raudenaerde: To me it's not _directly_ obvious why this `LOOKAHEAD` is needed. Is that explained somewhere?

src/test/java/net/sf/jsqlparser/statement/KeywordsTest.java
```diff
@@ -0,0 +1,55 @@
...
+public class KeywordsTest {
+    public final static Logger LOGGER = Logger.getLogger(KeywordsTest.class.getName());
+
+    public static Stream<String> keyWords() {
+        File file = new File("src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt");
+        List<String> keywords = new ArrayList<>();
+        try {
+            keywords.addAll( ParserKeywordsUtils.getAllKeywordsUsingRegex(file) );
+            for (String reserved: ParserKeywordsUtils.getReservedKeywords(ParserKeywordsUtils.RESTRICTED_JSQLPARSER)) {
+                keywords.remove(reserved);
```
d2a-raudenaerde, line 40: `removeAll()` ?

src/test/java/net/sf/jsqlparser/statement/KeywordsTest.java
```diff
@@ -0,0 +1,55 @@
...
+            }
+        } catch (Exception ex) {
+            LOGGER.log(Level.SEVERE, "Failed to generate the Keyword List", ex);
+        }
+        return keywords.stream();
+    }
+
+    @ParameterizedTest(name = "Keyword {0}")
```
d2a-raudenaerde, line 48: cool!

wumpz:
To revive this.  

> > I still do not get, why you want to hold this keyword list.
> 
> We want to:
> 
> 1. Create/Rewrite `RelObject....` automatically during the build process (at least whenever new tokens/productions have been added to the Grammer)

Again I don't get why you want to create the RelObject productions during the build process. Why shouldn't the grammar be the only point of defining keywords and reserved keywords? 

> 
> We will use the difference between ALL keywords and RESERVED keywords for this rewrite. ALL keywords parsed from the Grammer, RESERVED keywords from the manually defined list.
> 
> > Since we could extract all keywords, like you mentioned, and read all RelObj .. methods, we have all lists we need to get the reserved keyword list directly from the grammar without rewriting it.
> 
> It is the other way around: we get the ALLOWED keywords and rewrite `RelObjectNames...` accordingly.
> 

Again look at the first point. Why should it be better to go this way? My way of thinking is that the grammar is carved in stone and all needed data should be derived from it. 

> > Sure as I already mentioned, the RelObj - methods needs to be renamed to better reflect there meaning and use.
> 
> Certainly, I have asked you above for an explanation, what you had in mind when introducing the 4 methods. (The first 2 are clear to me, but I struggle to understand the methodology behind `...Ext` and `...Ext2`).

+5


You know I am very reluctant going your way. To clarify this, my big problem is this modifying the grammar during the build process and not reading the grammar using JavaCC and deriving all needed data from it. The build process is modified from the running source code and therefore introduces another level of maintenance for maven and gradle builds. 

I have no problem in fact I am very in favour including automatic JUnit tests for all possible RelObject rule tokens like you did. The RelObject rule naming and number must be clarified and decreased.

manticore-projects:
> You know I am very reluctant going your way. To clarify this, my big problem is this modifying the grammar during the build process

We do not and will not modify the Grammar during every build process.
We provide 1 special task, to be invoked only on demand and only when the Token Definition has changed (e. g. after someone provided a PR introducing new tokens).

You can compare that with your Maven License Task, which would insert missing license headers.
Now we have a Keyword Tasks which whitelists Tokens when possible.

> and not reading the grammar using JavaCC and deriving all needed data from it.
We read the Grammar and we use JavaCC to extract the Tokens. Oversimplified: White Listed Keywords = ALL Keywords - Restricted Keywords.

Please tell me how to maintain that equation automatically and in JavaCC only.

> I have no problem in fact I am very in favour including automatic JUnit tests for all possible RelObject rule tokens like you did. The RelObject rule naming and number must be clarified and decreased.

The tests are only a marginal side product of the approach. Somehow I still fail to convey the major problem we are trying to solve: Automatic management of the Token Whitlisting inside `RelObject` and friends.

Again, I think the best way for a decision would be a call to directly explain and respond to questions. Please PM me, when possible. Thank you!

d2a-raudenaerde:
I gave it some extra tought and I think the automatic step to patch the grammar when keywords are whitelisted makes sense. However, I would be a bit reluctant to include tokens that are composed, like CURRENT DATE_TIME. I think they should be handled in a specific parsing rule for (in this case) dates. That would eliminate al the LOOKAHEAD(2) as well I think?

manticore-projects:
> However, I would be a bit reluctant to include tokens that are composed

You can easily classify it as Restricted Keyword.
The over simplified equation is: Allowed Keywords = All Keywords - Restricted Keywords.

All Keywords will be determined by the Grammar.
Restricted Keywords will have to be declared manually.

I do not see a particular reason to NOT allow `CURRENT_DATE` or `CURRENT_TIMESTAMP` -- but that's not the hill I want to die on.

d2a-raudenaerde:
I meant CURRENT TIMESTAMP (without the underscore) .. But agree on the hill  :)

wumpz:
> We read the Grammar and we use JavaCC to extract the Tokens.
> Oversimplified: White Listed Keywords = ALL Keywords - RESTRICTED Keywords
> Please tell me how to maintain that equation automatically and in JavaCC only.

Using our method parsing the grammar itself we get a list of all keywords and those that are used in RelObj and friends, which you called the white listed. Therefore the equation now turns to:

Oversimplified: RESTRICTED Keywords = ALL Keywords - White Listed Keywords

Comparing changing the grammar file with license headers does not work, since changing the grammar changes behavior.


The only difference to me is, that this list is available at runtime and not only test time. Is that what you want?

manticore-projects:
> Oversimplified: RESTRICTED Keywords = ALL Keywords - White Listed Keywords

It does not make much sense in this form because WE do manually define ALL keywords and we do define manually the RESTRICTED Keywords, while the WHITE LISTED keywords shall be determined (by the code) -- in order to be complete and accurate.

Right now WHITE LISTED keywords is defined manually only and so is incomplete, We received error reports about and fix one by one instead having a holistic solution.

> Comparing changing the grammar file with license headers does not work, since changing the grammar changes behavior.

It changes code, which is supposed to be changed in anyway (adding the same keyword manually on demand). This change is guarded by comprehensive tests.

> The only difference to me is, that this list is available at runtime and not only test time. Is that what you want?

The difference is, that WHITE LISTED keywords are kept complete and consistent even when new tokens and productions will be added.

manticore-projects:
Btw, the `Keywords` task is completely optional -- it is up to you, if and when to execute it. It only spares you from editing the `RelObjectNames...` productions by hand.
Although I would advise to run it whenever new `Token` and/or `Production` as been introduced.

manticore-projects:
Recent addition of the `OVERLAPS` production is a perfect example:
1) Contributor commits `OVERLAPS` production without amending `RelObjectName....`
2) Keyword Tests fail
3) Gradle `generateKeywords` tasks is started and fixes the Grammar by amending  `RelObjectName....` and also updates the documentation of Reserved Keywords
4) User can continue to parse his personal favorite statement 
    ```sql
    select overlaps from overlaps.overlaps where overlaps='overlaps'
    ```

Result: everyone is happy and I do not need to answer 3 different issues about the new `OVERLAPS` token.

@d2a-raudenaerde : apologies for using your contribution in such a polemic way.

d2a-raudenaerde:
@manticore-projects thanks for noticing I missed the `RelObjectName` part. I thing it would be nice if there was a test that matches the keywords with the previous version so there is an error when it's not correctly added as RelObjectName. 
It would be good to have explicit lists imho. 
(I would never write sql like `select overlaps from overlaps.overlaps where overlaps='overlaps'`, but if it's valid, it should parse correctly)

manticore-projects:
> I think it would be nice if there was a test that matches the keywords with the previous version so there is an error when it's not correctly added as RelObjectName.

Yes, this is exactly what the PR does. It does not stop there, but also corrects the Grammar and adds your new Token -- unless you have explicitly marked it as RESTRICTED Token.

manticore-projects:
> (I would never write sql like `select overlaps from overlaps.overlaps where overlaps='overlaps'`, but if it's valid, it should parse correctly)

Well, please humor yourself and read some of the Error Reports we are receiving. Although I belong do the "everything goes" faction myself.

d2a-raudenaerde:
I think the PR could benefit from an extra bit of `Development` documentation, either in the readme or in a separate file on what to do if (for example, adding a keyword).  (basically, a small rewrite from some of the explanation (why&how) you already wrote in this PR)
So the workflow is clear for people that are not yet deeply into the code / build paths.

manticore-projects:
> I think the PR could benefit from an extra bit of `Development` documentation, either in the readme or in a separate file on what to do if (for example, adding a keyword). (basically, a small rewrite from some of the explanation (why&how) you already wrote in this PR) So the workflow is clear for people that are not yet deeply into the code / build paths.

Thank you for your feedback -- I do agree and this documentation belongs to the development section of the web-site, so contributors can read it and amend. Please refer to PR #1624.

Problem is, that we have contradicting objectives here:
1) PRs shall be small and self containing
2) PRs shall be complete and working fine before getting accepted
3) Simple PRs always get priority and get pulled first, which keeps breaking my own PRs.

@Wumpz: I am pretty much at the point, where I will just set up a "JSQLParser Development" branch in my own Repository and you can pull from it whatever you want , when you want. We are moving way too slow.

manticore-projects:
> I think the PR could benefit from an extra bit of `Development` documentation,

Here you go: http://manticore-projects.com/JSQLParser/contribution.html#manage-reserved-keywords

manticore-projects:
> What works really well for me, is an example

Great idea, there you go: http://manticore-projects.com/JSQLParser/contribution.html#id1

manticore-projects:
Good Morning.

Thank you for your feedback and providing a modified approach.

>     1. Automatically Derive all Keyword Tokens from the grammar file.
>     2. Automatically derive all whitelisted Tokens from the grammar file
>     3. Derives from 1 and 2 the list of restricted keywords and checks it against _RestrictedKeywordsTest.RESTRICTED_KEYWORDS_
>     4. Parametrized Tests for Keywords, testing all keywords
>     5. runs as a normal test every time somebody builds JSqlParser and _fails everytime_ somebody added a token without specifying it as restricted or used it in RelObj productions

I believe, until this point we have a similar understanding and do more or less the same. No objections.
Question now is: what exactly happens, when this test from 5) fails and detects a new Token Keyword.

Answer: we add it to `RelObjectNameWithoutValue()`. You want to do that strictly manually, I proposed doing that as per script.
It would result in exactly the same result. The irony is, that the Job is available in Groovy. So, when you want to do that manually -- be my guest. Run Maven and insert manually. I would just fire the Gradle task instead.

The result would be the same!

>     6. does not change the build process and does not need to run sometimes a grammar patching tool

As I said: you would still "patch the Grammar" -- just manually. Exactly the same change, about exactly the same keyword(s).
 
>     7. it uses JavaCC to parse the grammar and extracts all needed information, therefore we do not rely on special regexp cases, and IMHO the extraction in _JavaCCExtractor_ is way more simple

This is correct of course. At the same time it relies on people and manual intervention. I for my part trust the Regex more.

Long story short: from your explanation, I can see a difference only in the Gradle `updateKeywords` Task -- which is optional. If you do not want to run it, don't run it.

manticore-projects:
> * It makes the build process more complicated. (I know, not every time.)

The Gradle task is optional and you run it only when another user as added a new Token.

> * At the moment it only allows the patching using gradle (Right?).

Yes, because I refuse to deal with Maven. Also, because it is optional.
Last but not least, it could be added to Maven -- but where is the point, when you don't want to execute it in anyway?
 
> * If you forget to run those patching stuff you are running an inconsistent project.

No. Never anything would be (more) inconsistent.
If you don't run the Gradle `updateKeywords` task, then you would have some artificially restricted Tokens, which you can't use as Object names without any reason. Exactly the situation, we have right now already. Not worse.

> * If you merge PRs you have always to take care to merge changes made in RelObjName productions back to `ParserKeywordsUtils`.  

That is actually a valid concern, when the Gradle task is optional.
You would have to: a) merge first then b) run the Gradle task and then c) push another change.

Of course this could be solved, when making this `updateKeywords` Task a mandatory build step -- similar to `PMD` and `Checkstyle`, being mandatory yet.

> BTW does this grammar patching overwrite some additions to the RelObjName methods?

Indeed it overwrites/replaces the production, using its own template for it.
So when you would like to implement "additions", you would need to add to those templates too. (Although I can't see yet, what those additions would be.)

> * One patching is not enough since introducing a new token takes a guess of which type of restriction / whitelisting is working. So changing the grammer and rebuilding is not enough because you every time need to repatch the grammar with a new restriction type.

I disagree: First of all, `Reserved` Tokens are defined by the SQL Standard and adding additional Restrictions should be avoided.
Secondly, this is what happens NOW already: An experienced contributor will add new Token and whitelist it immediately -- then tests fail and he will remove it from the Whitelist again.

The PR would: whitelist the Token automatically and run the Keyword Test, which would fail. So he would add it to the Reserved Keyword List, rerun the `updateKeywords` Task and only commit after.

wumpz:
> I believe, until this point we have a similar understanding and do more or less the same. No objections.
Yes

> Question now is: what exactly happens, when this test from 5) fails and detects a new Token Keyword.
> 
> Answer: we add it to RelObjectNameWithoutValue(). You want to do that strictly manually, I proposed doing that as per script.
> It would result in exactly the same result. The irony is, that the Job is available in Groovy. So, when you want to do that manually -- be my guest. Run Maven and insert manually. I would just fire the Gradle task instead.

Not only we, the developer who extends JSqlParser is forced to do it. 
And this is only possible if the restricted keywords (keywords not in RelObjName) are listed somehow. 

> As I said: you would still "patch the Grammar" -- just manually. Exactly the same change, about exactly the same keyword(s).

Strange wordplay ... the meaning of patching is different. 
However, 

change grammar file and build
vs
change Restricted Keyword List, start optional Gradle task to patch grammar, build 

However, we are talking about the RelObjName productions. The other stuff has to be done as well in the grammar file. So the "change grammar file and build" has to be done anyway. This is a more complicated build process with optional parts. 

And you are right, the result would be the same. I would like to keep formatting or comments in RelObjName productions. Could you provide that? 

> I argue about that: The PR detects all Tokens and would whitelist and test any token, that has not explicitly restricted yet. Both ways are "complete", only difference is that you want to whitelist the tokens manually while I would like to use a script.

Since you only have a list of all whitelisted keywords in ParserKeywordsUtil are not detected. So every addition of tokens needs to be recognized by a developer. Or did I read your code wrong? That's the reason I prefer this 

RESTRICTED Keywords = ALL Keywords - White Listed Keywords

and I keep a list of the restricted keywords to check against. The whitelisted keywords do not change while adding new tokens. 

> Yes, because I refuse to deal with Maven. Also, because it is optional. Last but not least, it could be added to Maven -- but where is the point, when you don't want to execute it in anyway?

I pointed only out that the implementation is not complete. Even if it is optional it is part of the build process. 

> Indeed it overwrites/replaces the production, using its own template for it.
> So when you would like to implement "additions", you would need to add to those templates too. (Although I can't see yet, what those additions would be.)

Comments, ordering by meaning, grouping of keywords.

wumpz:
So I will merge, if we could let both methods coexist. So somehow this overwriting should be avoided.

manticore-projects:
> So I will merge, if we could let both methods coexist. So somehow this overwriting should be avoided.

I think this is a fantastic idea. Both ways will need a method to find ALL the tokens in the Grammar and so we easily agree on that and both ways must end with the same, tested white-listing.

Lets go for it!

manticore-projects:
> However, 
> change grammar file and build vs change Restricted Keyword List, start optional Gradle task to patch grammar, build

Not exactly. "Add Restricted Keyword" should be very very rare, because we want to avoid further restriction. It should be a rather static list.

Furthermore: We can build directly from the `updateKeywords` task (having a build `including` and a build `excluding` this step).

manticore-projects:
> Since you only have a list of all whitelisted keywords in ParserKeywordsUtil are not detected. So every addition of tokens needs to be recognized by a developer.

No, it is exactly the opposite: `Restricted Keywords` are rather static and would not be touched normally. Developer adds to `All Keywords` only by adding tokens/productions to the Grammar. Then the Gradle task builds `Whitelist Keywords` automatically and confirms them by brute force testing.

**Only when** those tests fail, we would have to amend `Restricted Keywords`. This could even be automated, it is simple Java Code after all (or could be a Text file or whatever.)

Where I still fail to understand you plan is this point:

1) We agree to run the brute force tests in order to identify/confirm `Restricted Keywords`, do we?
2) But those tests will work only AFTER updating the `Whitelisted Keyword` in `RelObjectNameWithoutValue()` -- because the tests rely on the final grammar.

So how would you run your Tests without whitelisting in the grammar first?

wumpz:
> So how would you run your Tests without whitelisting in the grammar first?

Simply this test will fail until you whitelist or restrict the new tokens. That's what forces developers to do this decision (for us). You are right, that the restricted list should change rarely. The test should then give an advice, how and were this change should be done. 
I am a fan of brute force testing changes a developer (can) make. In a sense this is the same like enforcing a specific code style.

> No, it is exactly the opposite: Restricted Keywords are rather static and would not be touched normally. Developer adds to All Keywords only by adding tokens/productions to the Grammar. Then the Gradle task builds Whitelist Keywords automatically and confirms them by brute force testing.

I was not aware of this automatic whitelist keyword building. So this is +1.

wumpz:
So are you going to resolve the conflicts? This improvement of parsing could be done afterwards.

vipcxj:
@wumpz When could this pr be merged? The parse error is really annoying.

manticore-projects:
> @wumpz When could this pr be merged? The parse error is really annoying.

I can supply you with a bleeding edge JAR file including those PRs, in case its urgent.

wumpz:
So after merging, I am not able to build anymore since `ConditionalKeywordsTest` fails. I merged another PR. 

So how should I proceed if I am not willing to use gradle? Isn't that the same problem you had with my proposal? Maven is the main build pipeline.

manticore-projects:
Good Morning. Thank you for finally accepting the PR.

What caused the problem:
1) you have accepted another PR before (introducing the new Token `LOCKED`)
2) you have not added `LOCKED` to the white- or black-list

Solution:
Amend the white- or black-list and then run the `buildKeywords` Task.

This should have been done by the author of the `LOCKED` PR from now on any author will be forced to do so. 

It is exactly as you have requested for in your explanations.

Although I will run through this procedure today and send a very small PR.

manticore-projects:
> So how should I proceed if I am not willing to use gradle?

If you insist in not using `gradle updateKeywords` then you still can amend `RelObjectNameWithoutValue` manually and add the whitelisted new token -- same as it has always been done before.

I would also provide an equivalent Maven task soon, since we are moving forward now.

Workflow should be like that:

    1) Amend Tokens in the Grammar
    2) When keyword-tests fail, run gradle updateKeywords
    3) Then build via mvn package or gradle build (it does not matter)
    4) Push/Deploy

Using gradle updateKeywords task is optional, you can also add the new Tokens manually has it has always been done before.

After accepting the Keywords PR, any further PR will automatically fail acceptance tests, unless keywords are in order.

wumpz:
No developer that is using maven knows what to do here? So this workflow has somehow to be included in a developer help. At the moment this is only your tool, since nobody knows and not everybody uses gradle.

wumpz:
Since we are not willing to run the gradle task and your tests use `ParserKeywordsUtils.getReservedKeywords` how should updating the grammar be enough?

wumpz:
Additional since some JavaCC methods are in place now there is a direct dependency for JavaCC.

manticore-projects:
> Additional since some JavaCC methods are in place now there is a direct dependency for JavaCC.

Yes, because you insisted to parse the Grammer using JavaCC instead of using Regular Expressions. I suggest to use Regular Expressions instead for that reason.

> Since we are not willing to run the gradle task and your tests use `ParserKeywordsUtils.getReservedKeywords` how should updating the grammar be enough?

Because this is exactly what you would have done before the semi-automation: edit the Production `RelObjectNameWithoutValue` manually. You still could very much do that and we should keep that option at all cost.

That said, I have added a Maven Task for running the same `updateKeywords` task.
You can run it via:
```bash
mvn compile exec:java
```

> No developer that is using maven knows what to do here? So this workflow has somehow to be included in a developer help. At the moment this is only your tool, since nobody knows and not everybody uses Gradle.

Yes, of course. It has been verbosely documented within the new website, which I have committed 1 month ago or so: https://manticore-projects.com/JSQLParser/contribution.html#manage-reserved-keywords

I have updated the Keywords2 PR #1653 accordingly with your requested Maven task.

d2a-raudenaerde:
The Maven task _could_ be implemented as a mojo (maven-plugin) I think, are you interested in that? Then you can run this plugin from the pom.xml during, for example, the generate-sources phase.

d2a-raudenaerde:
I have some experience writing these maven plugins, so if I infd some time this week, maybe I could draft up some integration so it 'just works' when you run for example `mvn compile`

d2a-raudenaerde:
But probable this wil work as well:
```
<plugin>
        <groupId>org.codehaus.mojo</groupId>
        <artifactId>exec-maven-plugin</artifactId>
        <executions>
            <execution>
                <id>codegeneration</id>
                <phase>generate-resources</phase>
                <goals><goal>java</goal></goals>
                <configuration>
                   <mainClass>com.codegenerator.CodeGeneratorApplication</mainClass>
                </configuration>
            </execution>
        </executions>

</plugin>
```

manticore-projects:
Thanks! I have add submitted another PR like that on Saturday already.

d2a-raudenaerde:
Ah excellent!
 I'll have to catch up I see :)

d2a-raudenaerde:
I can't find that PR? I think there might also be a chicken-and-egg problem, as the java code that does the work, also needs to be compiled first? (you can solve that using multi-module maven build, but that requires a bit more refactoring)

manticore-projects:
this one: https://github.com/JSQLParser/JSqlParser/pull/1653

Its pretty straight forward actually since we can execute Java Code directly without compiling the whole package. Also, this step is executed on demand: only when new tokens introduced and/or the Keyword Tests fail.

wumpz:
> Yes, because you insisted to parse the Grammer using JavaCC instead of using Regular Expressions. I suggest to use Regular Expressions instead for that reason.

That's why e.g. my proposal is only running in scope *test*. Then this dependency is not needed. I was talking about the IMHO right way to parse the grammar and not the context in which this parsing should happen. I thought this should be obvious. For the same reason, the JSqlParser jar should not include this generation class. It is part of the build process (I know you do not accept that, but I think it's evident.) So a quick solution would be to put all of it in the test folder and make this JavaCC dependency of scope test.  

However, to get a clean build again use for now Regular Expressions. 

Saying that, since this is a build step or could be one, it should be separated from JSqlParser source code and put in some kind of build helper module and then could be definitely used if JSqlparser is built. Now If you clone this project you will fail, since the generator class is not yet built. 

> I have updated the Keywords2 PR https://github.com/JSQLParser/JSqlParser/pull/1653 accordingly with your requested Maven task.

I have reviewed it already and found in this issue the needed information on how to run it, so I will merge it.

PR #1384, opened 2021-10-19
IN() with complex expressions

### LINKED ISSUE #905: Multiple inner query inside IN clause does not work
**Describe the bug**
SQL with multiple inner query inside **IN clause** does not work. Example below

**To Reproduce**
1. `select * from table_a where other_id in ((select id from table_b where name like '%aa%'), (select id from table_b where name like '%bb%'))`
2. `CCJSqlParserUtil.parse(sql)`
3. `Encountered unexpected token: "in" "IN"`

**Expected behavior**
The query runs properly when executing in SQL Server.

**System**
 - Database: MSSQL 2012
- Java Version: 1.7
- JSqlParser version: 2.1

manticore-projects:
Allows `IN()` to uses Complex Expression Lists including multiple `SubSelects`
Simplifies the `IN()` Statement by removing the Multi Expression List property
Fixes #905

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -71,12 +71,63 @@ jacocoTestReport {
 jacocoTestCoverageVerification {
     violationRules {
         rule {
+            //element = 'CLASS'
             limit {
-                minimum = JavaVersion.current().isJava8() // for any reason, different results
-                        ? 0.83                            // depending on the Java Version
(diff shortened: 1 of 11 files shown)


wumpz:
Please resolve conflicts.

PR #1389, opened 2021-10-25
Optimize assertCanBeParsedAndDeparsed

manticore-projects:
- Avoid redundant calls of buildSqlString()
- Replace String.replaceAll() with Matcher.replaceAll() based on precompiled Regex Patterns

The actual benefit is only small, but we call these methods very often during the tests.


wumpz:
Thx. I see the benefit more in readability. :)

PR #1395, opened 2021-10-29
Support Postgres Dollar Quotes #1372

### LINKED ISSUE #1372: $$text with space$$ inside function doesnt work
**To Reproduce**
1. `SELECT UPPER($$some text$$)`
3. 
```
org.junit.ComparisonFailure: Expected query to parse but it failed to parse (error: Encountered unexpected token: "(" "("
    at line 1, column 13.
```

**Expected behavior**
query to parse

**System**
 - postgres
- Java11
- JSqlParser 4.1

OlivierCavadenti:
- Support Postgres Dollar Quotes notation.
- Fixes #1372.

Examples of queries:

```sql
SELECT UPPER($$some text$$) FROM a
SELECT * FROM a WHERE a.test = $$where text$$
SELECT aa AS $$My Column Name$$ FROM a
```


manticore-projects:
Greetings! Thank you very much for your contribution, we do look forward receiving more soonest.
In my opinion, the PR looks very good and has incorporated sufficient Unit Test Cases. Well done.

Issue #1401, opened 2021-11-04
Fix dependencies in maven pom.xml


gitmotte:
@manticore-projects Please keep pom.xml and build.gradle in sync. 

```
<dependency>
        	<groupId>org.junit.jupiter</groupId>
        	<artifactId>junit-jupiter-api</artifactId>
        	<version>5.7.1</version>
        	<scope>test</scope>
</dependency>
```
mvn build has compile errors.

manticore-projects:
Sorry, my bad. I always forget about Maven. Thank you for stepping in and fixing it.
Cheers

PR #1403, opened 2021-11-04
#1401 add junit-jupiter-api

gitmotte:
#1401

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,6 +38,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+        	<groupId>org.junit.jupiter</groupId>
+        	<artifactId>junit-jupiter-api</artifactId>
+        	<version>5.7.1</version>
+        	<scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
             <version>2.28.2</version>


wumpz:
So you are simply replacing this jar which is using the vintage engine to run the junit 4 tests we have now? Or are you going to replace all annotations and used classes to the jupiter version?

gitmotte:
Not at all, this is just a bugfix for a missing dependency in Maven's pom.xml  
Please have a look at this comment https://github.com/JSQLParser/JSqlParser/issues/1401#issuecomment-961732746

wumpz:
Sorry. I see that within some of the former PRs there was this change so JSqlParser now indeed uses some kind of mixture.

wumpz:
However, I will change the import to be proper for this case.

manticore-projects:
The JUnit 5 dependency was brought in by me for allowing parallel tests with Gradle. Forgetting to add it to Maven was an oversight. (I never liked Maven and jumped from ANT to GRADLE directly).

That said, I think it would make sense to migrate to JUnit 5 and to drop JUnit 4 eventually.

wumpz:
No problem here. I had the same problematic in diff rent other projects. You have to check that your tests are running. If wrongly configured,. Junjt 4 or 5 tests are skipped.

However, with this knowledge I consider this my problem, since I did not see the chance before.

wumpz:
Look into #1427.

PR #1409, opened 2021-11-07
added USE SCHEMA <schema> and CREATE OR REPLACE <table> support

Bartman0:
To parse out Snowflake DDL that we use I needed to extend the allowed syntax for USE SCHEMA <schema> and CREATE OR REPLACE <table>


manticore-projects:
Thank your for your contribution. Very nice and clean PR, well done. We do look forward receiving more soonest.

PR #1410, opened 2021-11-09
table name finder return full table names

sheiy:


manticore-projects:
Greetings.

Thank you for the contribution. The final decision would be on @wumpz, but in my opinion this implementation is redundant: 

1) If we decide to pull the actual `Table` objects (rather then the names only), then we do not need to pull out the names in parallel as we can get the names from the list of `Tables`

2) some Test Cases and Documentation of the (new) methods would be good. It took me some minutes to actually understand what you have tried to achieve

3) I see the TableNames Finder as a kind of Documentation/Example, not really as part of an API. Thus I do not see much benefit in enhancing it. Instead the users would take it as a template and amend it according to their specific needs. I am not sure if TableNames Finder is a good example for 'one fits all'.

src/main/java/net/sf/jsqlparser/util/TablesNamesFinder.java
```diff
@@ -77,6 +77,7 @@
 
     private static final String NOT_SUPPORTED_YET = "Not supported yet.";
     private List<String> tables;
```
manticore-projects, line 79: maybe rename into a speaking name like `tableNames`

sheiy:
OK thanks for your suggestion 
I will close the PR

wumpz:
@sheiy  So you simply want to extract the full table names? This is easy. `TableNamesFinder` provides this handy method called: `extractTableName´ which is called with every found `Table` object to extract the tablename. Within this you can do and extract from the `Table` object what you want.

PR #1411, opened 2021-11-10
Add support for quoted cast expressions

tomershay:
PostgreSQL allows double quoting type names in "::" casting operations.

Note how "enum_test" is double quoted in this sample query, and it's valid to do it with built in types like "text" as well:
`SELECT col FROM test WHERE status = CASE WHEN anothercol = 5 THEN 'pending'::"enum_test" END`

Adding support for double quotes around the cast type.


wumpz:
@thomershay thx for your addition.

PR #1412, opened 2021-11-10
Support complex expressions in select clause

tomershay:
The SELECT clause can contain complex expressions, for example "(condition) OR (condition)", which actually returns a boolean.

Sample query:
`SELECT (CASE WHEN col IS NULL THEN False ELSE True END) OR (CASE WHEN col2 IS NULL THEN False ELSE True END) FROM tbl`

Adding support for such complex conditions in the SELECT clause.

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -2022,7 +2022,7 @@ SelectExpressionItem SelectExpressionItem():
     Alias alias = null;
 }
 {
-      expression=Condition() { selectExpressionItem = new SelectExpressionItem(); selectExpressionItem.setExpression(expression); }
+      expression=Expression() { selectExpressionItem = new SelectExpressionItem(); selectExpressionItem.setExpression(expression); }
              [alias=Alias() { selectExpressionItem.setAlias(alias); }] { return selectExpressionItem; }
 }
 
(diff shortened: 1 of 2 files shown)


manticore-projects:
Greetings. I have provided a similar PR already which is pending since a few days.Sent from my Galaxy
-------- Original message --------From: "Tomer Shay (Shimshi)" ***@***.***> Date: 10/11/2021  20:16  (GMT+01:00) To: JSQLParser/JSqlParser ***@***.***> Cc: Subscribed ***@***.***> Subject: [JSQLParser/JSqlParser] Support complex expressions in select clause
  (PR #1412) The SELECT clause can contain complex expressions, for example "(condition) OR (condition)", which actually returns a boolean.
Sample query:
SELECT (CASE WHEN col IS NULL THEN False ELSE True END) OR (CASE WHEN col2 IS NULL THEN False ELSE True END) FROM tbl
Adding support for such complex conditions in the SELECT clause.



You can view, comment on, or merge this pull request online at:
  https://github.com/JSQLParser/JSqlParser/pull/1412

Commit Summary

  Support complex expressions in select clause


File Changes  (2 files)

  
    M
    src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
    (2)
  
  
    M
    src/test/java/net/sf/jsqlparser/statement/select/SelectTest.java
    (6)
  


Patch Links:

  https://github.com/JSQLParser/JSqlParser/pull/1412.patch
  https://github.com/JSQLParser/JSqlParser/pull/1412.diff

manticore-projects:
@tomershay: I would like to close this PR since a similar PR has been merged quite a time ago and this challenge should be resolved.

PR #1428, opened 2021-11-22
Adjust Gradle to JUnit 5

### LINKED ISSUE #1398: TokenMgrError is a subclass of Error instead of Exception
**Describe the bug**
An error is thrown instead of an exception when a token parsing error occurs.

An error should only really be thrown for very serious errors like out of memory, thread death, etc. Even an NPE is only classes as an exception. If you should be catching it, it shouldn't be an Error, it should be an exception.

From the Java docs:
> An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch. Most such errors are abnormal conditions. The ThreadDeath error, though a "normal" condition, is also a subclass of Error because most applications should not try to catch it.

**To Reproduce**
1. Example SQL: Unknown as our server shutsdown instead of catching it (as should happen with Errors).
3. Exception:
```
net.sf.jsqlparser.parser.TokenMgrError: Lexical error at line 4, column 39.  Encountered: "\u00a8" (168), after : ""
	at net.sf.jsqlparser.parser.CCJSqlParserTokenManager.getNextToken(CCJSqlParserTokenManager.java:5943)
```

**Expected behavior**
An exception to be thrown instead of an error.

**System**
 - Database you are using Postgres
- Java Version 15
- JSqlParser version 4.3-SNAPSHOT

manticore-projects:
After JUnit 5 migration, tests were not found when building with Gradle. This PR fixes the issue and adds smaller Gradle goodies. 

Clean Gradle build takes on 4 seconds (when cached).
Clean Maven PACKAGE takes 50 seconds.

Parallel Test execution
Gradle Caching
Explicitly request for latest JavaCC 7.0.10
Fixes #1398

### DIFF
--- a/src/test/java/net/sf/jsqlparser/statement/select/SpeedTest.java
+++ b/src/test/java/net/sf/jsqlparser/statement/select/SpeedTest.java
@@ -26,7 +26,7 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.parallel.Execution;
 import org.junit.jupiter.api.parallel.ExecutionMode;
 
-@Execution(ExecutionMode.CONCURRENT)
+//@Execution(ExecutionMode.CONCURRENT)
 public class SpeedTest {
 
     private final static int NUM_REPS_500 = 500;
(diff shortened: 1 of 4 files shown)


src/test/java/net/sf/jsqlparser/statement/select/SpeedTest.java
```diff
@@ -23,7 +23,10 @@
 import net.sf.jsqlparser.test.TestException;
 import net.sf.jsqlparser.util.TablesNamesFinder;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.parallel.Execution;
+import org.junit.jupiter.api.parallel.ExecutionMode;
 
+@Execution(ExecutionMode.CONCURRENT)
```
wumpz, line 29: This parallel execution would make the numbers from this speed test meaningless.

manticore-projects: Am not entirely clear, how and why. But I have removed that annotation from the Speed Test.

Issue #1458, opened 2022-01-12
Error when parsing switch expressions with OR or AND.


mgoeminne:
**Describe the bug**

First things first: thank you for your work!

A switch expression of a CASE, if containing a OR or a AND, makes the parsing crash.

**To Reproduce**

```sql
SELECT true AND false; 
``` 

and 

```sql
SELECT true OR false;
```

work fine

```sql
SELECT CASE true | false
     WHEN true THEN 1
     ELSE 2
END
```

and

```sql
SELECT CASE true & false
     WHEN true THEN 1
     ELSE 2
END
```

also have the expected behaviour (with a bitwise expression being returned as the switch expression).

But

```sql
SELECT CASE true AND false
         WHEN true THEN 1
         ELSE 2
END;
```

raises an exception:

```
Encountered unexpected token: "AND" "AND"
```

(the issue is similar when AND is replaced by OR).

**Expected behavior**
The last statement should be successfully parsed, with a AndExpression/OrExpression used as the switch expression.
At least, https://www.w3schools.com accepts it, and Datagrip sends this kind of statement on Postgres databases.

**System**
 - Database you are using: my own, mimicking PostreSQL.
- Java Version: java 12.0.2
- JSqlParser version: 4.3

wumpz:
Confirmed. JSqlParser only allows there a condition. Within the where more complex expressions are allowed:

```
case when a and b then .. end
```

mgoeminne:
In the when, an Expression is required, while case switches only allow a Condition.

Should I submit a pull request that replaces the Condition with an Expression? By doing so, all the test I proposed in https://github.com/JSQLParser/JSqlParser/pull/1459 do pass.

wumpz:
Sorry  for the delay. Sure you could provide a PR. But be aware that the performance could degrade.

manticore-projects:
Closed, since no further question has been asked.

PR #1459, opened 2022-01-15
Extending CaseExpression, covering #1458

mgoeminne:


mgoeminne:
Related issue: #1458

wumpz:
Tests will only go into JSqlParser if they all succeed.

mgoeminne:
The last commit is an attempt to fix #1458 , and now all tested proposed in this pull request actually pass.

src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
```diff
@@ -4368,7 +4368,7 @@ Expression CaseWhenExpression() #CaseWhenExpression:
 }
 {
     <K_CASE> { caseCounter++; }
-    [ switchExp=Condition() ]
+    [ switchExp=Expression() ]
```
wumpz, line 4371: How much does the performance degrade? Where are the tests?

mgoeminne: There is no performance test for the moment. Do you have a standardized way to perform that king of tests? Do you have existing tests showing the current performances? Also, how to evaluate the tradeoff between a supposed drop in performance and a better support of the SQL language?

PR #1498, opened 2022-03-30
Update JsonAggregateFunction.java

doukai:
Add limit for JSON_ARRAYAGG function in MySQL


manticore-projects:
Thank your for working on JSQLParser. I have briefly looked at your changes:

1) You do not implement any support for the LIMIT term in the grammar. Without the grammar, your change won't do anything.

2) You do not implement your change in the Deparsers and the `toString()` method. Without, your statement would not be de-parsed correctly and also any tests would fail.

3) You did not implement any Unit Tests yet (which would have spotted 1) and 2) above )

PR #1506, opened 2022-04-07
Json function Improvements

### LINKED ISSUE #1504: cannot parse mysql column in JSON_OBJECT function
Mysql version: 8.0

JSQLParser verison: 4.3

sql example：

```sql
// this is bad sql
SELECT JSON_OBJECT('account',account) as obj FROM tb_team;

// this is success sql
SELECT JSON_OBJECT('account','some string') as obj FROM tb_team;
```

bad sql exception:

image

## DDL

```sql
CREATE TABLE `tb_team` (
  `id` bigint(20) NOT NULL COMMENT '雪花id',
  `process_instance_id` varchar(64) DEFAULT NULL COMMENT '流程id',
  `relevance_id` bigint(20) DEFAULT NULL COMMENT '发布id',
  `apply_id` bigint(20) DEFAULT NULL COMMENT '报名id',
  `account` varchar(32) DEFAULT NULL COMMENT '学工号',
  `originator` varchar(32) DEFAULT NULL COMMENT '发起人',
  `team_type` tinyint(4) DEFAULT NULL COMMENT '类型',
  `member_type` tinyint(4) DEFAULT NULL COMMENT '成员类型',
  `join_type` tinyint(4) DEFAULT NULL COMMENT '加入类型',
  `open_status` tinyint(4) DEFAULT NULL COMMENT '成员状态',
  `creator` varchar(64) DEFAULT NULL COMMENT '创建者',
  `updater` varchar(64) DEFAULT NULL COMMENT '修改者',
  `update_time` datetime DEFAULT NULL COMMENT '修改时间',
  `deleted` bit(1) DEFAULT b'0' COMMENT '删除标志',
  `tenant_id` bigint(20) DEFAULT NULL COMMENT '租户id',
  `create_time` datetime DEFAULT NULL COMMENT '创建时间',
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COMMENT='队伍成员表';
```

image

An exception will occur if the 'account' column comes from the tb_team table

manticore-projects:
Fixes #1504 

Correct parsing the Standard Compliant `JSON_OBJECT()` function, where `KEY` is supposed to be a String
Add support for MySQL `JSON_OBJECT()`, where multiple Key-Value Pairs are separated by Comma `,`

### DIFF
--- a/src/test/java/net/sf/jsqlparser/expression/JsonFunctionTest.java
+++ b/src/test/java/net/sf/jsqlparser/expression/JsonFunctionTest.java
@@ -154,6 +154,15 @@ public class JsonFunctionTest {
     }
 
     @Test
+    public void testObjectWithExpression() throws JSQLParserException {
+        TestUtils.assertSqlCanBeParsedAndDeparsed(
+                "SELECT JSON_OBJECT( KEY 'foo' VALUE cast( bar AS VARCHAR(40)), KEY 'foo' VALUE bar) FROM dual ", true);
+
+        TestUtils.assertSqlCanBeParsedAndDeparsed(
+                "SELECT JSON_ARRAYAGG(obj) FROM (SELECT trt.relevance_id,JSON_OBJECT('id',CAST(trt.id AS CHAR),'taskName',trt.task_name,'openStatus',trt.open_status,'taskSort',trt.task_sort) as obj FROM tb_review_task trt ORDER BY trt.task_sort ASC)", true);
+    }
+
+    @Test
     public void testObjectIssue1504() throws JSQLParserException {
         TestUtils.assertSqlCanBeParsedAndDeparsed(
                 "SELECT JSON_OBJECT(key 'person' value tp.account) obj", true);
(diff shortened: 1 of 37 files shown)


src/test/java/net/sf/jsqlparser/expression/JsonFunctionTest.java
```diff
@@ -131,24 +126,24 @@ public void testArrayAgg() throws JSQLParserException {
     @Test
     public void testObject() throws JSQLParserException {
         TestUtils.assertSqlCanBeParsedAndDeparsed(
-                "SELECT JSON_OBJECT( KEY foo VALUE bar, KEY foo VALUE bar) FROM dual ", true);
-        TestUtils.assertSqlCanBeParsedAndDeparsed("SELECT JSON_OBJECT( foo:bar, foo:bar) FROM dual ",
+                "SELECT JSON_OBJECT( KEY 'foo' VALUE bar, KEY 'foo' VALUE bar) FROM dual ", true);
```
wumpz, line 129: So the test before without those quotes was wrong?

manticore-projects: The documentation is scarce and Oracle mentions as `String` explicitly and all the examples show `KEY` as a `String`.
I do not use those functions myself and only implemented them, when they collided with the special `StringFunctions` and hung the parser.

PR #1534, opened 2022-05-10
Allow isolation keywords as column name and aliases

tomershay:
This allows to parse statements such as:
`SELECT col FROM tbl cs`

### DIFF
--- a/src/test/java/net/sf/jsqlparser/statement/select/SelectTest.java
+++ b/src/test/java/net/sf/jsqlparser/statement/select/SelectTest.java
@@ -2501,6 +2501,12 @@ public class SelectTest {
     }
 
     @Test
+    public void selectIsolationKeywordsAsAlias() throws JSQLParserException {
+        String stmt = "SELECT col FROM tbl cs";
+        assertSqlCanBeParsedAndDeparsed(stmt);
+    }
+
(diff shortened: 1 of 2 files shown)


src/test/java/net/sf/jsqlparser/statement/select/SelectTest.java
```diff
@@ -2500,6 +2500,12 @@ public void testMultiValueIn4() throws JSQLParserException {
         assertSqlCanBeParsedAndDeparsed(stmt);
     }
 
+    @Test
+    public void selectIsolationKeywordsAsAlias() throws JSQLParserException {
+        String stmt = "SELECT col FROM tbl cs";
```
wumpz, line 2505: You do not test your **isolation** usage.

wumpz (changes requested):

src/test/java/net/sf/jsqlparser/statement/select/SelectTest.java
tomershay: There are other tests already testing isolation usage (for example, _testWithIsolation_).
This PR just allows "cs" to be used as a column name / alias, as otherwise parsing fails when it's used.

PR #1545, opened 2022-05-28
compound statement tests

rathboma:
Hey!

I didn't see any tests for anonymous statement blocks so figured I'd add them!

So I added three new tests for compound statements / anonymous blocks
- https://docs.oracle.com/cd/E17952_01/mysql-5.7-en/begin-end.html
- https://blogs.oracle.com/connect/post/building-with-blocks

Note in the SQL files it says that the tests failed...but `mvn test` doesn't seem to show that at all? Not sure what I'm doing wrong, but wanted to make a PR anyway to see if the failures are legit?

### DIFF
--- a/src/test/resources/net/sf/jsqlparser/statement/select/oracle-tests/compound_statements01.sql
+++ b/src/test/resources/net/sf/jsqlparser/statement/select/oracle-tests/compound_statements01.sql
@@ -1,0 +1,32 @@
+---
+-- #%L
+-- JSQLParser library
+-- %%
+-- Copyright (C) 2004 - 2022 JSQLParser
+-- %%
+-- Dual licensed under GNU LGPL 2.1 or Apache License 2.0
+-- #L%
+---
+      DECLARE
+        PK_NAME VARCHAR(200);
+
+      BEGIN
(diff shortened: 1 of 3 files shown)


manticore-projects:
Greetings!

First of all, thanks a lot for interest and contributing. Well appreciated.

On the details, we may need a few changes:

1) You have touched that "Special Oracle Tests" section, which is not advisable. This section is reserved for exotic Oracle features, and some of them are meant/known to fail -- and thus its silent.

2) Instead move your Tests into the normal test section and use the standard `assertCanBeParsedAndDeparsed` method. Do not add new failing tests or missing features without a solution please.

3) Your tests are `BEGIN... END` blocks and should be supported and working. What exactly is different about your examples please?

manticore-projects:
Please have a look at `net/sf/jsqlparser/statement/BlockTest.java`. There are Tests on Block Statements already and your additional Tests could go there.

rathboma:
Ok, great! I'll add them there. I didn't see the block tests, so thought this might be a real contribution. I'll move them to that file this week (hopefully with a fix)

Matthew

Issue #1553, opened 2022-06-06
validate not support 'not rlike'


guandatawangjialin:
String sql = "select * from test where id  not rlike '111'";

ValidationError [
statement=select * from test where id  not rlike '111'
capability=parsing
errors=[ParseException: Cannot parse statement: Encountered unexpected token: "not" "NOT"
    at line 1, column 30.

Was expecting one of:

    "&"
    "&&"
    "::"
    ";"
    "<<"
    ">>"
    "AND"
    "AT"
    "COLLATE"
    "CONNECT"
    "FOR"
    "GROUP"
    "HAVING"
    "OR"
    "START"
    "XOR"
    "["
    "^"
    "|"
    <EOF>
]
]

PR #1554, opened 2022-06-06
Added support for NOT RLIKE expressions

guandatawangjialin:
Added support for NOT RLIKE expressions
#1553

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -3300,7 +3300,7 @@ Expression LikeExpression(Expression leftExpression) #LikeExpression:
     Expression escape;
 }
 {
-    [<K_NOT> { result.setNot(true); } ] ( <K_LIKE> | <K_ILIKE> { result.setCaseInsensitive(true); } ) rightExpression=SimpleExpression()
+    [<K_NOT> { result.setNot(true); } ] ( <K_LIKE> | <K_ILIKE> | <K_RLIKE> { result.setCaseInsensitive(true); } ) rightExpression=SimpleExpression()
     [<K_ESCAPE> escape=Expression() { result.setEscape(escape); }]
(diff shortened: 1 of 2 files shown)


guandatawangjialin:
#1553

manticore-projects:
Greetings!

Thank you very much for your contribution.
Although I have some concerns:

1) You extend the `LikeExpression`, but without defining what variant of `like` has been used.

2) So how will the `Deparser` and/or `toString()` method rebuild the original SQL statement correctly?
```java
public String getStringExpression() {
        -- something needs to be done for RLIKE here
        return caseInsensitive ? "ILIKE" : "LIKE";
    }
```

3) Are you sure, your provided test case ran successfully? Have you tried `gradle build` or `mvn package` before submitting?

src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
```diff
@@ -3300,7 +3300,7 @@ Expression LikeExpression(Expression leftExpression) #LikeExpression:
     Expression escape;
 }
 {
-    [<K_NOT> { result.setNot(true); } ] ( <K_LIKE> | <K_ILIKE> { result.setCaseInsensitive(true); } ) rightExpression=SimpleExpression()
+    [<K_NOT> { result.setNot(true); } ] ( <K_LIKE> | <K_ILIKE> | <K_RLIKE> { result.setCaseInsensitive(true); } ) rightExpression=SimpleExpression()
```
manticore-projects, line 3303: Unfortunately, you break the Logic here:
1) For `<K_LIKE>` only, `caseInsensitive:=false` and
2) For `<K_ILIKE>` only,   `caseInsensitive:=false` 

You set `caseInsensitive` for `<K_ILKE>` instead.

manticore-projects:
Incorporated via https://github.com/manticore-projects/JSqlParser/commit/f132547f56a1edd6e6a0e1454b53ac2e422a835b

PR #1568, opened 2022-06-16
Add test for LikeExpression.setEscape and LikeExpression.getStringExpression

lacinoire:
Hey 😊
I want to contribute a test.
Curious to hear what you think!

(I wrote this test as part of a research study at TU Delft. [Find out more](https://github.com/lacinoire/lacinoire/blob/main/README.md))

### DIFF
--- a/src/test/java/net/sf/jsqlparser/expression/operators/relational/LikeExpressionTest.java
+++ b/src/test/java/net/sf/jsqlparser/expression/operators/relational/LikeExpressionTest.java
@@ -9,6 +9,10 @@
  */
 package net.sf.jsqlparser.expression.operators.relational;
 
+import net.sf.jsqlparser.JSQLParserException;
+import net.sf.jsqlparser.expression.Expression;
+import net.sf.jsqlparser.expression.StringValue;
+import net.sf.jsqlparser.parser.CCJSqlParserUtil;
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.*;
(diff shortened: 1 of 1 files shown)


manticore-projects:
What problem exactly will this PR solve and why should your test become part of the Test Suite please?

The result isn't meaningful or valid.

lacinoire:
Hey 🙂

This tests the methods [`LikeExpression.setEscape`](https://github.com/JSQLParser/JSqlParser/blob/bf8a7db55c82afe5e75cce98f944a88b56af3b5c/src/main/java/net/sf/jsqlparser/expression/operators/relational/LikeExpression.java#L54) and [`LikeExpression.getStringExpression`](https://github.com/JSQLParser/JSqlParser/blob/bf8a7db55c82afe5e75cce98f944a88b56af3b5c/src/main/java/net/sf/jsqlparser/expression/operators/relational/LikeExpression.java#L36), which are not covered by another test as far as I can see.

It is based on / inspired by the test [`testLikeNotIssue660`](https://github.com/JSQLParser/JSqlParser/blob/bf8a7db55c82afe5e75cce98f944a88b56af3b5c/src/test/java/net/sf/jsqlparser/expression/operators/relational/LikeExpressionTest.java#L23).


Would it be better to include a more meaninigful query instead of the 'default' `null LIKE null ESCAPE null LIKE null` ?

Do you find it relevant to test `LikeExpression.setEscape` and `.getStringExpression` at all? 🙂

manticore-projects:
Adding test coverage is definitely good -- when the tests are meaningful and assert something. Test without purpose are considered harmful, since they pretend coverage where no insurance is provided.

Better craft an actual valid Expression using the `LIKE` clause in combination with an `ESCAPE` term -- instead of NULLS.

lacinoire:
Thank you for your input and patience 🙂

I've changed the expression to match all records with names that start with letter ’J’ and have the ’_’ character in them:
`"name LIKE 'J%$_%' ESCAPE '$'"`
(updated the whole test case)

What do you think?

PR #1585, opened 2022-07-06
Closes #1583:: Implement Postgresql optional TABLE in TRUNCATE

### LINKED ISSUE #1583: Postgresql: TRUNCATE does not require TABLE
I see this in scripts: 

`TRUNCATE myfancytable;`

This is allowed by Postgresql:

https://www.postgresql.org/docs/current/sql-truncate.html

```
TRUNCATE [ TABLE ] [ ONLY ] name [ * ] [, ... ]
    [ RESTART IDENTITY | CONTINUE IDENTITY ] [ CASCADE | RESTRICT ]
```

However, the parser complains with: 

```
net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: "myfancytable" <S_IDENTIFIER>
    at line 3, column 10.

Was expecting:

    "TABLE"
```

d2a-raudenaerde:
Implemented Postgresql optional TABLE in TRUNCATE.
Closes #1583

### DIFF
--- a/src/test/java/net/sf/jsqlparser/statement/truncate/TruncateTest.java
+++ b/src/test/java/net/sf/jsqlparser/statement/truncate/TruncateTest.java
@@ -10,12 +10,15 @@
 package net.sf.jsqlparser.statement.truncate;
 
 import java.io.StringReader;
+
 import net.sf.jsqlparser.JSQLParserException;
 import net.sf.jsqlparser.parser.CCJSqlParserManager;
 import net.sf.jsqlparser.schema.Table;
+
 import static net.sf.jsqlparser.test.TestUtils.assertDeparse;
 import static net.sf.jsqlparser.test.TestUtils.assertSqlCanBeParsedAndDeparsed;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+
 import org.junit.jupiter.api.Test;
 
 public class TruncateTest {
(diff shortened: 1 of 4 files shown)


src/test/java/net/sf/jsqlparser/statement/truncate/TruncateTest.java
```diff
@@ -41,6 +41,23 @@ public void testTruncate() throws Exception {
         assertEquals(statement, truncate.toString());
     }
 
+    @Test
+    public void testTruncatePostgresqlWithoutTableName() throws Exception {
```
wumpz, line 45: JSqlParser tries to rebuild the original SQL from its object tree. With your change, there would be always a `table` part. Use `assertSqlCanBeParsedAndDeparsed(statement)` to ensure the rebuilding behaviour. You have to introduce a flag, if the `table` keyword is used or not and then use this in the `toString` method of `Truncate` and the fitting deparser.

wumpz (changes requested):

src/test/java/net/sf/jsqlparser/statement/truncate/TruncateTest.java
d2a-raudenaerde: I see. I updated the PR with the flags (also for `ONLY`). 

btw.
I also noticed there is a bit of an overlap between the `StatementDeparser` `visit(Truncate t)` and the `toString()` of the `Truncate`.  I think you want to keep things loosely coupled, but added a `deparse()` method to the `Statement` interface might be a way to prevent this code duplicating, as the toString() could simply call `deparse()`. The reason I would be tempted do this, is that the project _forces_ you to have a deparse implementation for each statement.

wumpz (approved):

PR #1594, opened 2022-07-19
add support for drop column if exists

rrrship:
Adds support for "IF EXISTS" command in DROP COLUMN statement for PostgreSQL ([alter table docs](https://www.postgresql.org/docs/9.0/sql-altertable.html)).


wumpz:
thx for improving JSqlParser

Issue #1597, opened 2022-07-23
Using own Feature - constant for "delete with returning"


gitmotte:
#1527 introduced DELETE with RETURNING and used ``Feature.insertReturningExpressionList``
within the validation ``net.sf.jsqlparser.util.validation.validator.DeleteValidator`` - the ``Feature.insertReturningExpressionList`` is part of the validation group ``FeaturesAllowed.INSERT`` and therefore only allowed for INSERT statements.

The bugfix fixes the problem with a new feature constant ``Feature.deleteReturningExpressionList`` which is part of the validation group ``FeaturesAllowed.DELETE``.

manticore-projects:
Thank you for helping me out! I honestly have no idea how this validation framework works and tried brute force.

gitmotte:
Are the correlations clearer for you now, when you look at the changes?

gitmotte:
I've added a "Further development" section to the wiki of my fork - could you please merge it to the original wiki of JSQLParser @wumpz ?

https://github.com/gitmotte/JSqlParser/wiki/Examples-of-SQL-Validation/_compare/f2b297d92baa0729a2c49f8b6e54e670b222b38a

wumpz:
merged it :) manually

PR #1598, opened 2022-07-23
Using own Feature - constant for "delete with returning" #1597

gitmotte:
@manticore-projects would you please review the change?


wumpz:
Looked good so I merged it.

PR #1606, opened 2022-08-02
Closes #1604, added simple OVERLAPS support

### LINKED ISSUE #1604: Missing support for SQL operator 'OVERLAPS'
In SQL it is possible to do something like this:

`select 'yes' from dual where (sysdate-5,sysdate) overlaps (sysdate-2,sysdate-1)` (interval5.sql)

JsqlParser currently does not support this. 

I started a Discussion on how it could be implemented:  https://github.com/JSQLParser/JSqlParser/discussions/1600

d2a-raudenaerde:
Create simple support for OVERLAPS with RowValueConstructor (1 out of 3 cases)

### DIFF
--- a/src/main/java/net/sf/jsqlparser/expression/RowTypeConstructor.java
+++ b/src/main/java/net/sf/jsqlparser/expression/RowTypeConstructor.java
@@ -19,9 +19,6 @@ public class RowTypeConstructor extends ASTNodeAccessImpl implements Expression 
     private ArrayList<ColumnDefinition> columnDefinitions = new ArrayList<>();
     private String name = null;
 
-    public RowTypeConstructor() {
-    }
-    
     public ArrayList<ColumnDefinition> getColumnDefinitions() {
         return columnDefinitions;
     }
--- a/src/main/java/net/sf/jsqlparser/expression/RowTypeConstructor.java
+++ b/src/main/java/net/sf/jsqlparser/expression/RowTypeConstructor.java
@@ -14,12 +14,12 @@ import net.sf.jsqlparser.expression.operators.relational.ExpressionList;
 import net.sf.jsqlparser.parser.ASTNodeAccessImpl;
 import net.sf.jsqlparser.statement.create.table.ColumnDefinition;
 
-public class RowConstructor extends ASTNodeAccessImpl implements Expression {
+public class RowTypeConstructor extends ASTNodeAccessImpl implements Expression {
     private ExpressionList exprList;
     private ArrayList<ColumnDefinition> columnDefinitions = new ArrayList<>();
     private String name = null;
 
-    public RowConstructor() {
+    public RowTypeConstructor() {
     }
     
     public ArrayList<ColumnDefinition> getColumnDefinitions() {
@@ -30,6 +30,12 @@ public class RowConstructor extends ASTNodeAccessImpl implements Expression {
         return columnDefinitions.add(columnDefinition);
     }
 
+    public void addExpression( Expression expr) {
+        if (exprList == null ) {
+            exprList = new ExpressionList();
+        }
+        exprList.addExpressions(expr);
+    }
     public ExpressionList getExprList() {
         return exprList;
     }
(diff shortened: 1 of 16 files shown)


src/main/java/net/sf/jsqlparser/expression/RowTypeConstructor.java
```diff
@@ -14,14 +14,11 @@
 import net.sf.jsqlparser.parser.ASTNodeAccessImpl;
 import net.sf.jsqlparser.statement.create.table.ColumnDefinition;
 
-public class RowConstructor extends ASTNodeAccessImpl implements Expression {
+public class RowTypeConstructor extends ASTNodeAccessImpl implements Expression {
```
wumpz, line 17: Why did you rename an existing class? This will break code already running on JSqlParser.

src/main/java/net/sf/jsqlparser/util/deparser/ExpressionDeParser.java
```diff
@@ -367,7 +374,7 @@ public void visit(SubSelect subSelect) {
         if (selectVisitor != null) {
             if (subSelect.getWithItemsList() != null) {
                 buffer.append("WITH ");
-                for (Iterator<WithItem> iter = subSelect.getWithItemsList().iterator(); iter.hasNext();) {
+                for (Iterator<WithItem> iter = subSelect.getWithItemsList().iterator(); iter.hasNext(); ) {
```
wumpz, line 377: This reformatting introduced a lot of changes. Reformating should be done in a different PR.

src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
```diff
@@ -3300,13 +3301,24 @@ Expression Condition():
 {
     [ LOOKAHEAD(2) (<K_NOT> { not=true; } | "!" { not=true; exclamationMarkNot=true; })]
     (
-        LOOKAHEAD(RegularCondition()) result=RegularCondition()
+        LOOKAHEAD(OverlapsCondition()) result=OverlapsCondition()
```
wumpz, line 3304: Shouldn't this go into SQLCondition?

wumpz (changes requested):

src/main/java/net/sf/jsqlparser/expression/RowTypeConstructor.java
manticore-projects: I second that, please lets not rename Classes without a very strong reason and without any deprecation approach.
It would break code, examples and documentation -- for a rather exotic new Expression.

manticore-projects:
```sql
select 'yes' from dual where (sysdate-5,sysdate) overlaps (sysdate-2,sysdate-1)
```

Based on the example above, I do see:
```
"(" <EXPRESSION> "," <EXPRESSION> ")" <OVERLAPS> "(" <EXPRESSION> "," <EXPRESSION> ")"
```

Maybe a:
```
<EXPRESSION_LIST> <OVERLAPS> <EXPRESSION_LIST>
```

Please lets not tamper with `RowConstructor` here, it is not a SQL Row at all. Completely different semantics.

d2a-raudenaerde: Thanks for the feedback, I did not really take into consideration the effect of the rename. I will change it!

d2a-raudenaerde: There is no such thing as an 'SQL Row' in the sql grammar? They explicitly have two 'types' of row-constructs, one that defines a row-type (used in casts etc), and one that defines values.

src/main/java/net/sf/jsqlparser/util/deparser/ExpressionDeParser.java
d2a-raudenaerde: Agreed.

src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
d2a-raudenaerde: That is a better place, yes.

d2a-raudenaerde: I'll close this PR and open a new one

PR #1653, opened 2022-10-26
Keywords2: Update whitelisted Keywords

manticore-projects:
Updates the Keywords in `RelObjectNameWithoutValue` after PR #1649 introduced new tokens.

Workflow should be like that:

1) Amend Tokens in the Grammar
2) When keyword-tests fail, run `gradle updateKeywords`
3) Then build via `mvn package` or `gradle build` (it does not matter)
4) Push/Deploy

Using `gradle updateKeywords` task is optional, you can also add the new Tokens manually has it has always been done before.

After accepting the Keywords PR, any further PR will automatically fail acceptance tests, unless keywords are in order.


wumpz:
Since I have accepted those enhanced keywords, why is this pr so large?

manticore-projects:
> Since I have accepted those enhanced keywords, why is this pr so large?

It actually changes only 4 Files:
1) the Grammar, fixing the `RelObjectNameWithoutValue()` keywords
2) the Maven Build file, adding a Task for updating the `RelObjectNameWithoutValue()` keywords and dependencies
3) the Gradle Build file, updating dependencies
4) Oracle Special Tests, since Keywords enabled 2 more tests

I do not know, why GIT carries all the change set since branching, after I merged MASTER.

pom.xml
```diff
@@ -114,6 +116,18 @@
 
     <build>
         <plugins>
+            <!-- Update they Keyword Whitelist in the Grammar's production RelObjectNameWithoutValue() -->
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
```
wumpz, line 121: So how does this actually is processed? I am not sure how this exec-maven-plugin is activated in the maven build process?

wumpz: got it from the other PR

PR #1690, opened 2022-12-11
Fix #1686: add support for creating views with "IF NOT EXISTS" clause

### LINKED ISSUE #1686: Parsing "CREATE MATERIALIZED VIEW IF NOT EXISTS ..." is not supported
Parsing a string creating materialized view with "IF NOT EXISTS" clause fails with an error. It appears that `Feature.createViewMaterializedIfNotExists` is missing, which would be similar to `createViewMaterialized` and `createTableIfNotExists`.

zaza:


manticore-projects:
Greetings!

Thank you for your contribution. It looks great and complete. Kudos!
Maybe add a description to the PR and also link it to the issue using `Fixes #1686`. This way the issue will be closed automatically when the PR gets accepted.

Although only @wumpz can accept PRs.

PR #1725, opened 2023-02-07
Adding support for PostgreSQL NOTNULL and NOT ISNULL

tomershay:
JSQLParser currently supports expressions such as ISNULL, IS NULL, and IS NOT NULL, but the following expressions weren't yet supported: NOTNULL, and NOT ISNULL. Adding support for full coverage of all options, which can be used in various places, such as the FILTER section of the SELECT clause.

https://www.postgresql.org/docs/current/functions-comparison.html


manticore-projects:
@tomershay Thanks for your contribution, I have incorporated it. Sorry for late response.

PR #1816, opened 2023-06-29
feat: add support for INTERPRET function parsing

matteosist:
Update JSqlParserCC.jjt adding properly definition Added InterpretExpression java class
Add visit function implementation to classes that implements ExpressionVisitor Added InterpretExpression tests

BREAKING CHANGE: No


manticore-projects:
Greetings!

Thank you your contribution and interest in JSQLParser.
One question though: what is the difference between a `Cast()` function and an `Interpret() function (except the keyword `Interpret`). As far as I can see it, `Interpret` is a RDBMS specific `Cast()`, am I right?

If so, would it not be better to just to extend the existing `Cast` function to work with the `Interprete` keyword?
Saved us a lot of code.

src/main/java/net/sf/jsqlparser/util/TablesNamesFinder.java
```diff
@@ -9,58 +9,7 @@
...
-import net.sf.jsqlparser.expression.TimeValue;
-import net.sf.jsqlparser.expression.TimestampValue;
-import net.sf.jsqlparser.expression.TimezoneExpression;
-import net.sf.jsqlparser.expression.TranscodingFunction;
-import net.sf.jsqlparser.expression.TrimFunction;
-import net.sf.jsqlparser.expression.UserVariable;
-import net.sf.jsqlparser.expression.VariableAssignment;
-import net.sf.jsqlparser.expression.WhenClause;
-import net.sf.jsqlparser.expression.XMLSerializeExpr;
+import net.sf.jsqlparser.expression.*;
```
manticore-projects, line 12: Please do not use Star Imports, we prefer simple single class imports.

src/main/java/net/sf/jsqlparser/util/deparser/ExpressionDeParser.java
```diff
@@ -9,59 +9,7 @@
...
-import net.sf.jsqlparser.expression.TimestampValue;
-import net.sf.jsqlparser.expression.TimezoneExpression;
-import net.sf.jsqlparser.expression.TranscodingFunction;
-import net.sf.jsqlparser.expression.TrimFunction;
-import net.sf.jsqlparser.expression.UserVariable;
-import net.sf.jsqlparser.expression.VariableAssignment;
-import net.sf.jsqlparser.expression.WhenClause;
-import net.sf.jsqlparser.expression.WindowElement;
-import net.sf.jsqlparser.expression.XMLSerializeExpr;
+import net.sf.jsqlparser.expression.*;
```
manticore-projects, line 12: Please do not use Star Imports, we prefer simple single class imports.

src/main/java/net/sf/jsqlparser/util/validation/validator/ExpressionValidator.java
```diff
@@ -9,60 +9,7 @@
...
-import net.sf.jsqlparser.expression.TranscodingFunction;
-import net.sf.jsqlparser.expression.TrimFunction;
-import net.sf.jsqlparser.expression.UserVariable;
-import net.sf.jsqlparser.expression.VariableAssignment;
-import net.sf.jsqlparser.expression.WhenClause;
-import net.sf.jsqlparser.expression.WindowElement;
-import net.sf.jsqlparser.expression.WindowOffset;
-import net.sf.jsqlparser.expression.WindowRange;
-import net.sf.jsqlparser.expression.XMLSerializeExpr;
+import net.sf.jsqlparser.expression.*;
```
manticore-projects, line 12: Please do not use Star Imports, we prefer simple single class imports.

manticore-projects (changes requested):
Looks good in general, but I think it would be better merged into the existing `Cast()` function since only the keyword seems to be different.
Saves a lot of code for the Visitors and keeps the Grammar slim.

src/main/java/net/sf/jsqlparser/util/TablesNamesFinder.java
matteosist: Sorry, IDE automatically replaced the specific import with star import. Let me fix.

manticore-projects:
Thanks! Looks good now and I would merge it when the QA tests will have ran through.

manticore-projects:
Thank you again for your contribution, merged.

matteosist:
Fantastic. Thank you for this very useful library

PR #1823, opened 2023-07-08
add profile 'only-eclipse' with  org.eclipse.m2e plugin

10km:
add profile 'only-eclipse' with  org.eclipse.m2e plugin,so then this plugin be active only while opening in eclipse,
the org.eclipse.m2e plugin be used to suppressing IDE warning in eclipse IDE


manticore-projects:
Thank you for your effort, although I don't see why we should merge that upstream.
The project compiles perfectly fine with Maven and Gradle and this is sufficient for almost all use cases. 

If for whatever reason Eclipse has a problem with it, then maybe Eclipse needs to find a way around without tampering with the sources or build scripts.

manticore-projects:
Any feedback please? I would like top close this otherwise.

PR #1837, opened 2023-07-26
Fixing a problem with an OP_CONCAT in WhenExpression

amigalev:
Version: 4.7-SNAPSHOT

The problem was in an ELSE part of the WHEN expression. When we had a sub WHEN in ELSE and concatenate it with some another string:
`
CASE 
    WHEN 1 = 2 
    THEN 'string1'
    ELSE 
        CASE
        WHEN 2 = 3
        THEN 'string2'
        ELSE 'first part'
        END || 'second part'
END 
`

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -4796,7 +4796,7 @@ Expression CaseWhenExpression() #CaseWhenExpression:
     [
         <K_ELSE>
         (
-              LOOKAHEAD(3, {getAsBoolean(Feature.allowComplexParsing) && !interrupted}) elseExp=Expression()
+              LOOKAHEAD({getAsBoolean(Feature.allowComplexParsing) && !interrupted}) elseExp=Expression()
               | elseExp=SimpleExpression()
         )
     ]
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -4796,7 +4796,8 @@ Expression CaseWhenExpression() #CaseWhenExpression:
     [
         <K_ELSE>
         (
-             LOOKAHEAD(3, {!interrupted}) "(" elseExp=CaseWhenExpression() ")" { elseExp = new Parenthesis( elseExp ); }
+            LOOKAHEAD(ConcatExpression()) elseExp = ConcatExpression() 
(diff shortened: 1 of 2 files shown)


manticore-projects:
Thank you for your contribution.
Existing Tests fail, please ensure the complete test suite is running through before sending PRs.

amigalev:
> Thank you for your contribution. Existing Tests fail, please ensure the complete test suite is running through before sending PRs.
> 
> If also do not understand why you would like to catch `ConcatExpression` first when there is already a general `Expression` in the second branch.

I'm sorry for tests, for some reason some tests have been failed in my environment. I'll try to fix it asap. Which build system should I use? Gradle or Maven?

About your second question. 
The reason is that in an ELSE part we check use LOOKAHEAD(3, {!interrupted}) and as I understand the execution will go throught  `LOOKAHEAD(3, {!interrupted}) elseExp=CaseWhenExpression()` insead of `elseExp=SimpleExpression()`. To avoid an expensive LOOKAHEADs I decided to add additional ConcatExpression (the `LOOKAHEAD(ConcatExpression())` looks cheaper.).

If I understand incorrectly, please tell me and I'll find another solution

amigalev:
I've fixed tests, but the solution of my core problem has changed. I think, It will be a bit slower then previous, but THEN part in `WhenThenSearchCondition` works similar.

The problem has been occurred because the `ConcatExpression` (and a lot of other) uses * instead of + and catches single `S_IDENTIFIER` instead of full concat expression. I've tried to change this behaviour, but many parts of grammar use it.

amigalev:
You're right. I forgot to delete `LOOKAHEAD(3)` and it  this will be exactly the same as part `WHEN`. Will this snippet good or do you mean sth else?

```
        <K_ELSE>
        (
              LOOKAHEAD({getAsBoolean(Feature.allowComplexParsing) && !interrupted}) elseExp=Expression()
              | elseExp=SimpleExpression()
        )
```

amigalev:
Could you explain me, why did you close the PR? I really need the fix this problem. If my fix is not correct by performance reasons I'll try to fix it in some another way. 
I thought if you use the same way for the `THEN` part, it will be ok for the `ELSE` part too (why not?). This is snippet for `THEN`:

```
<K_THEN>
    (
        LOOKAHEAD({getAsBoolean(Feature.allowComplexParsing) && !interrupted}) thenExp=Expression()
        |
        thenExp=SimpleExpression()
    ) 
```

As an example I can try to rewrite the concat function to solve this problem. It will be a bit more complex but it will work faster

manticore-projects:
Apologies, I actually believed to have MERGED it!
At least this was what I wanted to do.

PR #1886, opened 2023-11-08
Unit tests support multi-os and higher versions of jdk

human-user:
When I run unit tests, there are two failures:
1. ParserKeywordsUtilsTest#getAllKeywords failed to run. Because my os is windows and the regex in ParserKeywordsUtils#getAllKeywordsUsingRegex seems not support multi-os.
2. Compilation fails when using JDK 17. Nashorn has been deprecated after jdk11 and removed after jdk15.


manticore-projects:
Greetings!
Thank you for your contribution, especially the fix of the Regex is great.

PR #1887, opened 2023-11-08
feat: add support for snowflake merge statements

### LINKED ISSUE #1315: Snowflake MERGE statement does not require parentheses around the ON statement
**Describe the bug**
The Snowflake MERGE statement does not require the ON statement to be wrapped in parantheses.

**To Reproduce**
1. Example SQL
```
MERGE INTO CONSULTANT T USING (
    SELECT
        *
    from
        CONSULTANT
) S ON S.ID = T.ID
WHEN MATCHED THEN
UPDATE
SET
    T.RESOURCE = S.RESOURCE
    WHEN NOT MATCHED THEN
INSERT
    (ID, RESOURCE)
values
    (S.ID, S.RESOURCE);
```

2. Parsing this SQL using JSqlParser with these statements
```
String text = """
MERGE INTO CONSULTANT T USING (
    SELECT
        *
    from
        CONSULTANT
) S ON S.ID = T.ID
WHEN MATCHED THEN
UPDATE
SET
    T.RESOURCE = S.RESOURCE
    WHEN NOT MATCHED THEN
INSERT
    (ID, RESOURCE)
values
    (S.ID, S.RESOURCE);
"""
Statement query = CCJSqlParserUtil.parse(text)
```

3. Exception

```
net.sf.jsqlparser.JSQLParserException: Encountered unexpected token: "S" <S_IDENTIFIER>
    at line 7, column 8.

Was expecting:

    "("
```

**Expected behavior**
When the parentheses are included around the ON, this works:
```
String text = """
MERGE INTO CONSULTANT T USING (
    SELECT
        *
    from
        CONSULTANT
) S ON (S.ID = T.ID)
WHEN MATCHED THEN
UPDATE
SET
    T.RESOURCE = S.RESOURCE
    WHEN NOT MATCHED THEN
INSERT
    (ID, RESOURCE)
values
    (S.ID, S.RESOURCE);
"""
Statement query = CCJSqlParserUtil.parse(text)
```

**System**
 - Database you are using: Snowflake
- Java Version: 11
- JSqlParser version: 4.1

### LINKED ISSUE #1880: [FEATURE] MERGE  ON without ()
Hello, when I use JSqlParser to parse 'MERGE' statement and there are missing parentheses after 'ON,' it throws an error. 

### SQL Example
merge into table1 USING ( select * from table2) c on table1.aa=c.aa WHEN NOT MATCHED THEN INSERT ( bb,aa ) VALUES (c.aa, c.bb ,c.cc)

davidjgoss:
This PR changes the parser to account for some Snowflake-specific aspects of the `MERGE` statement, namely:

- Tolerance for the `ON` condition not being wrapped in parentheses
  - fixes #1315
  - fixes #1880
- Support for an optional `AND` predicate after the `MATCHED`/`NOT MATCHED` part

See Snowflake's docs on MERGE: https://docs.snowflake.com/en/sql-reference/sql/merge - the new tests are derived from that doc.

- - -

With the `ON` logic, I did the simplest parser change which for existing cases (i.e. with parens), will now result in the `onCondition` being a `Parenthesis` containing the expression, where before it would have just been the expression (e.g. `EqualsTo` or whatever) itself. This feels innocuous to me, but if we consider this a breaking change and would prefer to avoid it, I can refactor to ensure full backwards compatibility, perhaps by adding a choice point in the parser and retaining the presence/absence of original parens as a boolean.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/merge/Merge.java
+++ b/src/main/java/net/sf/jsqlparser/statement/merge/Merge.java
@@ -186,9 +186,8 @@ public class Merge implements Statement {
         b.append(table);
         b.append(" USING ");
         b.append(fromItem);
-        b.append(" ON (");
+        b.append(" ON ");
         b.append(onCondition);
-        b.append(")");
 
         if (insertFirst && mergeInsert != null) {
             b.append(mergeInsert);
--- a/src/main/java/net/sf/jsqlparser/statement/merge/MergeInsert.java
+++ b/src/main/java/net/sf/jsqlparser/statement/merge/MergeInsert.java
@@ -20,9 +20,18 @@ import java.util.Optional;
 
 public class MergeInsert implements Serializable {
 
+    private Expression andPredicate;
     private ExpressionList<Column> columns;
     private ExpressionList<Expression> values;
     private Expression whereCondition;
+
+    public Expression getAndPredicate() {
+        return andPredicate;
+    }
+
+    public void setAndPredicate(Expression andPredicate) {
+        this.andPredicate = andPredicate;
+    }
 
     public ExpressionList<Column> getColumns() {
         return columns;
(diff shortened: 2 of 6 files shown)


davidjgoss:
Thanks for merging @manticore-projects! Do you know when this might be released?

manticore-projects:
This depends fully on @wumpz who is under serious time constraints.
We could aim for an EoY release though.

PR #1933, opened 2023-12-23
Fix: DESC and EXPLAIN

### LINKED ISSUE #1931: [FEATURE] cannot parse DESC table_name
- DESC table_name

### SQL Example
`DESC table_name`

ERROR MESSAGE
```
net.sf.jsqlparser.JSQLParserException: net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: "DESC" "DESC"
    at line 1, column 1.

Was expecting one of:

    "("
    "ALTER"
    "ANALYZE"
    "BEGIN"
    "CALL"
    "COMMENT"
    "COMMIT"
    "CREATE"
    "DECLARE"
    "DELETE"
    "DESCRIBE"
    "DROP"
    "EXEC"
    "EXECUTE"
    "EXPLAIN"
    "GRANT"
    "IF"
    "INSERT"
    "MERGE"
    "PURGE"
    "RENAME"
    "REPLACE"
    "RESET"
    "ROLLBACK"
    "SAVEPOINT"
    "SET"
    "SHOW"
    "TRUNCATE"
    "UPDATE"
    "UPSERT"
    "USE"
    "VALUE"
    "VALUES"
    "WITH"
    <K_SELECT>
```

### Additional context
`DESC TABLE` and `EXPALIN TABLE` cannot be parsed. 
Only `DESCRIBE TABLE` is successfully parsed, although they are all synonyms to each other. (in MySQL)
Thank you very much.

jxnu-liguobin:
close #1931, to support MySQL `{EXPLAIN | DESCRIBE | DESC}  tbl_name` 
it is necessary to use LOOKAHEAD(3), It made `explain01.sql` change. It seems to be expected?

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -1056,29 +1056,43 @@ PurgeStatement PurgeStatement(): {
 
 DescribeStatement Describe(): {
     Table table;
+    DescribeStatement stmt = new DescribeStatement();
+    Token tk = null;
 } {
-    <K_DESCRIBE> table = Table()
+    (tk=<K_DESCRIBE> | tk=<K_DESC>) 
+    table = Table() { stmt.setDescribeType(tk.image).setTable(table); }
     {
-        return new DescribeStatement(table);
+        return stmt;
     }
 }
 
 ExplainStatement Explain(): {
     Select select;
+    Table table = null;
     List<ExplainStatement.Option> options = null;
+    ExplainStatement es = new ExplainStatement();
 } {
     <K_EXPLAIN>
-    options=ExplainStatementOptions()
-    select = Select( )
-    {
-       ExplainStatement es = new ExplainStatement(select);
-       if(options != null && !options.isEmpty()) {
-         for(ExplainStatement.Option o : options) {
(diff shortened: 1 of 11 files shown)


src/test/java/net/sf/jsqlparser/util/validation/validator/StatementValidatorTest.java
```diff
@@ -53,7 +61,8 @@ public void testValidateBlock() throws JSQLParserException {
     @Test
```
manticore-projects, line 61: Next time please use a Parametrised Test for this. Loops in Tests are an Anti-Pattern because you won't know which one failed. You can lookup the KeywordTests for getting a template.

src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
```diff
@@ -1056,29 +1056,43 @@ PurgeStatement PurgeStatement(): {
 
 DescribeStatement Describe(): {
     Table table;
+    DescribeStatement stmt = new DescribeStatement();
```
manticore-projects, line 1059: Next time, create the statement at the end only after all tokens have parsed and passed, like: 

```
return new Statement().withX(token.image).withY(token.image);

```

Its not a hard requirement, but in general Objects should be initialized only when needed.

PR #1941, opened 2024-01-06
feat: Build with Automatic-Module-Name for compatibility with the Java module system.

broneill:
I use the Java module system with my project, but JSQLParser doesn't use modules. As a result, I need to use the --add-reads ALL-UNNAMED hack in order to depend on it.

Building a project with the Automatic-Module-Name feature allows it to work with the module system, but without actually depending on the module system itself. This change makes it easier to integrate JSQLParser with projects that are using the Java module system.

I ran both the Maven and Gradle builds and verified that `Automatic-Module-Name: net.sf.jsqlparser` appears in the jar's MANIFEST.MF file.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -141,6 +141,14 @@ javadoc {
     options.addBooleanOption("Xdoclint:none", true)
 }
 
+jar {
+    manifest {
+        attributes (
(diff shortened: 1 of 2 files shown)


manticore-projects:
It does not harm, I guess.

PR #1944, opened 2024-01-10
fix: make analytic expression visitor null-safe

davidjgoss:
This PR adds some null safety via `Optional`s to the `ExpressionVisitorAdapter` when handling certain analytic expressions.

The added test expression previously parsed fine but resulted in a `NullPointerException` when visited.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitorAdapter.java
+++ b/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitorAdapter.java
@@ -64,6 +64,8 @@ import net.sf.jsqlparser.statement.select.SelectVisitor;
 import net.sf.jsqlparser.statement.select.UnPivot;
 import net.sf.jsqlparser.statement.select.WithItem;
 
+import java.util.Optional;
+
 @SuppressWarnings({"PMD.CyclomaticComplexity", "PMD.UncommentedEmptyMethodBody"})
 public class ExpressionVisitorAdapter
         implements ExpressionVisitor, PivotVisitor, SelectItemVisitor {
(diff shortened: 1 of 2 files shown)


manticore-projects:
Can we please conclude on this? I don't like stale PRs.

davidjgoss:
Sorry @manticore-projects - what is this PR missing?

manticore-projects:
> Sorry @manticore-projects - what is this PR missing?

As explained before:

_Apologies for being difficult: Without the original expression, I have actually no idea what those code lines want or achieve (simply because I never understood or liked the streaming/functional API).

Since I can't stop you from using it (its valid after all), please keep at least the original code lines as comments so we have a clue what this is supposed to mean and to check.

Question: the Null test is done rather beautifully in Groovy and much easier to read.
Does anyone know a similar way avoiding the Streaming API?
_

davidjgoss:
Ah, I don't see that earlier comment for some reason.

> I can't stop you from using it

I mean, you definitely can. It's not in keeping with the style in the rest of the project so I'm more than happy to change it to a more conventional null check, and add some commentary.

manticore-projects:
Stopping you is not in my interest and your code is correct and good.
Just add comments to support understanding as the syntax is rather complex and not self explaining (at least to me).

Thanks a lot!

manticore-projects:
Run './gradlew :spotlessApply' to fix these violations.

manticore-projects (approved):
Thanks!

davidjgoss:
(fixed spotless, sorry about that)

manticore-projects:
Thank you for your contribution and effort!

Issue #1875, opened 2023-10-11
[BUG] JSQLParser 4.7: `ADD COLUMN IF NOT EXISTS` not supported


zdhuangelephant:
Always check against the **Latest SNAPSHOT of JSQLParser** and the [Syntax Diagram](https://jsqlparser.github.io/JSqlParser/syntax_snapshot.html)

### Failing SQL Feature:
- Brief description of the failing SQL feature
- Example: `Encountered unexpected token: "IF" "IF"` can't be parsed
- run java `static void buildCCJSqlParser() throws IOException, JSQLParserException {

        final Path path = Paths.get("/Users/xxx/xxx", "xxx-xxx-comm.sql");
        final byte[] bytes = Files.readAllBytes(path);

        ExecutorService executorService = Executors.newSingleThreadExecutor();

        final CCJSqlParser ccjSqlParser = CCJSqlParserUtil.newParser(new ByteArrayInputStream(bytes), StandardCharsets.UTF_8.name());
        final Statements statements = CCJSqlParserUtil.parseStatements(ccjSqlParser, executorService);

        final int[] updateCounter = {0};
        final int[] alterCounter = {0};
        final int[] selectCounter = {0};
        StatementVisitorAdapter adapter = new StatementVisitorAdapter(){
            @Override
            public void visit(Update update) {
                final String name = update.getTable().getName();
                final String schemaName = update.getTable().getSchemaName();

                updateCounter[0]++;
                System.out.printf("Update: table:%s, schema:%s%n", name, schemaName);
            }
            @Override
            public void visit(Alter alter) {
                final String name = alter.getTable().getName();
                final String schemaName = alter.getTable().getSchemaName();

                alterCounter[0]++;
                System.out.printf("Update: table:%s, schema:%s%n", name, schemaName);
            }
        };
        statements.accept(adapter);

        System.out.println(updateCounter[0] + ", " + alterCounter[0] + ", " + selectCounter[0]);
    }`

### SQL Example:
- Simplified Query Example, focusing on the failing feature
    ```sql
    -- Replace with your ACTUAL example
    ALTER TABLE IF EXISTS usercenter.dict_surgeries ADD COLUMN IF NOT EXISTS operation_grade_id int8 NULL;
    ```

### Software Information:
- JSqlParser version 4.7
- Database (PostgreSQL )

### Tips:
Please write in English and avoid Screenshots (as we can't copy and paste content from it).
[...]

PR #1957, opened 2024-02-07
fix: issue1875

SheldonKubor:
Temporarily solve the issue 1875 #1875 . 
Refactoring may take more time. And in my opinion, we should discuss how to refactor.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/alter/Alter.java
+++ b/src/main/java/net/sf/jsqlparser/statement/alter/Alter.java
@@ -90,10 +90,6 @@ public class Alter implements Statement {
             b.append("IF EXISTS ");
         }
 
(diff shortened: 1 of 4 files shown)


manticore-projects:
> we should discuss how to refactor.

The DDL statements are a big mess, best illustrated here: https://manticore-projects.com/JSQLParser/syntax_snapshot.html#createtable
The syntax diagram does not even fit on the page.

I would like to advise for a refactoring which splits this one into many smaller, re-usable productions where grammar and code can be shared between `CREATE` and `ALTER` as much as possible -- based on the SQL:2016 standard.

A good blueprint maybe here: https://manticore-projects.com/SQL2016Parser/syntax.html#table-definition and https://manticore-projects.com/SQL2016Parser/syntax.html#alter-table-statement

manticore-projects:
Clean and neat as always, thanks a lot for working on this. Cheers!

PR #1965, opened 2024-02-14
Guard Values against null/empty values

pingpingy1:
This PR intends to implement suitable guards for the constructors of the classes `DoubleValue`, `LongValue`, and `TimeValue`.
Both `null` and empty strings provided to their constructors fail, but they provide very different error messages (NullPointerException and StringIndexOutOfBoundsException),
which is neither sensible nor helpful in debugging.
This PR adds a guard to throw `IllegalArgumentException` for both cases in order to improve coherency and usefulness of the error messages.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/expression/DoubleValue.java
+++ b/src/main/java/net/sf/jsqlparser/expression/DoubleValue.java
@@ -24,6 +24,9 @@ public class DoubleValue extends ASTNodeAccessImpl implements Expression {
     }
 
     public DoubleValue(final String value) {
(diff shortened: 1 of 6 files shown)


manticore-projects:
Looks good.
Please fix the formatting by running:
```sh
./gradlew :spotlessApply
```

and then I would merge.

pingpingy1:
@manticore-projects Thank you for the swift response.
I believe that should have fixed it; PTAL.

manticore-projects:
Thank you for your contribution, much appreciated!

PR #1972, opened 2024-02-23
Handle select in ExpressionVisitorAdapter

sivaraam:
The ExpressionVisitorAdapter could handle Select to simplify the implementation of its children.

I've suggested an implementation that extracts portions from the ParenthesedSelect handling. Do take a look and let me know if this is fine.


manticore-projects:
Thank you for your contribution, good point!

PR #1985, opened 2024-03-31
feat: Salesforce SOQL `INCLUDES` and `EXCLUDES` operators

lucarota:
Hi all,

This pull request introduces the INCLUDES and EXCLUDES operators for parsing the Salesforce Object Query Language (SOQL), a variant of SQL specifically designed for querying Salesforce data.

The INCLUDES operator allows users to query for records where a picklist or multipicklist field includes one or more values from a specified set.

The EXCLUDES operator, on the other hand, enables users to query for records where a specified field does not include any values from a specified set. 

The syntax for using the INCLUDES and EXCLUDES operators in SOQL is similar as the standard IN and NOT IN as follows:

INCLUDES:

`SELECT fields FROM object WHERE field INCLUDES (value1, value2, ...)`

EXCLUDES:
`SELECT fields FROM object WHERE field EXCLUDES (value1, value2, ...)`

- The INCLUDES and EXCLUDES operators are case-insensitive.
- Multiple values can be provided within the parentheses, separated by commas.
- These operators can only be used with fields of type Picklist or Multi-Select Picklist.

Ciao,
 Luca


manticore-projects:
Clean! Thank you very much for your contribution, kudos!

manticore-projects:
Please fix the formatting, run './gradlew :spotlessApply' to fix these violations. Then I would merge.

Issue #2008, opened 2024-05-19
DoubleValue.setValue does not set stringValue


Qnzvna:
When using `setValue` on `DoubleValue` object only the `value` is set. The setter does not change `stringValue` which is then used in `toString` method.

Following test would fail:
```
    @Test
    public void shouldSetStringValue() {
        final DoubleValue doubleValue = new DoubleValue("42");

        doubleValue.setValue(43D);

        assertEquals(43D, doubleValue.getValue());
        assertEquals("43.0", doubleValue.toString());
    }
```

Qnzvna:
Thanks, I've opened PR -> https://github.com/JSQLParser/JSqlParser/pull/2009, so you can review.

Qnzvna:
@manticore-projects Thanks for reviewing. When you will be able to release a new version to maven ?

manticore-projects:
Greetings, we don't have any exact date for a new release 5.0 but my take is: before End of June.
Please note that as a matter of fact, JSQLParser has become a rolling-release where every last snapshot is deemed to be the latest greatest version and "stable release" does not bear any actual advantage.

PR #2009, opened 2024-05-19
fix: set `stringValue` in `DoubleValue.setValue`

Qnzvna:


manticore-projects:
Awesome, thank you for your contribution!

manticore-projects:
fixes #2008

PR #2010, opened 2024-05-23
support custom where sql construction logic

redkale:
I hope that XxxDeParser can overload this part of the logic when building the where condition sql statement. My requirement is to dynamically build the filter conditions in where according to the parameters.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/util/deparser/DeleteDeParser.java
+++ b/src/main/java/net/sf/jsqlparser/util/deparser/DeleteDeParser.java
@@ -108,7 +108,7 @@ public class DeleteDeParser extends AbstractDeParser<Delete> {
     protected void deparseWhereClause(Delete delete) {
         if (delete.getWhere() != null) {
             buffer.append(" WHERE ");
-			int len = buffer.length();
+            int len = buffer.length();
             delete.getWhere().accept(expressionVisitor);
             if (buffer.length() == len) {
                 buffer.delete(len - " WHERE ".length(), len);
--- a/src/main/java/net/sf/jsqlparser/util/deparser/DeleteDeParser.java
+++ b/src/main/java/net/sf/jsqlparser/util/deparser/DeleteDeParser.java
@@ -90,14 +90,7 @@ public class DeleteDeParser extends AbstractDeParser<Delete> {
             }
         }
 
-        if (delete.getWhere() != null) {
-            buffer.append(" WHERE ");
-            int len = buffer.length();
-            delete.getWhere().accept(expressionVisitor);			
-            if (buffer.length() == len) {
-                buffer.delete(len - " WHERE ".length(), len);
-            }
-        }
+        deparseWhereClause(delete);
 
         if (delete.getOrderByElements() != null) {
             new OrderByDeParser(expressionVisitor, buffer).deParse(delete.getOrderByElements());
@@ -112,6 +105,13 @@ public class DeleteDeParser extends AbstractDeParser<Delete> {
 
     }
 
+    protected void deparseWhereClause(Delete delete) {
+        if (delete.getWhere() != null) {
+            buffer.append(" WHERE ");
+            delete.getWhere().accept(expressionVisitor);
+        }
+    }
+    
     public ExpressionVisitor getExpressionVisitor() {
         return expressionVisitor;
     }
--- a/src/main/java/net/sf/jsqlparser/util/deparser/DeleteDeParser.java
+++ b/src/main/java/net/sf/jsqlparser/util/deparser/DeleteDeParser.java
@@ -108,7 +108,11 @@ public class DeleteDeParser extends AbstractDeParser<Delete> {
     protected void deparseWhereClause(Delete delete) {
         if (delete.getWhere() != null) {
             buffer.append(" WHERE ");
+			int len = buffer.length();
             delete.getWhere().accept(expressionVisitor);
+            if (buffer.length() == len) {
+                buffer.delete(len - " WHERE ".length(), len);
+            }
         }
     }
 
--- a/src/main/java/net/sf/jsqlparser/util/deparser/DeleteDeParser.java
+++ b/src/main/java/net/sf/jsqlparser/util/deparser/DeleteDeParser.java
@@ -22,7 +22,7 @@ import static java.util.stream.Collectors.joining;
 
 public class DeleteDeParser extends AbstractDeParser<Delete> {
 
-    private ExpressionVisitor expressionVisitor = new ExpressionVisitorAdapter();
+    protected ExpressionVisitor expressionVisitor = new ExpressionVisitorAdapter();
 
     public DeleteDeParser() {
         super(new StringBuilder());
@@ -90,10 +90,7 @@ public class DeleteDeParser extends AbstractDeParser<Delete> {
             }
         }
 
-        if (delete.getWhere() != null) {
-            buffer.append(" WHERE ");
-            delete.getWhere().accept(expressionVisitor);
-        }
+        acceptWhere(delete);
 
         if (delete.getOrderByElements() != null) {
             new OrderByDeParser(expressionVisitor, buffer).deParse(delete.getOrderByElements());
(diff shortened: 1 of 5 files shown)


src/main/java/net/sf/jsqlparser/util/deparser/DeleteDeParser.java
```diff
@@ -22,7 +22,7 @@
 
 public class DeleteDeParser extends AbstractDeParser<Delete> {
 
-    private ExpressionVisitor expressionVisitor = new ExpressionVisitorAdapter();
+    protected ExpressionVisitor expressionVisitor = new ExpressionVisitorAdapter();
```
manticore-projects, line 25: why do we change that to `protected` when there are public getters and setters?

src/main/java/net/sf/jsqlparser/util/deparser/DeleteDeParser.java
```diff
@@ -107,6 +104,13 @@ public void deParse(Delete delete) {
         }
 
     }
+    
+    protected void acceptWhere(Delete delete) {
```
manticore-projects, line 108: ok, but please rename into something like `deparsWhereClause` to stick with the established naming convention.
`accept` implies the visitor pattern which is not applicable here.

src/main/java/net/sf/jsqlparser/util/deparser/MergeDeParser.java
```diff
@@ -16,9 +16,9 @@
 import java.util.List;
 
 public class MergeDeParser extends AbstractDeParser<Merge> implements MergeOperationVisitor {
-    private final ExpressionDeParser expressionDeParser;
+    protected final ExpressionDeParser expressionDeParser;
```
manticore-projects, line 19: should be kept `private`

src/main/java/net/sf/jsqlparser/util/deparser/MergeDeParser.java
```diff
@@ -16,9 +16,9 @@
...
 
-    private final SelectDeParser selectDeParser;
+    protected final SelectDeParser selectDeParser;
```
manticore-projects, line 21: should be kept `private`

src/main/java/net/sf/jsqlparser/util/deparser/SelectDeParser.java
```diff
@@ -334,6 +331,13 @@ public void visit(PlainSelect plainSelect) {
 
     }
 
+    protected void acceptWhere(PlainSelect plainSelect) {
```
manticore-projects, line 334: should probably be moved into the abstract class instead of re-implementing it several times.

manticore-projects (changes requested):
Thank you for your contribution. Please consider the recommended changes and then I would merge it.

redkale:
@manticore-projects  thanks, updated
My requirement scenario is ：
 sql template  "SELECT * FROM user WHERE name = :param_name AND status = :param_status"
where param_name  and param_status both is null,  need generated sql: "SELECT * FROM user"
where param_name is not null, param_status  is null,  need generated sql: "SELECT * FROM user WHERE name = ? "
so deParse where expression maybe is empty

manticore-projects:
> sql template "SELECT * FROM user WHERE name = :param_name AND status = :param_status"

Greetings.

In my opinion, your approach is flawed. Instead of tampering with the `WHERE` clause, you will need to filter for the `JDBCNamedParameter`, then check it's parent AST Node and when it's an `EqualTo` rewrite it into an `IsNull`. 
Alternatively you can rewrite any `EqualTo` Expression into an `IsNull` Expression when the Left or Right Expression is a Null Value.

Those are specific solutions though that won't fit well into the generic de-parser.

src/main/java/net/sf/jsqlparser/util/deparser/DeleteDeParser.java
```diff
@@ -92,7 +92,11 @@ public void deParse(Delete delete) {
 
         if (delete.getWhere() != null) {
             buffer.append(" WHERE ");
-            delete.getWhere().accept(expressionVisitor);
+            int len = buffer.length();
```
manticore-projects, line 95: Please remove those specific parts from the PR as it does not fit into the generic Deparser.

src/main/java/net/sf/jsqlparser/util/deparser/SelectDeParser.java
```diff
@@ -251,7 +225,11 @@ public void visit(PlainSelect plainSelect) {
 
         if (plainSelect.getWhere() != null) {
             buffer.append(" WHERE ");
+            int len = buffer.length();
```
manticore-projects, line 228: Please remove those specific parts from the PR as it does not fit into the generic Deparser.

src/main/java/net/sf/jsqlparser/util/deparser/UpdateDeParser.java
```diff
@@ -90,7 +90,11 @@ public void deParse(Update update) {
 
         if (update.getWhere() != null) {
             buffer.append(" WHERE ");
-            update.getWhere().accept(expressionVisitor);
+            int len = buffer.length();
```
manticore-projects, line 93: Please remove those specific parts from the PR as it does not fit into the generic Deparser.

manticore-projects (changes requested):
I won't agree with the StringBuffer based `WHERE` clause exception. Rest is ok though and I would merge it.

manticore-projects:
Please fix QA exceptions and then I would merge.

redkale:
@manticore-projects    thanks,  I add  deparseWhereClause like before。
Statement will be caching, it cannot modify.  generate a new sql template based on parameters， like this:
`class SqlTemple {

    private final Statement sqlStatement;

    public SqlTemple() throws ParseException {
        String sql = "SELECT * FROM user WHERE name = :param_name AND status = :param_status";
        CCJSqlParser sqlParser = new CCJSqlParser(sql).withAllowComplexParsing(true);
        this.sqlStatement = sqlParser.SingleStatement();
    }
    
    public String createJdbcSql(Map<String, Object> map) {
        StringBuilder buffer = new StringBuilder();
        NativeExprDeParser myParser = new NativeExprDeParser(buffer, map);
        SelectDeParser parser = new SelectDeParser();
        StatementDeParser deParser = new StatementDeParser(myParser, parser, buffer);
        sqlStatement.accept(deParser);
        //map is empty  return "SELECT * FROM user"
        //map is {"param_name":"aa"}  reutrn "SELECT * FROM user WHERE name = :param_name"
        //map is {"param_status":2}  reutrn "SELECT * FROM user WHERE status = :param_status"
        //map is {"param_name":"aa","param_status":2}  reutrn "SELECT * FROM user WHERE name = :param_name AND status = :param_status"
        return buffer.toString();
    }

}`

redkale:
My framework source:  https://github.com/redkale/redkale-plugins/tree/main/src/main/java/org/redkalex/source/parser
@manticore-projects 
The function I designed is to define a SQL with full filtering conditions, and then dynamically build a new SQL with where conditions based on the existence of parameters.
Constructing the sql string into a Statement consumes performance and requires caching the Statement. In StatementDeParser, only the buffer can be dynamically controlled, so I want to dynamically adjust the buffer, not the AST Node.
```java
public interface ForumInfoMapper extends BaseMapper<ForumInfo> {

    @Sql("SELECT f.forum_groupid, s.forum_section_color "
        + "FROM forum_info f, forum_section s "
        + " WHERE f.forumid = s.forumid AND "
        + "s.forum_sectionid = #{bean.forumSectionid} AND "
        + "f.forumid = #{bean.forumid} AND s.forum_section_color = #{bean.forumSectionColor}")
    public ForumResult findForumResult(ForumBean bean);

    @Sql("SELECT f.forum_groupid, s.forum_section_color "
        + "FROM forum_info f, forum_section s "
        + " WHERE f.forumid = s.forumid AND "
        + "s.forum_sectionid = #{bean.forumSectionid} AND "
        + "f.forumid = #{bean.forumid} AND s.forum_section_color = #{bean.forumSectionColor}")
    public CompletableFuture<ForumResult> findForumResultAsync(ForumBean bean);
}
```

manticore-projects:
QA still failing.
Run './gradlew :spotlessApply' to fix these violations.

manticore-projects:
Before you push, please run the QA locally via `gradle check`.

manticore-projects:
Although I still believe that your approach is not best practise. Rather than tampering with the `WHERE` clause I would rewrite the query like:

```sql
select *
from t
where ( :param1 is null or column1= :param1) AND ( :param2 IS NULL or column2 = :param2 )
```

Then you can set the parameters `:param1` or `:param2` to NULL when not provided in your Parameter Map.
You may want to have a look at https://manticore-projects.com/MJdbcUtils/index.html which I wrote for exactly this purpose.

PR #2013, opened 2024-05-27
support custom DeParser

redkale:
My requirement scenario is ：
sql template "SELECT * FROM user WHERE name = :param_name AND status = :param_status"
where param_name and param_status both is null, need generated sql: "SELECT * FROM user"
where param_name is not null, param_status is null, need generated sql: "SELECT * FROM user WHERE name = ? "
so deParse where expression maybe is empty

deparseDistinctClause、deparseSelectItemsClause、deparseOrderByElementsClause requires two parameters ,
Parameter plainSelect mainly used to determine the current selection. like：
```java
  @Override
  protected void deparseDistinctClause(PlainSelect plainSelect, Distinct distinct) {
    if (this.countSelect != plainSelect) {
        super.deparseDistinctClause(plainSelect, distinct);
    } else {
        super.deparseDistinctClause(plainSelect, this.countDistinct);
    }
}

protected void deparseSelectItemsClause(PlainSelect plainSelect, List<SelectItem<?>> selectItems) {
    if (this.countSelect != plainSelect) {
        super.deparseSelectItemsClause(plainSelect, selectItems);
    } else {
        super.deparseSelectItemsClause(plainSelect, this.countRootSelectItems);
    }
}

protected void deparseOrderByElementsClause(PlainSelect plainSelect, List<OrderByElement> orderByElements) {
    if (this.countSelect != plainSelect) {
        super.deparseOrderByElementsClause(plainSelect, orderByElements);
    } else {
        super.deparseOrderByElementsClause(plainSelect, null);
    }
}
```

### DIFF
--- a/src/main/java/net/sf/jsqlparser/util/deparser/DeleteDeParser.java
+++ b/src/main/java/net/sf/jsqlparser/util/deparser/DeleteDeParser.java
@@ -90,10 +90,7 @@ public class DeleteDeParser extends AbstractDeParser<Delete> {
             }
         }
 
-        if (delete.getWhere() != null) {
-            buffer.append(" WHERE ");
-            delete.getWhere().accept(expressionVisitor);
-        }
+        deparseWhereClause(delete);
 
         if (delete.getOrderByElements() != null) {
             new OrderByDeParser(expressionVisitor, buffer).deParse(delete.getOrderByElements());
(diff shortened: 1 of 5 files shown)


redkale:
@manticore-projects  thanks, I resubmitted a request.

Issue #2040, opened 2024-07-17
[BUG] Compilation error after execution of updateKeywords


ssteinhauser:
The execution of the `updateKeywords` task adds the following tokens to `RelObjectNameWithoutValue`:

https://github.com/JSQLParser/JSqlParser/blob/411a3da9facf20634c619eb6ffb5626ea57ce7d1/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt#L529-L551

This causes the javacc-jjtree goal to fail:

```
Error: Line 1962, Column 313: String token "BISTRING" has been defined as a private regular expression.
Error: Line 1962, Column 1163: String token "ENUM" has been defined as a private regular expression.
Error: Line 1962, Column 1500: String token "HUGEINT" has been defined as a private regular expression.
Error: Line 1962, Column 1815: String token "MAP" has been defined as a private regular expression.
Error: Line 1962, Column 3401: String token "TIMETZ" has been defined as a private regular expression.
Error: Line 1962, Column 3527: String token "UBIGINT" has been defined as a private regular expression.
Error: Line 1962, Column 3542: String token "UHUGEINT" has been defined as a private regular expression.
Error: Line 1962, Column 3558: String token "UINTEGER" has been defined as a private regular expression.
Error: Line 1962, Column 3672: String token "USMALLINT" has been defined as a private regular expression.
Error: Line 1962, Column 3689: String token "UTINYINT" has been defined as a private regular expression.
Error: Line 1962, Column 3705: String token "UUID" has been defined as a private regular expression.
```

manticore-projects:
Thank you for reporting, I actually broke it myself when introducing complex compound Data Type tokens.
I will definitely fix it, but this will take a couple of days since I am in transit right now. Sorry for any inconvenience.

Issue #2041, opened 2024-07-17
[BUG] REGEXP_REPLACE as LikeExpression


ssteinhauser:
I am about to implement the Exasol syntax. For this I'd like to implement REGEXP_LIKE as LikeExpression (see https://docs.exasol.com/db/latest/sql_references/predicates/not_regexp_like.htm). Since it is also allowed to be a function in Oracle (as already implemented), it needs to be added to the whitelisted keywords.
Unfortunately, the following condition prevents it from being added to `RelObjectNameWithoutValue` since REGEXP_LIKE contains an underscore which is not allwed here:
`tokenValue.matches("[A-Za-z]+")`

https://github.com/JSQLParser/JSqlParser/blob/11cebcfd122094402548cc8d5b53a063134284f5/src/main/java/net/sf/jsqlparser/parser/ParserKeywordsUtils.java#L235

What was the reason for adding this condition and how can or should I proceed to support REGEXP_LIKE as LikeExpression?

manticore-projects:
Greetings!

Thank you for your interest and contribution.
In genera this syntax is supported already and the following works:

```sql
SELECT 'My mail address is my_mail@exasol.com'
       RLIKE '(?i).*[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,4}.*'
       AS contains_email;
```

Now we only need to register REGEXP_LIKE as an Like Operator (like we did for `RLIKE` yet).
So my advise is: follow the `RLIKE` implementation and also take reference to the `EXTRACT` production (which also can be a normal function as well as a special function). Something like this will work for the `REGEXP_LIKE` operator too.

I will be able to help you on this by next week only since I am somewhere between Jakarta and Lagos right now. Cheers.

ssteinhauser:
This was also my first intension to implement it similar to RLIKE, so I've added a new token `<K_REGEXP_LIKE: "REGEXP_LIKE">` and added it as an option to `LikeExpression` as well as to the Java implementation.

But the problem is, that the `updateKeywords` task won't whitelist the `REGEXP_LIKE` token (unlike `EXTRACT`), because it contains an Underscore. The condition `tokenValue.matches("[A-Za-z]+")` in the following line prevents it from being whitelisted for function names:
https://github.com/JSQLParser/JSqlParser/blob/11cebcfd122094402548cc8d5b53a063134284f5/src/main/java/net/sf/jsqlparser/parser/ParserKeywordsUtils.java#L235

[...]

PR #2044, opened 2024-07-19
Avoid private tokens to be white listed and allow any word character in token value

ssteinhauser:
The idea of this PR is to parse and iterate over each token definition and don't whitelist anything which is defined within a private token.
Furthermore, this PR allows any token containing word character as token value to be whitelisted.

Closes JSQLParser/JSqlParser#2040, JSQLParser/JSqlParser#2041


manticore-projects:
Thank you for your interest, work and effort! I really appreciate you solving this especially when I did not find the time yet to look into it.

manticore-projects:
Please help to fix the code formatting exceptions, Run './gradlew :spotlessApply' to fix these violations.
Thank you!

PR #2083, opened 2024-10-02
Failure to parse query with PRIOR in select list

hannes92:
Encountered a failure with parsing queries having PRIOR in the select list (select t.id, prior t.id as parent_id ...).
Thought I would give it a try and fix it myself, but this is my first encounter with AST:s and parser libs in general so do keep that in mind.


manticore-projects:
Excellent work, very clean!
Thank you very much for your interest and contribution and we are looking for seeing more from you soon. Cheers!

Issue #2016, opened 2024-06-07
[BUG] JSQLParser Version : 4.0 : ON condition not parsed correctly when modifying SQL with natural join


Createsequence:
**Description**

Hi, I am trying to dynamically add condition statements to SQL using JSqlparser. However, it seems there is an issue when the SQL contains a natural join:

~~~java
public static void main(String[] args) throws JSQLParserException {
    String sql = "SELECT * FROM A a, B b where a.id = b.id";
    Statement statement = CCJSqlParserUtil.parse(sql, null);
    PlainSelect plainSelect = (PlainSelect) ((Select) statement).getSelectBody();
    Join join = CollUtil.getFirst(plainSelect.getJoins());
    // append a condition "a.id = '100'"
    join.setOnExpression(new EqualsTo(new Column("a.id"), new StringValue("100")));

    // = "SELECT * FROM A a, B b WHERE a.id = b.id"
    System.out.println(statement.toString());
}
~~~

I am not sure how JSqlparser handles this situation, but `SELECT * FROM A a, B b WHERE a.id = b.id` is clearly not the expected result.

**Analysis**

I tried debugging to find the root cause of this issue. I noticed that in the above code, the query of table B is parsed into a `net.sf.jsqlparser.statement.select.Join` object, and at this point, `Join.simple` is `true`.

After setting the ON condition with `join.setOnExpression(new EqualsTo(new Column("a.id"), new StringValue("100")));` and attempting to convert it back to SQL, the ON condition is not correctly parsed into SQL because `Join.simple` is `true`:

~~~java
@Override
public String toString() {
    if (isSimple() && isOuter()) {
        return "OUTER " + rightItem;
    } else if (isSimple()) {
        return "" + rightItem;
    } else {

        // isSimple() returns true, so this branch is ignored

        String type = "";

        if (isRight()) {
            type += "RIGHT ";
        } else if (isNatural()) {
            type += "NATURAL ";
        } else if (isFull()) {
            type += "FULL ";
        } else if (isLeft()) {
            type += "LEFT ";
        } else if (isCross()) {
            type += "CROSS ";
        }

        if (isOuter()) {
            type += "OUTER ";
        } else if (isInner()) {
            type += "INNER ";
        } else if (isSemi()) {
            type += "SEMI ";
        }

        if (isStraight()) {
            type = "STRAIGHT_JOIN ";
        } else if (isApply()) {
            type += "APPLY ";
        } else {
            type += "JOIN ";
[...]

PR #2108, opened 2024-11-13
Fix issue 2106: Add parsing functionality for MySQL ADD PARTITION and DROP PARTITION clauses in ALTER TABLE statements(2)

minleejae:
Summary
This PR addresses the issue #2016 with JSQLParser's inability to parse ADD PARTITION clauses in ALTER TABLE statements for MySQL 8.0. Previously, these clauses were not recognized, causing parsing errors for certain ALTER TABLE operations.

Changes
- Modified the parser to correctly identify and handle the ADD PARTITION clauses, including `storage options`, allowing for accurate parsing of these statements.
- Added support for parsing `DROP PARTITION` clauses.
- Refactored the `TRUNCATE PARTITION` parsing functionality.

References
MySQL Documentation: [ALTER TABLE Partition Operations](https://dev.mysql.com/doc/refman/8.4/en/alter-table-partition-operations.html)


minleejae:
Hello,
I'm encountering a StackOverflowError during the compileJava task in the CI environment, as shown in the error log below. However, I haven't been able to reproduce this issue locally, I'd greatly appreciate your guidance.
Notably, this error did not occur before the fix Codacy commit.

manticore-projects:
Thank you for your contribution.
Unfortunately I have no idea yet what causes this Stack Overflow:
1) I can't re-produce it locally
2) It worked fine on GitHub one time, only to fail afterwards many times -- without any changes on the Java code

minleejae:
@manticore-projects 
Thank you for checking into this. Could the StackOverflowError in the Gradle CI environment be due to an insufficient stack size? I’m not sure, but it might be the issue.

manticore-projects:
Right now, my uneducated guess was about a build-caching issue because the builds succeed first time and then fail sub-sequentially. We will figure this out eventually.

manticore-projects:
@minleejae: If you ever come to Thailand for a course of golf, please do let me know!

minleejae:
@manticore-projects 
Thank you for the warm welcome!

Regarding the Gradle CI issue, it's possible that the errors we’re seeing might be related to build-caching inconsistencies, which makes it an interesting topic to look into further. 

Additionally, I’m not very familiar with JavaCC syntax, I’m curious—does using `LOOKAHEAD` in JavaCC have any notable impact on performance compared to not using it?

manticore-projects:
> Additionally, I’m not very familiar with JavaCC syntax, I’m curious—does using `LOOKAHEAD` in JavaCC have any notable impact on performance compared to not using it?

Yes, it does come with a performance penalty although you need to distinguish:

1) a numeric LOOKAHEAD like `LOOKAHEAD(2)` is harmless because it will check the next 2 tokens only. In fact I have seen grammars that apply a default LOOKAHEAD(3) to just everything (JSQLParser's default is LOOKAHEAD(1))

2) a semantic LOOKAHEAD is really expensive and must be avoided at all cost -- although we use it especially for `FUNCTION`. Unfortunately SQL is a terrible dialect by design with no clear blocks and to make things worse, JSQLParser has grown over 20+ years and also tries to be dialect agnostic. So we have no choice but to engage some functional lookahead -- even when you can actually break the parser with it. For this reason we have `COMPLEX` vs. `SIMPLE` parsing modes. All not very beautiful from a technical point of view, but in practice it just works most of the time.

Issue #2109, opened 2024-11-13
[BUG] JSQLParser 5.0: Generic: Update with true as value is parsed as column


lucarota:
### Failing SQL Feature:
When updating a Boolean column, the value true is parsed as column.

### SQL Example:
```sql
UPDATE TestTable SET Column1 = 'Tester', Column2 = true WHERE Id = 'AAS'    
```
Column1 and Column2 are parsed as a Column object.
Tester is parsed as a StringValue
**true is parsed as Column object. Why?** I expect it to be parsed as BooleanValue or something else value object.

```java
Statement queryData = CCJSqlParserUtil.parse("UPDATE TestTable SET Column1 = 'Tester', Column2 = true WHERE Id = 'AAS'");
````
![image](https://github.com/user-attachments/assets/6297ab59-5af6-4e01-99ec-35ffb2dcfe05)

### Software Information:
- JSqlParser version 5.0 (same problem with the snapshot)

manticore-projects:
Solved via #2110

PR #2110, opened 2024-11-14
fix: Issue #2109 true and false value parsed as column instead of BooleanValue

lucarota:
Hi all,

this patch addresses Issue #2109 in the JSQLParser library, where TRUE and FALSE were incorrectly handled as column names rather than as boolean values.

According to the SQL standard ISO/IEC 9075:2016, TRUE and FALSE are reserved keywords representing boolean literals, not column identifiers.

This patch updates the parser to recognize TRUE and FALSE as boolean literals, ensuring compliance with the SQL standard and preventing potential errors in query interpretation.

Ciao,
Luca


manticore-projects:
Very clean, well done! Thank you for your help and contribution!

Issue #2123, opened 2024-12-12
[FEATURE] TablesNamesFinder does not support CREATE VIEW


cptnricard:
Is there a reason why `TablesNamesFinder` supports a `CREATE TABLE` but throws an exception for a `CREATE VIEW`?
Wouldn't the implementation be almost identical?

The current implementation is 
```java
    @Override
    public <S> Void visit(CreateView createView, S context) {
        throwUnsupported(createView);
        return null;
    }
```

It seems that it could easily be replaced with 
```java
    @Override
    public <S> Void visit(CreateView create, S context) {
        visit(create.getView(), null);
        if (create.getSelect() != null) {
            create.getSelect().accept((SelectVisitor<?>) this, context);
        }
        return null;
    }
```

Which is almost a copy&paste of the CREATE TABLE implementation
```java
    @Override
    public <S> Void visit(CreateTable create, S context) {
        visit(create.getTable(), null);
        if (create.getSelect() != null) {
            create.getSelect().accept((SelectVisitor<?>) this, context);
        }
        return null;
    }
```

manticore-projects:
Greetings!
You are right of course and your PR was very welcome.

manticore-projects:
fixed via #2124

PR #2124, opened 2024-12-15
fix: [FEATURE] TablesNamesFinder does not support CREATE VIEW #2123

cptnricard:
TablesNamesFinder did not have an implementation for `CREATE VIEW AS SELECT ...`


manticore-projects:
great job, thank you for your contribution and interest! cheers!

PR #2128, opened 2024-12-22
add support for MATERIALIZED in WITH clause

tomershay:
This PR adds support for MATERIALIZED CTEs: `WITH table AS MATERIALIZED ...`

Reference: https://www.postgresql.org/docs/current/queries-with.html#QUERIES-WITH-CTE-MATERIALIZATION


manticore-projects:
Thank you!

PR #2133, opened 2024-12-28
Fix null exception in ExpressionVisitorAdapter with simple INTERVAL expression

tomershay:
`ExpressionVisitorAdapter` was throwing a `NullPointerException` when a simple `INTERVAL` expression was provided, which didn't contain an actual expression (`.getExpression()` returns `null`).

Sample expression that causes the issue: `INTERVAL 1 DAY`

This PR adds validation around that code to fix this issue.


manticore-projects:
Thank you @tomershay for fixing this, much appreciated!

PR #2141, opened 2025-01-09
feat: Support unknown keyword

EmilyOng:
This PR adds support for the `UNKNOWN` keyword and `IsUnknownExpression` in order to parse [`IS UNKNOWN`](https://dev.mysql.com/doc/refman/8.4/en/comparison-operators.html#operator_is) and [`IS NOT UNKNOWN`](https://dev.mysql.com/doc/refman/8.4/en/comparison-operators.html#operator_is-not) operations in MySQL.


manticore-projects:
Thank you for your contribution, very nice and clean!
Hope to see more from you soonest, cheers!

PR #2142, opened 2025-01-12
feat: Adding coverage for Postgres "OVERRIDING SYSTEM VALUE"

nicky6s:
As part of Postgres syntax you can use an "OVERRIDING SYSTEM VALUE" clause as part of an INSERT.
Doc -> https://www.postgresql.org/docs/current/sql-insert.html

```
OVERRIDING SYSTEM VALUE
If this clause is specified, then any values supplied for identity columns will override the default sequence-generated values.
```

The parser doesn't currently parse INSERTs that include this clause, this PR adds support.

Note that I have not added OVERRIDING as a restricted keyword, so you can still have schemas, tables, columns named "overriding".


manticore-projects:
Clean! Thank you very much for your help and contribution. Cheers!

PR #2154, opened 2025-02-02
Add more tests to TablesNamesFinderTest for SubqueryAliasesIssue1987

pinkfloyda:
When I test the FindTableNames, I created more complex SQL queries to ensure it works and able to extract correct set of tables out. So contribute such tests to avoid regressions introduced in future


manticore-projects:
Thank you!

PR #2180, opened 2025-03-04
IsNullExpression and IsBooleanExpression children are not visited in …

SuperPat45:
…TablesNamesFinder

IsNullExpression and IsBooleanExpression children are not visited in TablesNamesFinder


manticore-projects:
Thank you for your help and contribution!

PR #2197, opened 2025-03-19
Add support for AllTableColumns in PrimaryExpression

tomershay:
PostgreSQL supports expressions such as `posts.*` (all columns from the `posts` table) in many places, such as the `GROUP BY` clause.

For example, in this query:
```sql
select c.post_id, p.* from posts p inner join comments c on c.post_id = p.post_id group by p.post_id, c.post_id, p.*;
```

This PR adds support for AllTableColumns in PrimaryExpression.

I wasn't sure if PrimaryExpression is the best fit for this, so will be happy for suggestions on relocating it if there's a better expression group for it.

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -3252,11 +3252,7 @@ SelectItem<?> SelectItem() #SelectItem:
     // @fixme: Oracle's SEQUENCE.nextval is parsed as COLUMN with a name part nextval
     // @todo: parse a proper SEQUENCE instead of a COLUMN
     (
-        expression = AllColumns()
-        |
         LOOKAHEAD( 3 ) expression = ConnectByPriorOperator()
-        |
-        LOOKAHEAD(AllTableColumns()) expression = AllTableColumns()
         |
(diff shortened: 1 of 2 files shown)


manticore-projects:
@tomershay 

Thank you for this contribution. My understanding is: A `PrimaryExpression` can be used anywhere and everywhere -- which is not the case for `AllColumns` or `AllTableColumns` and especially seems to depend on the RDBMS also. However, from a pragmatical view it comes close and indeed seem to make sense.

However, we segregate both Expression at other places already and I believe even have to use semantic lookaheads (which are expensive). So in my opinion, we either

a) keep segregating `AllColumns` or `AllTableColumns` wherever it is allowed (cleaner and faster) or
b) remove all those segregations in all productions and incorporate `AllColumns` or `AllTableColumns` into `PrimaryExpression` (slower, but more pragmatic)

What I will not want is a mix of both approaches because it can make the Grammar even harder to maintain in the future.

tomershay:
@manticore-projects, thanks, makes total sense.

> remove all those segregations in all productions and incorporate AllColumns or AllTableColumns into PrimaryExpression (slower, but more pragmatic)

I went with the approach above, so I removed the segregation of `AllColumns` or `AllTableColumns` in other places, and added them just to `PrimaryExpression`.

Looking forward to hearing your thoughts.

PR #2203, opened 2025-03-24
Allow specific keywords to be parsed as identifiers (fix issue with RENAME COLUMN)

minleejae:
## Summary
This PR introduces a new parser rule `KeywordOrIdentifier()` to handle cases where certain keywords are used as identifiers. This addresses issues that arose after specific tokens like `NAME` were introduced as keywords.

## Problem
While parsing a statement like:
```sql
ALTER TABLE test_table RENAME COLUMN name TO full_name;
```

the parser failed because `name` was recognized strictly as a `<K_NAME>` keyword after [this commit](https://github.com/JSQLParser/JSqlParser/commit/f932b8f9807530cebbcb280308fb05fe0dd1d983), instead of being allowed as an identifier.

## Solution
- Introduced `KeywordOrIdentifier()` rule that accepts:
  - `<S_IDENTIFIER>`
  - `<S_QUOTED_IDENTIFIER>`
  - Selected keywords like `<K_NAME>`, `<K_NEXT>`, `<K_VALUE>`, `<K_PUBLIC>`, `<K_STRING>`
  - Additional tokens can be added as needed in the future.
- Replaced relevant parts of the grammar to use `KeywordOrIdentifier()`.

---
I’m open to any feedback or suggestions on this approach!

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -719,6 +719,35 @@ TOKEN:
     }
 }
 
+/**
+ * Parses identifiers including standard SQL identifiers, quoted identifiers, and specific keywords.
+ *
+ * This is used in cases where certain SQL keywords (like NAME, NEXT, VALUE, etc.) can appear
+ * as identifiers (e.g., table names, column names) depending on the SQL dialect or context.
+ *
+ * Supported tokens:
+ *  - <S_IDENTIFIER>: Standard unquoted SQL identifier
+ *  - <S_QUOTED_IDENTIFIER>: Quoted identifier (e.g., `identifier` or "identifier")
+ *  - <K_NAME>, <K_NEXT>, <K_VALUE>, <K_PUBLIC>, <K_STRING>: Specific keywords treated as identifiers
+ *
+ * @return Token representing the identifier or keyword used as identifier
+ */
+Token KeywordOrIdentifier():
+{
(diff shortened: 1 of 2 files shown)


manticore-projects:
We need to rework the whole keyword management eventually. But in the meantime this can be a remedy.

minleejae:
Hi, @manticore-projects 
I also agree that the current keyword handling could use a more systematic approach.
Out of curiosity, could you share your thoughts on what the ultimate direction for keyword management should look like? I'd like to align any future refactoring efforts accordingly.

minleejae:
Hi, @manticore-projects 
This sounds like quite a large-scale refactoring effort. I’m curious — do you have any rough timeline or plan in mind for when you'd like to tackle this?

Also, just wondering, have you ever considered adopting something like ANTLR instead, or is the plan to stick with JavaCC/CongoCC going forward?

manticore-projects:
`ANTLR` has been found producing slower code and so is out of scope for me. Although this claim could need a verification again.

minleejae:
@manticore-projects 
It's interesting that ANTLR was found to be slow. I wonder if that has something to do with the structural differences between LL and LR parsers. Of course, replacing ANTLR would come with the high cost of building everything from scratch, but I thought one of its strengths was the ability to support multiple languages.

manticore-projects:
I am not doubting the strength or superiority of ANTLR in general but only point out why I was more interested in sticking with the JavaCC origin as long as possible. JavaCC/CongoCC also support multiple languages although I am really interested in Java only (and maybe C).

PR #2207, opened 2025-03-26
feat: add support SELECT all columns from function result

0xlianhu:
PostgreSQL supports composite type expansion using `.*`, for example,  below query is valid in PostgreSQL:
```SQL
SELECT (pg_stat_file('postgresql.conf')).*
```
More generic one is like below one in queries or sub-queries:
```
SELECT (function_name(param1, ... paramN)).* FROM my_table ...
```
There might be multiple duplicated parentheses surround the function call, such as:
```
SELECT (((pg_stat_file('postgresql.conf')))).* -- this is same effect as one `()` surrounding
```

The PR is adding the support to parse/deparse this type of SQLs.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitor.java
+++ b/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitor.java
@@ -60,6 +60,7 @@ import net.sf.jsqlparser.schema.Column;
 import net.sf.jsqlparser.statement.piped.FromQuery;
 import net.sf.jsqlparser.statement.select.AllColumns;
 import net.sf.jsqlparser.statement.select.AllTableColumns;
+import net.sf.jsqlparser.statement.select.FunctionAllColumns;
 import net.sf.jsqlparser.statement.select.ParenthesedSelect;
 import net.sf.jsqlparser.statement.select.Select;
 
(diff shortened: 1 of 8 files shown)


manticore-projects:
Beautifully executed! Well done and thank you much for your effort and contribution.

manticore-projects:
Greetings!

Unfortunately, I found out that commit 30cf5d7b causes a huge performance deterioration:

```
(version)  Mode  Cnt    Score    Error  Units
   latest  avgt    3   81.637 ± 21.386  ms/op <-- `FunctionAllColumns()` commented out
      5.2  avgt    3  393.523 ± 54.199  ms/op <-- release w/ `FunctionAllColumns()`
      5.1  avgt    3   86.419 ± 22.178  ms/op <-- release w/o `FunctionAllColumns()`
```

This toll is not acceptable for a rather exotic feature only and I am going to disable it.
Although all code is still there and only the call of  `FunctionAllColumns()` in `PrimaryExpression` has been commented out.

You are welcome to refactor your code and I will most happily accept it again once performance can be proved.

Recommendations:

1) try to avoid the syntactic Lookahead 
2) your `(((( FUNCTION )))).*` is nothing else but a `ParenthesedExpressionList` with just one parameter of type `Function`, followed by ".*". So maybe parse for a possible extension by ".*" at the end of a `ParenthesedExpressionList` when it has only one single `Function` element.

I am open for discussion of course.
Best and cheers!

Issue #2076, opened 2024-09-14
[BUG] JSQLParser 5.0 : MySQL : ALTER INDEX and ADD INDEX ... INVISIBLE


minleejae:
## Issue Description:
There is a parsing failure in JSQLParser 5.0 when dealing with MySQL's ALTER INDEX and ADD INDEX ... INVISIBLE statements. These statements, introduced in [MySQL 8.0](https://dev.mysql.com/doc/refman/8.4/en/invisible-indexes.html)

```
ALTER TABLE t1 ADD INDEX k_idx (k) INVISIBLE;
ALTER TABLE t1 ALTER INDEX i_idx INVISIBLE;
ALTER TABLE t1 ALTER INDEX i_idx VISIBLE;
```

## Error
```
Parsing failed for DDL: ALTER TABLE t1 ADD INDEX k_idx (k) INVISIBLE; ==> Unexpected exception thrown: net.sf.jsqlparser.JSQLParserException: net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: "INVISIBLE" <S_IDENTIFIER>
    at line 1, column 36.

Was expecting one of:

    ","
    "DEFERRABLE"
    "DISABLE"
    "ENABLE"
    "NOT"
    "NOVALIDATE"
    "USING"
    "VALIDATE"
    <EOF>
    <ST_SEMICOLON>

org.opentest4j.AssertionFailedError: Parsing failed for DDL: ALTER TABLE t1 ADD INDEX k_idx (k) INVISIBLE; ==> Unexpected exception thrown: net.sf.jsqlparser.JSQLParserException: net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: "INVISIBLE" <S_IDENTIFIER>
    at line 1, column 36.

Was expecting one of:

    ","
    "DEFERRABLE"
    "DISABLE"
    "ENABLE"
    "NOT"
    "NOVALIDATE"
    "USING"
    "VALIDATE"
    <EOF>
    <ST_SEMICOLON>

	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:152)
	at org.junit.jupiter.api.AssertDoesNotThrow.createAssertionFailedError(AssertDoesNotThrow.java:84)
	at org.junit.jupiter.api.AssertDoesNotThrow.assertDoesNotThrow(AssertDoesNotThrow.java:53)
[...]

PR #2234, opened 2025-04-30
feat: enhance ALTER TABLE support with UNIQUE KEY and INVISIBLE index

minleejae:
# Description
This PR adds support for two MySQL DDL statements:

`ALTER TABLE ADD CONSTRAINT UNIQUE KEY syntax`
`ALTER TABLE ALTER INDEX INVISIBLE syntax`

# Problem Solved
Previously, JSqlParser couldn't properly parse the following SQL statements:
```sql
ALTER TABLE sbtest1 ADD CONSTRAINT UNIQUE KEY ux_c3 (c3);
ALTER TABLE sbtest1 ALTER INDEX c4 INVISIBLE;
```

These are valid MySQL syntax for adding a unique constraint with a custom name and for making an index invisible to the query optimizer while keeping it maintained.

# Related Issue
#2076

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java
+++ b/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java
@@ -9,7 +9,17 @@
  */
 package net.sf.jsqlparser.statement.alter;
 
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
+
 import net.sf.jsqlparser.expression.Expression;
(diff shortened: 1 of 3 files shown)


minleejae:
Hi, @manticore-projects 

Could you please let me know when the JSqlParser 5.2 release is scheduled? 
I'd like to reference this for planning purposes in my project.

Thanks!

manticore-projects:
> Could you please let me know when the JSqlParser 5.2 release is scheduled?

Unfortunately this is not on me, but I will ask @wumpz  immediately and with some luck, he will get this done in May.
Would this work for you?

minleejae:
@manticore-projects 
Thanks for the reply!

I'm currently working on a project that depends on JSqlParser, and I'm considering switching to the official release version rather than using a custom build.
There's no urgency at all — I just wanted to get a rough idea of the possible timeline.
Appreciate your help, thank you!

manticore-projects:
Thank you for your work and effort! Cheers!

PR #2238, opened 2025-05-12
feat: support distinctrow keyword

EmilyOng:
This PR adds support for the `DISTINCTROW` keyword available in [MySQL](https://dev.mysql.com/doc/refman/8.4/en/select.html) and [MariaDB](https://mariadb.com/kb/en/select/#distinct).

> ```sql
> SELECT
>     [ALL | DISTINCT | DISTINCTROW ]
>     ....
>     select_expr [, select_expr]
> ...
> ```

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/select/Distinct.java
+++ b/src/main/java/net/sf/jsqlparser/statement/select/Distinct.java
@@ -20,6 +20,7 @@ public class Distinct implements Serializable {
 
     private List<SelectItem<?>> onSelectItems;
     private boolean useUnique = false;
+    private boolean useDistinctRow = false;
 
     public Distinct() {}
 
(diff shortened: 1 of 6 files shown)


src/main/java/net/sf/jsqlparser/statement/select/Distinct.java
```diff
@@ -43,9 +44,18 @@ public void setUseUnique(boolean useUnique) {
         this.useUnique = useUnique;
     }
 
+    public boolean isUseDistinctRow() {
```
EmilyOng, line 47: Since [`DISTINCTROW` is a synonym for `DISTINCT`](https://dev.mysql.com/doc/refman/8.4/en/select.html), I re-used the existing `Distinct` class and only augmented it with the `useDistinctRow` attribute to determine the string representation of `Distinct`.

src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
```diff
@@ -3027,6 +3028,8 @@ PlainSelect PlainSelect() #PlainSelect:
             [ LOOKAHEAD(2) "ON" "(" distinctOn=SelectItemsList() { plainSelect.getDistinct().setOnSelectItems(distinctOn); } ")" ]
         )
         |
+        <K_DISTINCTROW> { Distinct distinct = new Distinct(); distinct.setUseDistinctRow(true); plainSelect.setDistinct(distinct); }
```
EmilyOng, line 3031: This is separate from the `<K_DISTINCT>` case right above, because `DISTINCT ON` or `DISTINCTROW ON` is not applicable in MySQL.

manticore-projects:
You will need to resolve conflicts please before I can merge.

manticore-projects:

PR #2256, opened 2025-06-05
Exasol: IMPORT/EXPORT

ssteinhauser:
This PR adds support for Exasols IMPORT and EXPORT statements:
https://docs.exasol.com/db/latest/sql/import.htm
https://docs.exasol.com/db/latest/sql/export.htm

The JMH benchmark looks good on my machine:
```
Benchmark                               (version)  Mode  Cnt    Score    Error  Units
JSQLParserBenchmark.parseSQLStatements     latest  avgt   15  119.730 ± 35.438  ms/op
JSQLParserBenchmark.parseSQLStatements        5.3  avgt   15  244.662 ± 99.050  ms/op
JSQLParserBenchmark.parseSQLStatements        5.1  avgt   15  185.566 ± 76.161  ms/op
```


ssteinhauser:
> @ssteinhauser:
> 
> Thank you for your contribution, I will look into it over the weekend. Though I have a question regarding your JMH: can you please share the specs of your machine?
> 
> a) your tests look super slow in general (I am recording only 80 ms on a cheap laptop) b) your tests record a huge improvement over 5.3 (which I could not explain) c) your tests have very high variances
> 
> Not a real concern or show stopper, I am just curious!

Yes, the benchmark results are super high. Probably because I had quite a few other tools and tasks running consuming lots of CPU and memory. However, my specifications are 32GB of RAM and an Intel Core i7-1185G7 (3.00GHz) CPU.

I've just reexecuted the benchmark (inside WSL) while having lower load on the machine, but the results are still a bit high although the variances are lower:
```
Benchmark                               (version)  Mode  Cnt    Score    Error  Units
JSQLParserBenchmark.parseSQLStatements     latest  avgt   15  140.343 ± 26.198  ms/op
JSQLParserBenchmark.parseSQLStatements        5.3  avgt   15  103.779 ± 20.458  ms/op
JSQLParserBenchmark.parseSQLStatements        5.1  avgt   15  119.490 ± 27.174  ms/op
```

manticore-projects:
Good Morning Stefan,

the decision is all yours because you understand your implementation best.
My take is:

- you added lots of Exasol specific Grammar (kudos!)
- this Grammar slows down the Parser somewhat (expected and justified)
- this Grammar does not serve most of the users (unless they want Exasol)

Best way would be to have RDBMS specific parsers and I will look for ways to achieve that soon.
In the meantime, I wonder if we can introduce Features like 'DIALECT_EXASOL' or 'RDBMS_EXASOL' and then wrap all/most of your extensions into semantic lookaheads like:

``` LOOKAHEAD( { "DIALECT_EXASOL".equals(Features.get("dialect")) } ) production = ExasolProduction() ```

(This is an abstract illustration only, of course.)

My hope and goal was to reduce the performance penalty for common RDMS and use-cases to a minimum.
And to have a template for similar exotic dialects like apache hive etc.

manticore-projects:
You have introduced new tokens `RTRIM` and `LTRIM`, please double-check that those won't collide with parsing the normal functions.

Beside that we are good to merge!

manticore-projects:
Greetings!

I have added already a new Feature for `dialect` which you can set to `Dialect.EXASOL`. 
This feature can be used now in your lookaheads.

PR #2303, opened 2025-08-28
support opengauss "on duplicate key update nothing" grammar

zhangconan:
Add a new feature to support opengauss "on duplicate key update nothing" syntax. I have executed mvn clean package locally, and all code format checks and unit tests have passed.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/insert/Insert.java
+++ b/src/main/java/net/sf/jsqlparser/statement/insert/Insert.java
@@ -167,7 +167,8 @@ public class Insert implements Statement {
 
     @Deprecated
     public boolean isUseDuplicate() {
-        return duplicateAction != null && duplicateAction.getUpdateSets() != null && !duplicateAction.getUpdateSets().isEmpty();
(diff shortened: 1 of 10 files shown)


src/main/java/net/sf/jsqlparser/statement/insert/Insert.java
```diff
@@ -18,18 +18,10 @@
 import net.sf.jsqlparser.statement.ReturningClause;
 import net.sf.jsqlparser.statement.Statement;
 import net.sf.jsqlparser.statement.StatementVisitor;
-import net.sf.jsqlparser.statement.select.PlainSelect;
```
manticore-projects, line 21: Please avoid wildcard imports.

manticore-projects (changes requested):
Looks goo in general, but please avoid Wildcard imports and import single Classes instead.

src/main/java/net/sf/jsqlparser/statement/insert/Insert.java
zhangconan: I have adjusted.

Issue #1142, opened 2021-03-27
No support for LOCK TABLES


ksqsf:
**Describe the bug**
JSqlParser doesn't support `LOCK TABLES` and throws `ParseException: Encountered unexpected token`.

**To Reproduce**
SQL:
```
LOCK TABLES `table` WRITE;
```

Parsing:
```
        val file = File(path).bufferedReader()
        val parser = CCJSqlParser(file.readText())
```

Exception:
```
Exception in thread "main" net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: "LOCK" <S_IDENTIFIER>
    at line 42, column 1.

Was expecting one of:

    "("
    ";"
    "ALTER"
    "BEGIN"
    "CALL"
    "COMMENT"
    "COMMIT"
    "CREATE"
    "DECLARE"
    "DELETE"
    "DESCRIBE"
    "DROP"
    "EXEC"
    "EXECUTE"
    "EXPLAIN"
    "GRANT"
    "INSERT"
    "MERGE"
    "REPLACE"
    "SET"
    "SHOW"
    "TRUNCATE"
    "UPDATE"
    "UPSERT"
    "USE"
    "VALUES"
    "WITH"
    <K_SELECT>

	at net.sf.jsqlparser.parser.CCJSqlParser.generateParseException(CCJSqlParser.java:26538)
	at net.sf.jsqlparser.parser.CCJSqlParser.jj_consume_token(CCJSqlParser.java:26377)
	at net.sf.jsqlparser.parser.CCJSqlParser.Statements(CCJSqlParser.java:511)
```

**Expected behavior**
Parse it into an AST.

**System**
Ref: https://dev.mysql.com/doc/refman/8.0/en/lock-tables.html

ksqsf:
Is there a way to ignore such unparsable statements?

wumpz:
Look into `StatementsTest`. There are examples of `parser.setErrorRecovery(true);`. This skips unparsable statements.

manticore-projects:
See `UnsupportedStatements` as explain in #1945 (unless you will sponsor or provide a specific implementation of this feature.)

PR #2321, opened 2025-10-07
[feat] Support for LOCK-Statements

ANeumann82:
This adds support for LOCK-Statements (Requested for example here https://github.com/JSQLParser/JSqlParser/issues/1142, but we also need it)

It supports:
- DB2: https://www.ibm.com/docs/en/db2/12.1.0?topic=statements-lock-table
- Most of Oracle: https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/LOCK-TABLE.html

I just saw that MySQL has a bit different syntax, but as I don't need it I wouldn't add it for now.

Please have a look and let me know if there's something missing or wrong.


manticore-projects:
Thank you much for your contribution!

PR #2328, opened 2025-10-15
[feat] Support for JSON_TABLE

ANeumann82:


src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
```diff
@@ -761,6 +776,7 @@ TOKEN : /* Data Types */
     | <#TYPE_REAL: "REAL" | "FLOAT4" | "FLOAT">
     | <#TYPE_DOUBLE: "DOUBLE" | "PRECISION"  | "FLOAT8" | "FLOAT64">
     | <#TYPE_VARCHAR: "NVARCHAR" | "VARCHAR" | "NCHAR" | <K_CHAR> | "BPCHAR" | "TEXT" | "STRING" | <K_CHARACTER> | "VARYING">
+    | <#TYPE_VARCHAR2: <K_VARCHAR2>>
```
manticore-projects: This should got to TYPE_VARCHAR since we try to establish a reasonable grouping please.

src/main/java/net/sf/jsqlparser/util/validation/validator/ExpressionValidator.java
```diff
@@ -9,66 +9,7 @@
...
-import net.sf.jsqlparser.expression.TranscodingFunction;
-import net.sf.jsqlparser.expression.TrimFunction;
-import net.sf.jsqlparser.expression.UserVariable;
-import net.sf.jsqlparser.expression.VariableAssignment;
-import net.sf.jsqlparser.expression.WhenClause;
-import net.sf.jsqlparser.expression.WindowElement;
-import net.sf.jsqlparser.expression.WindowOffset;
-import net.sf.jsqlparser.expression.WindowRange;
-import net.sf.jsqlparser.expression.XMLSerializeExpr;
+import net.sf.jsqlparser.expression.*;
```
manticore-projects: Please avoid wildcard imports.

src/main/java/net/sf/jsqlparser/util/deparser/SelectDeParser.java
```diff
@@ -9,13 +9,7 @@
...
 package net.sf.jsqlparser.util.deparser;
 
-import net.sf.jsqlparser.expression.Alias;
-import net.sf.jsqlparser.expression.Expression;
-import net.sf.jsqlparser.expression.ExpressionVisitor;
-import net.sf.jsqlparser.expression.MySQLIndexHint;
-import net.sf.jsqlparser.expression.OracleHint;
-import net.sf.jsqlparser.expression.SQLServerHints;
-import net.sf.jsqlparser.expression.WindowDefinition;
+import net.sf.jsqlparser.expression.*;
```
manticore-projects: please avoid wildcard imports.

src/main/java/net/sf/jsqlparser/statement/select/FromItemVisitorAdapter.java
```diff
@@ -148,4 +149,10 @@ public <S> T visit(Import imprt, S context) {
     public <S> T visit(FromQuery fromQuery, S context) {
         return fromQuery.accept(selectVisitor, context);
     }
+
+    @Override
```
manticore-projects: please implement this properly (since the Adaptor is actually used)

src/main/java/net/sf/jsqlparser/expression/JsonTable.java
```diff
@@ -0,0 +1,200 @@
...
+
+import net.sf.jsqlparser.parser.ASTNodeAccessImpl;
+import net.sf.jsqlparser.statement.select.AbstractFromitem;
+import net.sf.jsqlparser.statement.select.FromItem;
+import net.sf.jsqlparser.statement.select.FromItemVisitor;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class JsonTable extends AbstractFromitem implements FromItem {
```
manticore-projects: is `expression` really the best package? I don't mind but I would have rather packed this close to normal `Table` or `TableFunction`. No show stopper though.

manticore-projects (changes requested):
Great work and I do appreciate your effort and dedication. Thank you much!
A few small concerns though, nothing severe just the normal German nit-picking.

Kudos and cheers!

src/main/java/net/sf/jsqlparser/util/validation/validator/ExpressionValidator.java
ANeumann82: Yeah, my IntelliJ wasn't configured correctly.

src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
ANeumann82: This is a problem. When I move this into the `TYPE_VARCHAR`, the generated TokenManager gets a `code too large`. I guess we're reaching the limits of JavaCC again...

I'll have to look into why exactly this happens, but I've not applied this change yet in my last commit.

manticore-projects:
Greetings @ANeumann82, as promised I have completely reworked the token handling.
This should solve the  "code too large" problem once and for all. We also have achieved **much** better performance.

So we should be able to proceed with your PR please.
@dengliming: in case you get bored and look for a challenge :-D

ANeumann82:
Hey @manticore-projects, thanks for the update. I'll take a look at this in the coming week or so.

@dengliming Yeah, I'll have to take a look at that. Seems like there was some JSON_TABLE functionality added a couple weeks ago, while my PR is a bit older. I'll take a look if and how I can consolidate that.

manticore-projects (changes requested):
Great stuff and work overall.
I only have 1 question on the grammar and it is optional. If there is no quick improvement possible, we could just go ahead.

(Ignore the Maven failures please, I will rip out this old garbage very soon anyways.)

PR #2348, opened 2025-11-21
Support PostgreSQL-specific [CREATE SEQUENCE name START n ...]

### LINKED ISSUE #2347: [BUG] JSQLParser 5.4 PostgreSQL treats the WITH and BY in CREATE SEQUENCE expressions as optional
### Failing SQL Feature:

PostgreSQL treats the `WITH` in `CREATE SEQUENCE name START WITH …` expressions as [optional](https://www.postgresql.org/docs/current/sql-createsequence.html). At least one SQL generator (Hibernate) is generating PostgreSQL-specific SQL when creating new sequences, without the `WITH`.

### SQL Example:

```sql
CREATE SEQUENCE example_seq START 1 INCREMENT 1;
```

### Software Information:

PostgreSQL [7.1](https://www.postgresql.org/docs/7.1/sql-createsequence.html) through to current (7.x did not support `WITH` at all, which is maybe why Hibernate is not including it)

JSqlParser version: 4.9 (this project still requires Java 8 for the moment, unfortunately)

However, [JSqlParser 5.4-SNAPSHOT still requires it](https://jsqlparser.github.io/JSqlParser/syntax_snapshot.html):

Image

mjog:
I.e. no `START WITH`

Fixes #2347

### DIFF
--- a/src/main/java/net/sf/jsqlparser/schema/Sequence.java
+++ b/src/main/java/net/sf/jsqlparser/schema/Sequence.java
@@ -158,7 +158,7 @@ public class Sequence extends ASTNodeAccessImpl implements MultiPartName {
      * The available parameters to a sequence
      */
     public enum ParameterType {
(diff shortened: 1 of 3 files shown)


mjog:
Static Code Analysis is complaining about a `_` in a test method name, but I used the same format that was in the file already.

manticore-projects:

Issue #2247, opened 2025-05-28
[FEATURE] JSQLParser 5.3 cannot parse name of the index in alter table


FredyFerrari:
### Grammar or Syntax Description
- `ALTER TABLE` with a `USING INDEX` clause cannot parse the name of the index. The name of the index is reported in an UnsupportedStatement
- Example: `ALTER TABLE` with a `USING INDEX PK_TKONS` is not supported

### SQL Example
- Simplified Query Example, focusing on the failing feature
    ```sql
    ALTER TABLE TNWAV ADD CONSTRAINT PK_TNWAV PRIMARY KEY (NWNAME, ZEILE, BESTGRU) USING INDEX PK_TNWAV
    ```

PR #2384, opened 2026-02-23
Fix ALTER TABLE with a USING INDEX clause

dengliming:
Fixes [https://github.com/JSQLParser/JSqlParser/issues/2247](https://github.com/JSQLParser/JSqlParser/issues/2247)


manticore-projects:
Thank your for your work and effort!

Issue #2329, opened 2025-10-22
[BUG] JSQLParser 5.4 SNAPSHOT : Oracle : Outer join case with NVL


mantesap:
### Failing SQL Feature:
- Oracle supports outer joins with (+) notation and can be used in conjunction with nvl or coalesce
- Example: The following example cannot be parsed

### SQL Example:
- Simplified Query Example
    ```sql
     SELECT * FROM dual d, dual d2 where d.dummy = nvl(d2.dummy (+),'y')
    ```

### Software Information:
- JSqlParser version 5.4-SNAPSHOT
- Database: Oracle

PR #2386, opened 2026-02-24
Fix oracle outer join case with nvl/coalesce

dengliming:
Fixes [https://github.com/JSQLParser/JSqlParser/issues/2329](https://github.com/JSQLParser/JSqlParser/issues/2329)


manticore-projects:
You good Sir are highly productive and I tip my hat!

Issue #2125, opened 2024-12-18
[BUG] JSQLParser Version : 4.9  RDBMS :  not support clickhouse quantile expression


Knight-ZXW:
Always check against the **Latest SNAPSHOT of JSQLParser** and the [Syntax Diagram](https://jsqlparser.github.io/JSqlParser/syntax_snapshot.html)

### Failing SQL Feature:
 quantile(0.95)(cost)   not support

### SQL Example:
select toStartOfDay(timestamp) AS date, count(1) as count, quantile(0.95)(cost) as `95cost` from apm_log_event

### Software Information:
- JSqlParser 4.9
- Database :Clickhouse

![image](https://github.com/user-attachments/assets/aead0caf-ed25-4667-ac22-ae85cbc0e12e)

manticore-projects:
Greetings.

This is unsupported syntax far away from any standard and you would need to provide or sponsor an implementation to make it happen.

PR #2393, opened 2026-03-03
fix: support ClickHouse parametric aggregate calls

dengliming:
Fixes [https://github.com/JSQLParser/JSqlParser/issues/2125](https://github.com/JSQLParser/JSqlParser/issues/2125)

Handle chained function args like quantile(0.95)(cost) and add regression test for #2125.


manticore-projects:
Thank you!

Issue #2362, opened 2026-01-12
[BUG] JSQLParser 5.4-SNAPSHOT : Clickhouse : SETTINGS not supported in Clickhouse SELECT


tomershay:
### Failing SQL Feature:

When you add `SETTINGS` ([docs](https://clickhouse.com/docs/operations/settings/query-level?utm_source=chatgpt.com)) after a query, you tell ClickHouse to temporarily use different values for certain internal options. These options control things like `max_threads`, `max_rows_to_read`, etc.

### Reproduce the issue with this query:

```
SELECT *
FROM events
SETTINGS max_threads = 1;
```

### Error:

`ParseException: Encountered: <S_IDENTIFIER> / "max_threads", at line 3, column 10,`

PR #2395, opened 2026-03-04
fix(clickhouse): support SELECT ... SETTINGS

dengliming:
Fixes [https://github.com/JSQLParser/JSqlParser/issues/2362](https://github.com/JSQLParser/JSqlParser/issues/2362)


manticore-projects:
Thank you!

PR #2425, opened 2026-04-08
fix: split CCJSqlParserTokenManager static initializers to avoid 64KB…

hayssams:
… method limit

The generated CCJSqlParserTokenManager class has a clinit method that reaches 64,452 bytes -- dangerously close to the JVM 65,535 byte limit. ASM-based tools like sbt-assembly add transformation overhead that pushes it over, causing MethodTooLargeException during fat JAR creation.

Add a splitTokenManagerStaticInit Gradle task that post-processes the generated Java file after JavaCC code generation, extracting 6 large static array initializations (stringLiterals, jjstrLiteralImages, jjmatchKinds, jjnewLexState, jjcompositeState, jjnextStateSet) into separate private static methods. This reduces clinit from 64,452 bytes to 404 bytes.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -198,6 +198,79 @@ compileJavacc {
     ]
 }
 
+// Post-process the generated CCJSqlParserTokenManager.java to split large static
+// array initializers into separate methods, preventing the <clinit> method from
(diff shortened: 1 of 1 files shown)


manticore-projects:
Thank you! Please see https://github.com/javacc/javacc-8-core/issues/3
I will try to move this upstream to JavaCC itself.

PR #2437, opened 2026-06-15
use Locale.ROOT for case-insensitive keyword normalisation

dxbjavid:
several parser helpers fold SQL keywords and identifiers to upper case with the no-argument toUpperCase(), which uses the JVM default locale. under a Turkish or Azeri locale a keyword containing the letter i (union, intersect, minus, time, timestamp, like and friends) folds to a dotted İ instead of ASCII I, so the following Enum.valueOf lookup throws IllegalArgumentException and a valid statement fails to parse, while deparsing emits the wrong casing. switching the conversions to Locale.ROOT keeps case folding of SQL text independent of the runtime locale.


manticore-projects:
Looks good but I need to fix the build script first :-(

dxbjavid:
no rush at all. happy to rebase onto master once the build script is sorted, just give me a shout if you need anything changed on my side.

dxbjavid:
Thank you for the merge. Appreciated

PR #2444, opened 2026-07-24
[Fix] #2443 - NPE In TableNamesFinder with AnalyticExpression

### LINKED ISSUE #2443: [BUG] NPE in TableNamesFinder in AnalyticExpression visitor
```sql
SELECT c, SUM(COUNT(*)) OVER (ORDER BY c ASC ROWS UNBOUNDED PRECEDING) FROM tbl GROUP BY c
```
The window element here does not contain a range or an offset, yet this will throw an NPE in TableNamesFinder when it tries to walk the AnalyticExpression:

```java
        String sqlStr = "SELECT c, SUM(COUNT(*)) OVER (ORDER BY c ASC ROWS UNBOUNDED PRECEDING) FROM tbl GROUP BY c";

        assertThatCode(() -> TablesNamesFinder.findTables(sqlStr))
                .doesNotThrowAnyException();
```

Gives:
```
Expecting code not to raise a throwable but caught                                                                                                                                                                                                                                                                                                                                                                                          
  "java.lang.NullPointerException: Cannot invoke "net.sf.jsqlparser.expression.WindowRange.getStart()" because the return value of "net.sf.jsqlparser.expression.WindowElement.getRange()" is null 
```

davyboyhayes:
If you attempt to tree walk an AnalyticExpression in TableNamesFinder with an SQL that has a window function without a range and without an offset, such as `SELECT c, SUM(COUNT(*)) OVER (ORDER BY c ASC ROWS UNBOUNDED PRECEDING) FROM tbl GROUP BY c`, an NPE is thrown. This adds defensive checks around the range and offset expressions being null.

Fixes #2443

### DIFF
--- a/src/main/java/net/sf/jsqlparser/util/TablesNamesFinder.java
+++ b/src/main/java/net/sf/jsqlparser/util/TablesNamesFinder.java
@@ -766,17 +766,20 @@ public class TablesNamesFinder<Void>
         }
 
         if (analytic.getWindowElement() != null) {
-            if (analytic.getWindowElement().getRange().getStart().getExpression() != null) {
-                analytic.getWindowElement().getRange().getStart().getExpression().accept(this,
-                        context);
+            if (analytic.getWindowElement().getRange() != null) {
+                if (analytic.getWindowElement().getRange().getStart().getExpression() != null) {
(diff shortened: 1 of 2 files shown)


manticore-projects:
Thank you!

Issue #2403, opened 2026-03-08
compileJavacc emits JavaCC warnings for shadowed literals and unreachable DATA_TYPE branches


dengliming:
```
Warning: RR cannot be matched as a string literal token. It will be matched as the non string literal token of kind 149.
Warning: RS cannot be matched as a string literal token. It will be matched as the non string literal token of kind 149.
Warning: TEXT cannot be matched as a string literal token. It will be matched as the non string literal token of kind 407.
Warning: TIMESTAMPTZ cannot be matched as a string literal token. It will be matched as the non string literal token of kind 92.
Warning: TIMEZONE cannot be matched as a string literal token. It will be matched as the non string literal token of kind 522.
Warning: TINYTEXT cannot be matched as a string literal token. It will be matched as the non string literal token of kind 407.
Warning: UR cannot be matched as a string literal token. It will be matched as the non string literal token of kind 149.
Warning: MATCHING cannot be matched as a string literal token. It will be matched as the non string literal token of kind 522.
Warning: DISTRIBUTED cannot be matched as a string literal token. It will be matched as the non string literal token of kind 522.
Warning: RECOVERY cannot be matched as a string literal token. It will be matched as the non string literal token of kind 522.
Warning: Line 748, Column 5: Regular Expression choice : K_BYTES can never be matched as : DATA_TYPE
Warning: Line 748, Column 5: Regular Expression choice : K_BINARY can never be matched as : DATA_TYPE
Warning: Line 748, Column 5: Regular Expression choice : K_BOOLEAN can never be matched as : DATA_TYPE
Warning: Line 748, Column 5: Regular Expression choice : K_UNSIGNED can never be matched as : DATA_TYPE
Warning: Line 748, Column 5: Regular Expression choice : K_SIGNED can never be matched as : DATA_TYPE
Warning: Line 748, Column 5: Regular Expression choice : K_CHARACTER can never be matched as : DATA_TYPE
Warning: Line 748, Column 5: Regular Expression choice : K_CHAR can never be matched as : DATA_TYPE
```

manticore-projects:
Greetings!

Thank you, I am aware of this and have started to rework the Token Logic completely. I will share more details on this next week with you.

Issue #2339, opened 2025-11-12
[BUG] JSQLParser 5.3 : PostgreSQL : Cannot parse DROP COLUMN command where data is the column name


willneedham93:
### Failing SQL Feature:
- Cannot parse DROP COLUMN command where `data` is the column name
- `ALTER TABLE example DROP COLUMN data;` can't be parsed

### SQL Example:
- ```sql
    ALTER TABLE example DROP COLUMN data;
    ```

### Software Information:
- JSQLParser 5.3
- PostgreSQL

### Tips:
Appears that adding a column with name `data` works ok

PR #2448, opened 2026-08-05
fix: add 'K_TYPE' to KeywordOrIdentifier to allow 'type' as a column name

### LINKED ISSUE #2447: [BUG] JSQLParser 5.3 : MySQL : Cannot parse DROP COLUMN where "type" is the column name
### Failing SQL Feature:

- Cannot parse `ALTER TABLE ... DROP COLUMN` command where `type` is the column name
- `TYPE` is a non-reserved keyword in MySQL 8.0 (see [MySQL keyword list](https://dev.mysql.com/doc/refman/8.0/en/keywords.html)), so `ALTER TABLE nullable_tbl DROP COLUMN type;` is valid MySQL syntax without any quoting

### SQL Example:

```sql
ALTER TABLE nullable_tbl DROP COLUMN type;
```

fails with:

```
net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: "type" "TYPE"
    at line 1, column 38.

Was expecting one of:

    "IF"
    "NAME"
    <S_IDENTIFIER>
    <S_QUOTED_IDENTIFIER>
```

while the same identifier parses fine in other contexts:

```sql
SELECT type FROM nullable_tbl;                  -- OK
ALTER TABLE nullable_tbl DROP COLUMN `type`;    -- OK (quoted)
```

### Software Information:

- JSQLParser 5.3 (also reproducible with current master: `KeywordOrIdentifier()` accepts `K_NAME | K_NEXT | K_VALUE | K_PUBLIC | K_STRING | K_DATA` but not `K_TYPE`)
- MySQL

### Tips:

- Same class of bug as #2339 (column named `data`), which was fixed by #2340 adding `K_DATA` to `KeywordOrIdentifier()`. Adding `K_TYPE` the same way fixes this case — I will raise a PR shortly.
- For reference, other MySQL non-reserved words that currently fail in the same `DROP COLUMN` position on master: `comment`, `action`, `position`, `format`, `domain` (and the datatype-token family `text`, `date`, `year`, `timestamp`, which probably needs a different approach).

minleejae:
Fixes #2447

### Problem

`TYPE` is a non-reserved keyword in MySQL 8.0, so this is valid MySQL syntax without quoting:

```sql
ALTER TABLE nullable_tbl DROP COLUMN type;
```

but it fails with `ParseException: Encountered unexpected token: "type" "TYPE"`, while the same identifier parses fine in a `SELECT` (via `RelObjectName()`, since `K_TYPE` is inside the non-reserved word range).

### Fix

Adds `K_TYPE` to `KeywordOrIdentifier()`, following the exact same approach as #2340 (which added `K_DATA` for the same class of bug, see #2339).

This covers all `KeywordOrIdentifier()` call sites: `DROP [COLUMN]`, `RENAME [COLUMN]`, and `CHANGE [COLUMN]`.

### Tests

- `testAlterTableDropColumnIssue2447` — `ALTER TABLE test DROP COLUMN type`
- `testAlterTableChangeColumnIssue2447` — `ALTER TABLE tb_test CHANGE type INT (10)`
- `testAlterTableRenameColumnIssue2447` — `ALTER TABLE test_table RENAME COLUMN type TO type2`

Full test suite passes (4646 tests, 0 failures). JavaCC parser generation reports the same pre-existing warnings as master (#2403) with no new choice conflicts.

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -1756,7 +1756,7 @@ TOKEN:
  * Supported tokens:
  *  - <S_IDENTIFIER>: Standard unquoted SQL identifier
  *  - <S_QUOTED_IDENTIFIER>: Quoted identifier (e.g., `identifier` or "identifier")
- *  - <K_NAME>, <K_NEXT>, <K_VALUE>, <K_PUBLIC>, <K_STRING>, <K_DATA>: Specific keywords treated as identifiers
+ *  - <K_NAME>, <K_NEXT>, <K_VALUE>, <K_PUBLIC>, <K_STRING>, <K_DATA>, <K_TYPE>: Specific keywords treated as identifiers
  *
  * @return Token representing the identifier or keyword used as identifier
  */
@@ -1774,6 +1774,7 @@ Token KeywordOrIdentifier():
       | tk = <K_PUBLIC>
       | tk = <K_STRING>
       | tk = <K_DATA>
+      | tk = <K_TYPE>
     )
     { return tk; }
 }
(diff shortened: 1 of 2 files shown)


manticore-projects:
Thank you much!

PR #2450, opened 2026-08-07
fix: stop 0x hex literal from swallowing trailing whitespace (#2435)

### LINKED ISSUE #2435: [BUG] JSQLParser 5.4-SNAPSHOT : MySQL : 0x hexadecimal literal (0xFF) not supported as select item
**Failing SQL Feature**
  
MySQL supports a hexadecimal literal shorthand using the `0x` prefix, e.g. `0xFF`. In numeric contexts it evaluates to its integer value (`0xFF` = `255`); otherwise it is treated as a binary string. This is a MySQL-specific form (distinct from the SQL-standard `x'1F'` hex-string literal) and is commonly used to express byte/hex constants.

  See the MySQL docs:
  https://dev.mysql.com/doc/refman/8.0/en/hexadecimal-literals.html

  **SQL Example**

  ```sql
  CREATE TABLE t (
    id INT PRIMARY KEY
  );

  INSERT INTO t (id) VALUES (1);

  SELECT 0xFF FROM t;
  ```
  **Parsing error**

`ParseException: Encountered: <S_IDENTIFIER> / "t", at line 1, column 18, in lexical state DEFAULT.`

fudianchn:
## What

`SELECT 0xFF FROM t` (any `0x`-hex literal followed by whitespace/a keyword) failed to parse, while `SELECT 0xFF,1 FROM t` parsed fine. This makes the `0x` hex literal form stop greedily consuming trailing whitespace.

## Why

MySQL and other dialects accept `0x`-prefixed hex literals as values/select items (#2435). Today such a literal followed by `FROM` or a keyword fails:

```
SELECT 0xFF FROM t
  -> ParseException: Encountered: <S_IDENTIFIER> / "t", at line 1, column 18
```

while the same literal followed by a comma parses fine — a confusing inconsistency.

## Root cause

`HEX_VALUE` was defined as `["0"-"9","A"-"F", " "]` and shared by **both** hex literal forms in `S_HEX`:

```
<S_HEX: ("X" ("'" ( <HEX_VALUE> )* "'" (" ")*)+ | "0x" ( <HEX_VALUE> )+ )>
```

The space is intentional for the `X'..'` form, which allows internal whitespace (e.g. `X'01 bc 2a'`, exercised by `StringValueTest#testParseInput_BYTEA`). But for the `0x` form, `<HEX_VALUE>+` then greedily matched the trailing space **and** any following hex letter — so `0xFF FROM` was tokenised as a single `S_HEX` (`0xFF F`), leaving `ROM t`, and the parser choked on `t` at column 18.

## How

Split the shared token so each form gets the character class it needs:

- `X'..'` keeps `HEX_VALUE` (with space — preserves `X'01 bc 2a'`).
- `0x..` uses a new `HEX_DIGIT` token (`["0"-"9","A"-"F"]`, no space).

`IGNORE_CASE = true` already makes both forms match upper/lower case, so `0xff` works too. `HexValue` itself is unchanged.

## Testing

- New `HexValueTest#testHexLiteralFollowedByPrimaryKeyword`: `SELECT 0xFF FROM t`, `SELECT 0xff FROM t`, `SELECT 0xDEADBEEF FROM t`, `SELECT 0xab FROM dual` — all parse to a `HexValue` select item.
- Full suite: `mvn test` → **4647 tests, 0 failures, 0 errors** (26 pre-existing skipped), including the `X'01 bc 2a'` regression in `StringValueTest` and existing `InsertTest#testHexValues*`.
- `checkstyle:check` / `pmd:check` / `license:check-file-header` clean.

Fixes #2435

### DIFF
--- a/src/test/java/net/sf/jsqlparser/expression/HexValueTest.java
+++ b/src/test/java/net/sf/jsqlparser/expression/HexValueTest.java
@@ -9,6 +9,8 @@
  */
 package net.sf.jsqlparser.expression;
 
+import static net.sf.jsqlparser.test.TestUtils.assertSqlCanBeParsedAndDeparsed;
+
 import net.sf.jsqlparser.JSQLParserException;
 import net.sf.jsqlparser.parser.CCJSqlParserUtil;
 import net.sf.jsqlparser.statement.select.PlainSelect;
@@ -53,7 +55,7 @@ class HexValueTest {
 
     private static void assertHexSelectItem(String sql, String expectedDigits)
             throws JSQLParserException {
-        PlainSelect select = (PlainSelect) CCJSqlParserUtil.parse(sql);
+        PlainSelect select = (PlainSelect) assertSqlCanBeParsedAndDeparsed(sql, true);
         Expression expr = select.getSelectItem(0).getExpression();
         Assertions.assertTrue(expr instanceof HexValue, () -> "Expected HexValue for: " + sql);
         Assertions.assertEquals(expectedDigits, ((HexValue) expr).getDigits());
--- a/src/test/java/net/sf/jsqlparser/expression/HexValueTest.java
+++ b/src/test/java/net/sf/jsqlparser/expression/HexValueTest.java
@@ -39,4 +39,23 @@ class HexValueTest {
 
         Assertions.assertEquals("'\\xC3\\xBC'", hex3.getBlob().toString());
     }
+
+    @Test
+    void testHexLiteralFollowedByPrimaryKeyword() throws JSQLParserException {
+        // Issue #2435: a "0x"-prefixed hex literal must not greedily consume the
+        // following whitespace nor adjacent hex letters (e.g. the "F" of FROM),
+        // which previously made "SELECT 0xFF FROM t" fail to parse.
+        assertHexSelectItem("SELECT 0xFF FROM t", "FF");
+        assertHexSelectItem("SELECT 0xff FROM t", "ff");
+        assertHexSelectItem("SELECT 0xDEADBEEF FROM t", "DEADBEEF");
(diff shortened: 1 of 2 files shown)


src/test/java/net/sf/jsqlparser/expression/HexValueTest.java
```diff
@@ -39,4 +39,23 @@ void testHexCode() throws JSQLParserException {
...
+        // which previously made "SELECT 0xFF FROM t" fail to parse.
+        assertHexSelectItem("SELECT 0xFF FROM t", "FF");
+        assertHexSelectItem("SELECT 0xff FROM t", "ff");
+        assertHexSelectItem("SELECT 0xDEADBEEF FROM t", "DEADBEEF");
+        assertHexSelectItem("SELECT 0xab FROM dual", "ab");
+    }
+
+    private static void assertHexSelectItem(String sql, String expectedDigits)
+            throws JSQLParserException {
+        PlainSelect select = (PlainSelect) CCJSqlParserUtil.parse(sql);
```
manticore-projects: Again, please use `assertThatSqlCanBeParsedAndDeparsed`, thanks!

manticore-projects (changes requested):

src/test/java/net/sf/jsqlparser/expression/HexValueTest.java
fudianchn: Sorry about that—I was already applying the same test update here, but your review came in before I had finished. I have now updated the regression test in `255bc72` accordingly, while retaining the AST assertions. Thanks!

manticore-projects:
Thank you much for your work and effort.

PR #2451, opened 2026-08-07
fix: support nested parametric data types as CAST target

### LINKED ISSUE #2441: [BUG] JSQLParser 5.4-SNAPSHOT : ClickHouse : Parametric type Nullable(Decimal(p, s)) not supported as CAST target
## Failing SQL Feature

  ClickHouse supports parametric (constructor-style) data types as the target of a `CAST`, including nested ones such as `Nullable(Decimal(10, 2))`. Here `Decimal(10, 2)` declares a decimal with precision 10 and scale 2, and `Nullable(...)` wraps it so the column can also hold NULL. JSQLParser fails to parse this type expression when it is used as the target type of a `CAST(... AS ...)`.

  See the ClickHouse docs:
  - https://clickhouse.com/docs/en/sql-reference/data-types/nullable
  - https://clickhouse.com/docs/en/sql-reference/data-types/decimal

  ## Full SQL Example
  
  ```
CREATE TABLE cast_demo (id UInt32, x Float64) ENGINE = MergeTree ORDER BY id;
  INSERT INTO cast_demo (id, x) VALUES (1, 10.5), (2, 3.14159), (3, 1234567.891), (4, 0);
  SELECT CAST(x AS Nullable(Decimal(10, 2))) FROM cast_demo;
```

  ## Parsing Error

```
ParseException: Encountered: <OPENING_BRACKET> / "(", at line 3, column 25, in lexical state DEFAULT.
Was expecting one of these terminals within expansion starting at 9127:5:
    <K_FORMAT> (inside 9125:5) ...
    <CLOSING_BRACKET> (inside 9127:5) 
```

Tested with ClickHouse version: 25.8

### PR DESCRIPTION
**Pull request by fudianchn**
## What
Allow a parametric (constructor-style) data type — including a nested one such as `Nullable(Decimal(10, 2))` — to be used as the target type of a `CAST`.

## Why
ClickHouse supports parametric data types as a `CAST` target, and they can be nested:

```sql
SELECT CAST(x AS Nullable(Decimal(10, 2))) FROM cast_demo;
```

JSQLParser failed on the inner opening bracket:

```
ParseException: Encountered: <OPENING_BRACKET> / "(", ... Was expecting ... <CLOSING_BRACKET>
```

Reported in #2441.

## How
The argument production of `ColDataType` only accepted atomic tokens (`S_LONG`/`K_MAX` with optional `BYTE`/`CHAR`, `S_CHAR_LITERAL`, `S_IDENTIFIER`, `K_CHAR`), so a nested parametric type was unreachable.

This adds one alternative to the argument loop: when the next token is a `DATA_TYPE` or identifier immediately followed by `(`, parse the argument as a nested `ColDataType` (recursing through the existing production) and store its `toString()` rendering in the arguments string list. The new alternative is gated by an explicit `LOOKAHEAD`, so the existing atomic arguments are unaffected.

Kept intentionally narrow: only parametric arguments (a type immediately followed by its own argument list) are handled. A bare type-keyword argument (e.g. `String` in `Nullable(String)`) is a separate concern and out of scope for this issue.

## Testing
- New regression test `ClickHouseTest#testCastToNestedParametricTypeIssue2441` covering `Nullable(Decimal(10, 2))` (the reported case) and `LowCardinality(Decimal(10, 2))` (another ClickHouse parametric type wrapping a precision type). Both parse and round-trip.
- `mvn clean test`: 4647 tests, 0 failures, 0 errors (26 skipped) — no regressions on the shared `ColDataType` production.
- `mvn spotless:check`: clean.

Fixes #2441

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -10707,6 +10707,7 @@ ColDataType ColDataType():
     List<Integer> array = new ArrayList<Integer>();
     List<String> name;
     ColDataType arrayType;
+    ColDataType nestedType = null;
 
     int precision = -1;
     int scale = -1;
@@ -10756,6 +10757,9 @@ ColDataType ColDataType():
         LOOKAHEAD(2) "(" {tk2 =null;}
         (
             (
+            LOOKAHEAD((<DATA_TYPE> | <S_IDENTIFIER>) "(") nestedType=ColDataType()
+            |
+            (
                 (
                     ( tk=<S_LONG> | tk=<K_MAX> ) [ LOOKAHEAD(2) (tk2=<K_BYTE> | tk2=<K_CHAR>) ]
                 )
@@ -10766,8 +10770,14 @@ ColDataType ColDataType():
                 |
                 tk=<K_CHAR>
             )
+            )
 		    {
-		        argumentsStringList.add(tk.image + (tk2!=null?" " + tk2.image:""));
+            if (nestedType != null) {
+                argumentsStringList.add(nestedType.toString());
+                nestedType = null;
+            } else {
+                argumentsStringList.add(tk.image + (tk2!=null?" " + tk2.image:""));
+            }
             }
 
             [ "," ]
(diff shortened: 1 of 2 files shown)


manticore-projects:
Thank you for your contribution, good job!

Issue #2088, opened 2024-10-16
[BUG] JSQLParser Version 5.1-SNAPSHOT: Lateral View with two aliases recognizes second alias as table


arodriguezdlc:
### Failing SQL Feature:

When using a LATERAL VIEW with two aliases, the second alias is incorrectly being interpreted as a JOIN with a table. Instead, it should be detected as a second alias for the exploded data.

The LateralView class currently has a method getColumnAlias that returns only one alias. This behavior could be incorrect because in cases where there are multiple aliases, such as with exploded arrays or maps, the method should return a list of aliases, not just one.

### SQL Example:

```sql
SELECT k, v 
FROM table 
LATERAL VIEW EXPLODE(a) exploded_data AS k, v;
```

In this query, the alias k is correctly detected as the first alias for the exploded data. However, the alias v is mistakenly identified as a table named v, instead of being treated as the second alias for the exploded data.

The expected behavior is that both k and v should be recognized as aliases for the exploded data, with no table join implied.

### Software Information:
- JSqlParser version: 5.1-SNAPSHOT
- Database: Spark SQL

manticore-projects:
Greetings!

Would that not be a `Struct` based Alias? We support this already, but not in the form of a  plain comma separated column list.
We support compound aliases like `A(k,v)`.

manticore-projects:
A bit quick and dirty, but it works now. Cheers!

arodriguezdlc:
Thank you @manticore-projects !! We are testing with the new snapshot. Any plans for a 5.1 release? Cheers!

PR #2452, opened 2026-08-07
fix: allow arbitrary column aliases in LATERAL VIEW

### LINKED ISSUE #2433: [BUG] JSQLParser Version : 5.3 : LATERAL VIEW with three or more column aliases silently mis-parses extras as cross-join tables
Failing SQL Feature:
  
  When using a LATERAL VIEW with three or more column aliases, only the first two aliases are correctly absorbed into LateralView.columnAlias. The third and subsequent aliases are silently mis-parsed as additional comma-separated tables in the FROM clause (implicit cross joins) and end up in PlainSelect.joins as bare Table items.

  I noticed that closed issue #2088 addressed the case of a LATERAL VIEW with two column aliases. However, that fix appears to cover only the two-alias scenario; the grammar production for the column-alias list is still capped at two identifiers, so any case with three or more aliases is still broken in the same way described in #2088. Hive/Spark allow arbitrary-length alias lists (for example, json_tuple typically yields many output columns), so this remaining gap is hit easily in real workloads.
  
  The parse does not raise a syntax error — the failure is silent. Worse, Statement.toString() re-emits a textually identical SQL string (because the leaked aliases get re-printed as the joined table list), which makes the broken AST very hard to detect by round-trip inspection.

  SQL Example:

  SELECT a
  FROM t
  LATERAL VIEW json_tuple(j, 'a', 'b', 'c', 'd', 'e', 'f', 'g')
         x AS c1, c2, c3, c4, c5, c6, c7;

  In this query, c1 and c2 are correctly registered as column aliases of the lateral view. However, c3, c4, c5, c6, c7 are incorrectly interpreted as five separate tables joined to the FROM clause. Reflective inspection of the parsed AST shows:

  - PlainSelect.fromItem = Table("t")
  - PlainSelect.joins = [Join(simple=true, right=Table("c3")), ..., Join(simple=true, right=Table("c7"))]  ← unexpected
  - LateralView.columnAlias.aliasColumns.size() = 2  ← should be 7

  The expected behavior is that all of c1, c2, c3, c4, c5, c6, c7 are recognized as column aliases of the lateral view, with PlainSelect.joins being null (or empty), and no implicit table joins implied.

  For comparison, the parenthesized form LATERAL VIEW ... x AS (c1, c2, ..., c7) is rejected outright with a ParseException at the (, so there is no alternative syntax that currently works for more than two aliases.

  Boundary behavior observed (jsqlparser 5.3):

  Aliases                       Result
  ----------------------------  -----------------------------------------------------
  AS c1, c2                     OK     — both absorbed into columnAlias.aliasColumns
  AS c1, c2, c3                 WRONG  — c3 becomes a join Table
  AS c1, c2, ..., cN  (N >= 3)  WRONG  — c3..cN become join Tables
  AS (c1, c2, ..., cN)          ERROR  — ParseException at "("

  Software Information:
  
  - JSqlParser version: 5.3
  - Database: Spark SQL (Hive-style LATERAL VIEW)

fudianchn:
## What
Allow a Hive/Spark `LATERAL VIEW` to declare an arbitrary number of column aliases, not just two.

## Why
`LATERAL VIEW ... x AS c1, c2, c3, ..., cN` is valid Hive/Spark SQL — `json_tuple` and similar UDTFs commonly yield many output columns. The grammar only absorbed the first two aliases (#2088); from the third onward the aliases were silently parsed as comma-separated tables in the `FROM` clause (implicit cross joins):

- `PlainSelect.joins` incorrectly contained `Join(Table("c3")) ... Join(Table("cN"))`
- `LateralView.columnAlias.aliasColumns` stopped at 2

Because `toString()` re-emits the leaked aliases as the join list, the broken AST round-trips to textually identical SQL, so the bug is invisible to round-trip checks.

Reported in #2433.

## How
The column-alias production (`LateralView`) handled the second alias with an optional `[ "," name ]` block. This adds a `("," name)*` loop immediately after it, so the 3rd…Nth aliases are consumed by the lateral view instead of leaking to the `FROM` clause. The existing single- and two-alias behaviour is unchanged (the loop simply runs zero times).

## Testing
- New `HiveTest#testLateralViewManyColumnAliasesIssue2433` parses `LATERAL VIEW json_tuple(...) x AS c1, c2, c3, c4, c5, c6, c7` and asserts at the AST level (a round-trip assertion would be vacuous here): `PlainSelect.joins` is `null` and `LateralView.columnAlias.aliasColumns` has size 7.
- Verified the test fails on `master` (`expected: <null> but was: <[c3, c4, c5, c6, c7]>`) and passes with this change.
- `mvn clean test`: 4648 tests, 0 failures, 0 errors (26 skipped) — the existing two-alias case (#2088) and all other tests unchanged.
- `mvn spotless:check`: clean.

Fixes #2433

### DIFF
--- a/src/test/java/net/sf/jsqlparser/statement/select/HiveTest.java
+++ b/src/test/java/net/sf/jsqlparser/statement/select/HiveTest.java
@@ -9,12 +9,16 @@
  */
 package net.sf.jsqlparser.statement.select;
 
-import net.sf.jsqlparser.schema.Table;
-import org.junit.jupiter.api.Test;
-
 import static net.sf.jsqlparser.test.TestUtils.assertSqlCanBeParsedAndDeparsed;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import net.sf.jsqlparser.expression.Alias;
+import net.sf.jsqlparser.parser.CCJSqlParserUtil;
+import net.sf.jsqlparser.schema.Table;
+import org.junit.jupiter.api.Test;
 
 public class HiveTest {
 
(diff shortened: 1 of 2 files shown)


src/test/java/net/sf/jsqlparser/statement/select/HiveTest.java
```diff
@@ -54,4 +58,30 @@ public void testGroupSimplified() throws Exception {
                 + "GROUP BY GROUPING SETS (())";
         assertSqlCanBeParsedAndDeparsed(sql, true);
     }
+
+    @Test
+    public void testLateralViewManyColumnAliasesIssue2433() throws Exception {
```
manticore-projects, line 62: Please add at least one test `assertSqlCanBeParsedAndDeparsed(...)`, e.g. line 72.

manticore-projects (changes requested):

manticore-projects:
Thank you for your work and effort, good job!

PR #2454, opened 2026-08-08
fix: support ClickHouse tuple positional access

### LINKED ISSUE #2442: [BUG] JSQLParser 5.4-SNAPSHOT : ClickHouse : Tuple positional access via .N not supported in SELECT
## Failing SQL Feature

  ClickHouse supports accessing tuple elements by their 1-based position using a `.N` suffix, such as `tuple(1, 2, 3).2` which returns `2`. The `.N` acts like a member accessor but with an integer index rather than a name. JSQLParser fails to parse this positional tuple access, treating the `.2` as a floating-point token instead of an element accessor.

  See the ClickHouse docs:

  https://clickhouse.com/docs/en/sql-reference/data-types/tuple

  ## Full SQL Example

      CREATE TABLE tuple_demo (id UInt32) ENGINE = MergeTree ORDER BY id;
      INSERT INTO tuple_demo (id) VALUES (1);
      SELECT tuple(1, 2, 3).2 FROM tuple_demo;

  ## Parsing Error

      ParseException: Encountered: <S_DOUBLE> / ".2", at line 1, column 22,
      in lexical state DEFAULT.
      Was expecting one of these terminals within expansion starting at 1753:23:
          <EOF> (inside 1753:23)
          <ST_SEMICOLON> (inside 1753:23)

Tested with ClickHouse version: 25.8

jianjindream:
Fixes #2442

## Problem

ClickHouse supports tuple positional access using `tuple(...).N`.
JSQLParser tokenizes `.2` as `S_DOUBLE`, so the existing function postfix grammar does not see a standalone `.` token and parsing fails.

## Fix

Keep the global numeric lexer unchanged. Recognize only strict `.[0-9]+` `S_DOUBLE` tokens locally in the `InternalFunction` postfix and represent the position with the existing `Function` attribute and `LongValue` AST support.

The existing `.` + `Function()` / `Column()` postfix paths remain unchanged.

## Tests

Added a parse/deparse regression test for:

```sql
SELECT tuple(1, 2, 3).2 FROM tuple_demo
```

The full Gradle test suite and check pass, and JavaCC generation introduces no new warnings.

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -9559,6 +9559,7 @@ Function InternalFunction(boolean escaped):
     Expression expr = null;
     Expression attributeExpression = null;
     Column attributeColumn = null;
+    Token attributeToken = null;
     Limit limit;
     List<Function.KeywordArgument> keywordArgs = null;
 }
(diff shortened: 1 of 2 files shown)


manticore-projects:
Thank you!

Issue #2436, opened 2026-06-14
[BUG] JSQLParser 5.4-SNAPSHOT : ClickHouse : C-style ternary operator (? :) not supported in SELECT


tomershay:
**Failing SQL Feature**

ClickHouse supports the C-style ternary conditional operator `cond ? then : else` as an alias for the `if(cond, then, else)` function. For example, `x > 0 ? 'y' : 'n'` returns `'y'` when `x > 0` and `'n'` otherwise. JSQLParser fails to parse this operator when used as a select item.

  See the ClickHouse docs:
  https://clickhouse.com/docs/en/sql-reference/functions/conditional-functions

  **SQL Example**

  ```sql
  CREATE TABLE t (x Int32) ENGINE = MergeTree ORDER BY x;

  INSERT INTO t VALUES (1);

  SELECT x > 0 ? 'y' : 'n' FROM t;
```
  Parsing error

`ParseException: Encountered: "?" / "?", at line 1, column 14, in lexical state DEFAULT.`

PR #2466, opened 2026-08-14
fix: support ClickHouse C-style ternary operator (? :) (#2436)

fudianchn:
## What

Support the ClickHouse C-style ternary conditional operator `cond ? then : else` (an alias for `if(cond, then, else)`), fixing #2436:

```sql
SELECT x > 0 ? 'y' : 'n' FROM t
SELECT * FROM t WHERE a OR b ? c : d
SELECT abs(x > 0 ? 1 : 2) FROM t
```

JSQLParser currently rejects these with `ParseException: Encountered: "?"`.

## Why / Root cause

`?` is already double-booked in the grammar: as a JDBC parameter atom (`JdbcParameter`) and as the PostgreSQL JSON existence operator (`RegularConditionRHS` consumes `expr ? 'key'`). The ternary needs the third reading: `?` **after** a complete boolean expression, closed by a `:`. Neither existing path can express that, so `x > 0 ? 'y' : 'n'` fails at the `?`.

## How

The ternary is integrated into the existing Pratt boolean loop `prattExpressionRest`, at the same precedence level as `OR` (C semantics: binds looser than `AND`/`OR`, right-associative):

- a pending `?` is routed to the ternary via a new `isTernaryAhead()` scan (same pattern as `isUnparenthesizedSelectAhead`): it is a ternary when a standalone `:` closes the then-branch at the same nesting depth before any expression boundary (`,`, `;`, `)`, EOF, clause keyword). Otherwise the `?` keeps its current meaning as the JSON operator — `isConditionSuffixAhead()`/`isComparisonOperatorAhead()` decline it only when it is a ternary.
- the `:` separator collided with the JSON path operator (`b : path` in `PrimaryExpression`/`JsonExpression`); inside a then-branch a top-level `:` now closes the ternary instead (guarded by a `ternaryThenBranchDepth` counter, `LOOKAHEAD(2, { ... })`).

AST: new `TernaryExpression(condition, thenExpression, elseExpression)` wired into `ExpressionVisitor`, `ExpressionVisitorAdapter`, `TablesNamesFinder`, `ExpressionDeParser` and `ExpressionValidator`, following the pattern of `RangeExpression`.

## Scope

Works in every expression context (select items, `WHERE`, `GROUP BY`/`HAVING`, `ORDER BY`, function arguments, join `ON`, `UPDATE SET`, `INSERT VALUES`, subqueries, nested ternaries in both branches, JDBC parameters as branches). The PostgreSQL JSON operators (`?`, `?|`, `?&`, `->`, `:` paths) and JDBC parameters (`?`, `?5`, `LIMIT ?`) keep parsing unchanged; a JSON `?` followed later by a top-level `:` at depth 0 is inherently ambiguous across dialects and stays a JSON operator only when no `:` closes it first.

## Testing

`TernaryExpressionTest` (32 tests): the issue case with AST shape assertions, 16 context round-trips (parse + deparse), precedence/associativity AST assertions (`a OR b ? c : d` groups as `(a OR b) ? c : d`, `a ? b : c ? d : e` nests right), and 11 unaffected-syntax guards for JSON operators, JSON paths, JDBC parameters and array ranges.

Local: `spotlessApply`, `checkstyleMain/Test`, `spotbugsMain`, `pmdMain` clean; full test suite green (4737 tests, 0 failures).

## Performance

`gradle jmh`, `JSQLParserBenchmark.parseSQLStatements` on `performance.sql`, `version=latest`, 10 forks × 10 iterations (100 samples) on a 32-core host:

| build | ms/op |
|---|---|
| master | `3.714 ± 0.021` |
| this change | `3.730 ± 0.023` |

The `+0.43%` delta lies within the 99.9% confidence intervals (the CIs overlap from `3.707` to `3.735`): no regression. When no `?` is present the only added cost is one string comparison per `prattExpressionRest` iteration.

`SELECT x > 0 ? 'y' : 'n' FROM t` now parses and round-trips:

```
SELECT x > 0 ? 'y' : 'n' FROM t
condition      : x > 0 (GreaterThan)
thenExpression : 'y'
elseExpression : 'n'
```

### DIFF
--- a/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitor.java
+++ b/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitor.java
@@ -737,6 +737,12 @@ public interface ExpressionVisitor<T> {
         this.visit(rangeExpression, null);
     }
 
+    <S> T visit(TernaryExpression ternaryExpression, S context);
+
+    default void visit(TernaryExpression ternaryExpression) {
+        this.visit(ternaryExpression, null);
+    }
+
     <S> T visit(TSQLLeftJoin tsqlLeftJoin, S context);
 
     default void visit(TSQLLeftJoin tsqlLeftJoin) {
--- a/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitorAdapter.java
+++ b/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitorAdapter.java
@@ -823,6 +823,12 @@ public class ExpressionVisitorAdapter<T>
     }
 
     @Override
+    public <S> T visit(TernaryExpression ternaryExpression, S context) {
+        return visitExpressions(ternaryExpression, context, ternaryExpression.getCondition(),
+                ternaryExpression.getThenExpression(), ternaryExpression.getElseExpression());
+    }
+
+    @Override
     public <S> T visit(TSQLLeftJoin tsqlLeftJoin, S context) {
         return visitBinaryExpression(tsqlLeftJoin, context);
     }
--- a/src/main/java/net/sf/jsqlparser/expression/TernaryExpression.java
+++ b/src/main/java/net/sf/jsqlparser/expression/TernaryExpression.java
@@ -1,0 +1,68 @@
+/*-
+ * #%L
+ * JSQLParser library
+ * %%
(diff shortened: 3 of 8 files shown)


manticore-projects:
Excellent work! Thank your for your contribution!

PR #2469, opened 2026-08-15
fix: support ClickHouse ORDER BY WITH FILL and INTERPOLATE (#2467)

### LINKED ISSUE #2467: [BUG] JSQLParser 5.4-SNAPSHOT : ClickHouse : ORDER BY ... WITH FILL / INTERPOLATE not supported
**Failing SQL Feature**

  ClickHouse supports the `WITH FILL` modifier on `ORDER BY` expressions, which fills in missing rows in a sorted sequence (gap filling for time series). It accepts optional `FROM`, `TO` and `STEP` bounds, where `STEP` may be an `INTERVAL` literal, and an optional query-level `INTERPOLATE` clause that computes the values of non-sorted columns for the generated rows. JSQLParser fails to parse the `WITH FILL` modifier when used in `ORDER BY`.

  See the ClickHouse docs:
  https://clickhouse.com/docs/en/sql-reference/statements/select/order-by#order-by-expr-with-fill-modifier

  **SQL Example**

```sql
CREATE TABLE t (d Date, v UInt32) ENGINE = MergeTree ORDER BY d;

INSERT INTO t VALUES ('2024-01-03', 10), ('2024-01-06', 20), ('2024-01-10', 30);

SELECT d, v FROM t ORDER BY d WITH FILL FROM toDate('2024-01-01') TO toDate('2024-02-01') STEP INTERVAL 1 DAY INTERPOLATE (v AS v + 1);
```

  **Parsing error**

ParseException: Encountered: <K_WITH> / "WITH", at line 4, column 3, in lexical state DEFAULT.

Tested with Clickhouse 25.8.

fudianchn:
## What
Support ClickHouse's `ORDER BY ... WITH FILL` modifier and the query-level `INTERPOLATE` clause, following the [ClickHouse grammar](https://clickhouse.com/docs/en/sql-reference/statements/select/order-by#order-by-expr-with-fill-modifier): per element `[WITH FILL] [FROM expr] [TO expr] [STEP expr] [STALENESS expr]` (the step may be an `INTERVAL` literal), and after the ORDER BY list an optional `INTERPOLATE [(col [AS expr], ...)]` where the bare form fills all allowed columns.

## Why
`SELECT d, v FROM t ORDER BY d WITH FILL FROM toDate('2024-01-01') TO toDate('2024-02-01') STEP INTERVAL 1 DAY INTERPOLATE (v AS v + 1)` currently fails with a ParseException on `WITH`. WITH FILL is the standard ClickHouse construct for gap filling in time series, so statements using it cannot be parsed at all.

## How
- `OrderByElement` carries a new `WithFill` holder (its presence marks the element as filled; from/to/step/staleness are optional expressions). This mirrors ClickHouse, where the bounds belong to the order by element.
- `Select` (shared by `PlainSelect`, `ParenthesedSelect` and `SetOperationList`) carries the `List<InterpolateElement>` as a sibling of `orderByElements`, mirroring ClickHouse, where INTERPOLATE is a query-level attribute. The existing union hoisting for `orderByElements` is extended so `SELECT ... UNION SELECT ... ORDER BY x WITH FILL INTERPOLATE (...)` lands on the `SetOperationList`.
- Rendering in both `toString()` and the deparser (bounds rendered through the expression visitor), and `OrderByValidator`/`SelectValidator` walk the new expressions.
- `FILL`, `STEP`, `STALENESS` and `INTERPOLATE` are declared as non-reserved keywords (inside the `NonReservedWord` range), so they remain usable as plain column names (`SELECT fill, step FROM t ORDER BY interpolate` still parses).

## Root cause
The grammar had no production for the WITH FILL modifier family and no INTERPOLATE clause; the element-level bounds and the query-level column list are modeled after the ClickHouse reference parser (bounds only parsed after `WITH FILL`, in fixed FROM/TO/STEP/STALENESS order).

## Testing
- `ClickHouseTest`: 9 new tests covering the issue statement, bare `WITH FILL`, FROM/TO/STEP, `STEP INTERVAL` plus `STALENESS`, DESC/NULLS LAST combinations, INTERPOLATE variants (multiple items, item without AS, bare `INTERPOLATE`), a UNION with hoisting, a subquery, and column names using the new keywords.
- `./gradlew test`: 4793 tests, 0 failures (26 skipped, pre-existing).

## Performance
gradle jmh, `JSQLParserBenchmark.parseSQLStatements` on `performance.sql`, version=latest, 10 forks × 10 iterations (100 samples) on a 32-core host:

| build | ms/op |
|---|---|
| master | 3.689 ± 0.019 |
| branch | 3.690 ± 0.020 |

The difference is within the confidence intervals: no regression.

## Notes
- `compileJavacc` emits one additional choice-conflict warning (12 → 13) caused by the INTERPOLATE optional in `PlainSelect`'s clause tail; the token has exactly one valid parse at that position, and the repo already carries such warnings (tracked in #2403).
- `INTERPOLATE ()` with an empty list is accepted and normalized to the bare `INTERPOLATE` form (both fill all allowed columns in ClickHouse).

Fixes #2467


manticore-projects:
Thank you much!

PR #2471, opened 2026-08-15
feat: support CREATE DATABASE (#2070)

### LINKED ISSUE #2070: How to parse CREATE DATABASE DATABASE_NAME  ?
package org.example;

import net.sf.jsqlparser.parser.CCJSqlParser;
import net.sf.jsqlparser.parser.CCJSqlParserUtil;
import net.sf.jsqlparser.statement.Statement;
import net.sf.jsqlparser.statement.Statements;

class Main {

        try {
            String sqlStr = "CREATE DATABASE USERS; ";
            Statement statement=CCJSqlParserUtil.parse(sqlStr);

        } catch (Exception e) {
            System.out.println(e.getMessage());
        }

    }
}

fudianchn:
## What
Adds a typed `CreateDatabase` statement for `CREATE DATABASE [IF NOT EXISTS] <name> [options]`, which previously fell through to the generic `UnsupportedStatement` capture.

## Why
The grammar has no production for `CREATE DATABASE` (#2070), so users cannot access the database name from the AST. `CREATE SCHEMA` already has a typed statement and MySQL treats `CREATE SCHEMA` as a synonym of `CREATE DATABASE`.

## How
- New `CreateDatabase()` production dispatched from `Create()` right after `CreateSchema()`, following the same shape: `LOOKAHEAD(2)`-guarded `IF NOT EXISTS` and `RelObjectName()` for the name.
- Vendor-specific options after the name are captured verbatim via the existing `captureRest()` helper (same approach as `UnsupportedStatement` and `CreateFunctionalStatement`), so `CREATE DATABASE mydb DEFAULT CHARACTER SET utf8mb4` keeps parsing and round-trips unchanged instead of failing.
- AST class `net.sf.jsqlparser.statement.create.database.CreateDatabase` plus wiring in `StatementVisitor`, `StatementVisitorAdapter`, `StatementDeParser`, `StatementValidator` (new `Feature.createDatabase` validated against `NamedObject.database`) and `TablesNamesFinder`.

## Root cause
Missing feature: `CREATE DATABASE` had no grammar production and always hit the `captureRest()` fallback.

## Testing
- New `CreateDatabaseTest` (5 cases): the issue's SQL, `IF NOT EXISTS`, quoted names, options capture with AST assertions; `./gradlew test --tests ...CreateDatabaseTest` passes.
- Full local suite on the branch: 4801 tests, 0 failures.

Before (master `406a4d4`): `CCJSqlParserUtil.parse("CREATE DATABASE USERS")` returns `UnsupportedStatement` with no access to the database name.
After: returns `CreateDatabase` with `getDatabaseName() = "USERS"` and the deparse round-trips; same for `CREATE DATABASE IF NOT EXISTS mydb DEFAULT CHARACTER SET utf8mb4`.

Grammar change, paired `gradle jmh` run (`JSQLParserBenchmark.parseSQLStatements` on `performance.sql`, `version=latest`, 10 forks x 10 iterations = 100 samples, 32-core host):

| build | ms/op |
|---|---|
| master `406a4d4` | 3.734 ± 0.030 |
| this PR | 3.725 ± 0.028 |

Delta within the confidence intervals -> no regression.

*Limitation:* options are kept as raw tokens rather than structured AST nodes (the option sets differ heavily between MySQL and PostgreSQL); they can be modelled individually in a follow-up if preferred. The bare form `CREATE DATABASE` (no name) now throws a `ParseException` instead of returning `UnsupportedStatement`; both MySQL and PostgreSQL require the name.

Fixes #2070

### DIFF
--- a/src/main/java/net/sf/jsqlparser/parser/feature/Feature.java
+++ b/src/main/java/net/sf/jsqlparser/parser/feature/Feature.java
@@ -30,6 +30,7 @@ import net.sf.jsqlparser.statement.alter.Alter;
 import net.sf.jsqlparser.statement.alter.sequence.AlterSequence;
 import net.sf.jsqlparser.statement.analyze.Analyze;
 import net.sf.jsqlparser.statement.comment.Comment;
+import net.sf.jsqlparser.statement.create.database.CreateDatabase;
 import net.sf.jsqlparser.statement.create.function.CreateFunction;
 import net.sf.jsqlparser.statement.create.index.CreateIndex;
 import net.sf.jsqlparser.statement.create.procedure.CreateProcedure;
@@ -475,6 +476,12 @@ public enum Feature {
      */
     drop, dropTable, dropIndex, dropView, dropSchema, dropSequence, dropTableIfExists, dropIndexIfExists, dropViewIfExists, dropSchemaIfExists, dropSequenceIfExists,
 
+    /**
+     * SQL "CREATE DATABASE" statement is allowed
+     *
+     * @see CreateDatabase
+     */
+    createDatabase,
     /**
      * SQL "CREATE SCHEMA" statement is allowed
      *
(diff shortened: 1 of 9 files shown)


manticore-projects:
Thank you much!

PR #2453, opened 2026-08-07
feat: support MERGE ... WHEN NOT MATCHED BY TARGET / BY SOURCE (#2421)

fudianchn:
## Problem

BigQuery and SQL Server allow the `MERGE ... WHEN NOT MATCHED` clause to be qualified with `BY TARGET` (default, allows `INSERT`) or `BY SOURCE` (allows `UPDATE`/`DELETE`). JSQLParser currently rejects it as an `UnsupportedStatement` (#2421):

```sql
MERGE INTO target_table AS tt
USING (SELECT key, field FROM source_table) AS st ON tt.key = st.key
WHEN NOT MATCHED BY TARGET THEN INSERT (key, field) VALUES (st.key, st.field)
WHEN NOT MATCHED BY SOURCE THEN DELETE;
```

The grammar's `MergeWhenNotMatched` had no `BY` branch and was hard-wired to `INSERT`.

## Change

- **Grammar** (`JSqlParserCC.jjt`): add an optional `[BY TARGET|SOURCE]` to `WHEN NOT MATCHED`, and let that branch produce `INSERT`, `UPDATE`, or `DELETE` (extracted `MergeInsertClause` helper so all three branches stay simple, mirroring `MergeWhenMatched`). Standard SQL `WHEN [NOT] MATCHED` (no `BY`) is unchanged.
- **Model**: new `MergeSide { TARGET, SOURCE }` enum + a `side` field on `MergeInsert`/`MergeUpdate`/`MergeDelete`, rendered by both `toString()` and the `MergeDeParser` visitor.
- `TARGET`/`SOURCE` are matched as identifiers (`S_IDENTIFIER`) rather than reserved keywords, so existing SQL that uses them as column/table names is unaffected.

## Verification

- New tests in `MergeTest` cover `BY TARGET` (INSERT) + `BY SOURCE` (DELETE), and `BY SOURCE ... THEN UPDATE` with an `AND` predicate — each with a round-trip `assertSqlCanBeParsedAndDeparsed` plus AST assertions on the `side` field.
- Full suite: **4659 tests, 0 failures**. `checkstyle` / `pmd` / `license:check-file-header` clean.
- Backward compatible: existing `WHEN MATCHED ... UPDATE/DELETE` and `WHEN NOT MATCHED ... INSERT` tests keep passing (23 pre-existing Merge tests unaffected).

Fixes #2421.

Issue #2421, opened 2026-03-27
[FEATURE] BigQuery statement MERGE ... WHEN NOT MATCHED BY TARGET


il-bert:
### Grammar or Syntax Description
- BigQuery in the "merge ... when not matched " allows to define when it's not matched **by target** or **by source**
currently jsqlparser is failing with these kind of statemnt stating UnsupportedStatement

https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/dml-syntax#merge_statement

```
when_clause ::= matched_clause | not_matched_by_target_clause | not_matched_by_source_clause
matched_clause ::= WHEN MATCHED [ AND search_condition ] THEN { merge_update_clause | merge_delete_clause }
not_matched_by_target_clause ::= WHEN NOT MATCHED [BY TARGET] [ AND search_condition ] THEN merge_insert_clause
not_matched_by_source_clause ::= WHEN NOT MATCHED BY SOURCE [ AND search_condition ] THEN { merge_update_clause | merge_delete_clause }
```

### SQL Example

```
MERGE INTO target_table AS tt
USING ( SELECT key, field FROM source_table) AS st
ON tt.key = st.key
WHEN NOT MATCHED BY TARGET THEN
    INSERT (key, field)
    VALUES (st.key, st.field)
WHEN NOT MATCHED BY SOURCE THEN DELETE
;
```

can it be included?

thanks

PR #2481, opened 2026-08-17
fix: reject illegal MERGE WHEN NOT MATCHED clause pairings (#2480)

### LINKED ISSUE #2480: [BUG] MERGE silently rewrites illegal WHEN NOT MATCHED clause pairings into the opposite semantics
### Failing SQL Feature:

`MERGE` accepts illegal `WHEN NOT MATCHED` side / clause pairings and silently rewrites them into semantically different, legal SQL. SQL Server and BigQuery (the dialects providing `BY TARGET` / `BY SOURCE`) reject these pairings as syntax errors.

Full behavior matrix on master `f41c0b8`:

| Clause pairing | Legal? | Result on master |
|---|---|---|
| `WHEN MATCHED THEN UPDATE/DELETE` | yes | accepted, round-trips |
| `WHEN MATCHED THEN INSERT` | no | rejected |
| `WHEN NOT MATCHED THEN INSERT` | yes | accepted, round-trips |
| `WHEN NOT MATCHED THEN UPDATE/DELETE` | no | **accepted, deparses as `WHEN MATCHED`** (condition inverted) |
| `WHEN NOT MATCHED BY TARGET THEN INSERT` | yes | accepted, round-trips |
| `WHEN NOT MATCHED BY TARGET THEN UPDATE/DELETE` | no | **accepted, deparses as `WHEN MATCHED`** (condition inverted) |
| `WHEN NOT MATCHED BY SOURCE THEN UPDATE/DELETE` | yes | accepted, round-trips |
| `WHEN NOT MATCHED BY SOURCE THEN INSERT` | no | **accepted, deparses as `WHEN NOT MATCHED`** (`BY SOURCE` dropped, different row set) |

The parser therefore silently changes the meaning of the statement instead of failing, which is the worst failure mode for auditing / firewall tools built on the parser.

Root cause: `MergeWhenNotMatched` (`JSqlParserCC.jjt:4472`) accepts any clause for any side, while the AST can only represent the legal pairings: `MergeUpdate:82` / `MergeDelete:55` render `WHEN MATCHED` for any side other than `SOURCE`, and `MergeInsert:77` always renders `WHEN NOT MATCHED`. Introduced with the `BY TARGET` / `BY SOURCE` support from #2421 / #2453, which validated the six legal combinations but not the illegal ones.

### SQL Example:

```java
// accepted on master, silently rewritten:
CCJSqlParserUtil.parse("MERGE INTO t USING s ON t.id = s.id"
        + " WHEN NOT MATCHED BY TARGET THEN UPDATE SET a = 1").toString()
// => MERGE INTO t USING s ON t.id = s.id WHEN MATCHED THEN UPDATE SET a = 1

CCJSqlParserUtil.parse("MERGE INTO t USING s ON t.id = s.id"
        + " WHEN NOT MATCHED BY SOURCE THEN INSERT (a) VALUES (1)").toString()
// => MERGE INTO t USING s ON t.id = s.id WHEN NOT MATCHED THEN INSERT (a) VALUES (1)
```

### Software Information:

- JSqlParser version: 5.4-SNAPSHOT (master f41c0b8)
- Database: SQL Server / BigQuery (`WHEN NOT MATCHED [BY TARGET|BY SOURCE]` dialects)

fudianchn:
## What

Rejects the five illegal `WHEN NOT MATCHED` side / clause pairings at parse time instead of silently rewriting them: `WHEN NOT MATCHED [BY TARGET] THEN UPDATE/DELETE` and `WHEN NOT MATCHED BY SOURCE THEN INSERT` now throw, matching what SQL Server and BigQuery do. Fixes #2480.

## Why

On master these pairings are accepted and the deparse collapses them into the opposite semantics: `WHEN NOT MATCHED [BY TARGET] THEN UPDATE/DELETE` prints as `WHEN MATCHED` (condition inverted) and `WHEN NOT MATCHED BY SOURCE THEN INSERT` prints as `WHEN NOT MATCHED` (`BY SOURCE` dropped, a different row set). A silent semantic rewrite is the worst failure mode for tools that parse SQL to inspect it. This closes the gap left by #2453, which calibrated the six legal combinations but did not guard the illegal ones.

## How

`MergeWhenNotMatched` ends with `MergeSide.validatePairing(side, operation)`, which enforces the pairing rule already documented on `MergeSide`: `BY TARGET` (the default) only allows `INSERT`, `BY SOURCE` only allows `UPDATE`/`DELETE`. The validation lives in `MergeSide` next to the existing `fromImage` validation and keeps the grammar action a single statement. `WHEN MATCHED THEN INSERT` is already rejected by the grammar and is pinned by a test as well.

Alternative considered (not implemented): accept the illegal pairings and render them faithfully. This would require extending `MergeUpdate` / `MergeDelete` / `MergeInsert` with combinations no dialect accepts (`MergeUpdate` currently has no `NOT MATCHED` dimension), adding AST surface for SQL that no database executes. Happy to switch to that direction or to move the check into the statement validator instead, if preferred.

## Root cause

`MergeWhenNotMatched` accepted any clause for any side, while the AST can only represent the legal pairings (`MergeUpdate:82` / `MergeDelete:55` render `WHEN MATCHED` for any side other than `SOURCE`, `MergeInsert:77` always renders `WHEN NOT MATCHED`), so the deparse silently changed the meaning.

## Testing

- New `MergeTest#testMergeRejectsInvalidWhenNotMatchedClausePairings`: on master the five illegal pairings parse (test fails), with this change they throw `JSQLParserException` with the pairing message; `./gradlew test --tests net.sf.jsqlparser.statement.merge.MergeTest` -> 26 tests, 0 failures.
- The six legal combinations keep round-tripping (existing `#2421` tests, untouched).

| Case from #2480 | master `f41c0b8` | with this change |
|---|---|---|
| `WHEN NOT MATCHED [BY TARGET] THEN UPDATE/DELETE` | accepted, deparses as `WHEN MATCHED` | rejected with `WHEN NOT MATCHED [BY TARGET] cannot take an UPDATE or DELETE clause` |
| `WHEN NOT MATCHED BY SOURCE THEN INSERT` | accepted, deparses as `WHEN NOT MATCHED` | rejected with `WHEN NOT MATCHED BY SOURCE cannot take an INSERT clause` |
| all six legal pairings | accepted, round-trip | unchanged |

Grammar tokens, productions and LOOKAHEADs are untouched (javacc warnings stay at 13) and the benchmark input `performance.sql` contains no `MERGE` statement, so the changed path never executes in `gradle jmh`; no performance impact is possible.

Programmatic AST construction (building an illegal `MergeUpdate` by hand and printing it) is not guarded; the check protects the parse entry point.

Fixes #2480

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/merge/MergeSide.java
+++ b/src/main/java/net/sf/jsqlparser/statement/merge/MergeSide.java
@@ -29,6 +29,29 @@ public enum MergeSide {
      * @throws IllegalArgumentException if {@code image} is neither {@code TARGET} nor
      *         {@code SOURCE}
      */
+    /**
+     * Validates the clause pairing of a {@code WHEN NOT MATCHED} branch: {@code BY TARGET} (the
+     * default) only allows an {@code INSERT} clause, {@code BY SOURCE} only allows {@code UPDATE}
+     * or {@code DELETE}.
+     *
+     * @param side the parsed {@code BY TARGET}/{@code BY SOURCE} qualifier, null when absent
+     * @param operation the parsed clause
+     * @return the passed {@code operation}
+     * @throws IllegalArgumentException when the pairing is not legal in any dialect
+     */
+    public static MergeOperation validatePairing(MergeSide side, MergeOperation operation) {
+        if (side == MergeSide.SOURCE) {
+            if (operation instanceof MergeInsert) {
+                throw new IllegalArgumentException(
+                        "WHEN NOT MATCHED BY SOURCE cannot take an INSERT clause");
+            }
+        } else if (!(operation instanceof MergeInsert)) {
+            throw new IllegalArgumentException(
+                    "WHEN NOT MATCHED [BY TARGET] cannot take an UPDATE or DELETE clause");
+        }
+        return operation;
+    }
+
     public static MergeSide fromImage(String image) {
         for (MergeSide value : values()) {
             if (value.name().equalsIgnoreCase(image)) {
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -4489,7 +4489,7 @@ MergeOperation MergeWhenNotMatched() : {
         |
         operation = MergeDeleteClause(predicate) { ((MergeDelete) operation).setSide(side); }
     )
-    { return operation; }
+    { return MergeSide.validatePairing(side, operation); }
 }
 
 // table names seem to allow ":" delimiters, e.g. for Informix see #1134
--- a/src/test/java/net/sf/jsqlparser/statement/merge/MergeTest.java
+++ b/src/test/java/net/sf/jsqlparser/statement/merge/MergeTest.java
@@ -25,6 +25,7 @@ import java.util.stream.Stream;
 import static net.sf.jsqlparser.test.TestUtils.assertOracleHintExists;
 import static net.sf.jsqlparser.test.TestUtils.assertSqlCanBeParsedAndDeparsed;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.Mockito.mock;
 
 /**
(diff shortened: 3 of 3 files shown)


fudianchn:
The only failing check is Gradle Check (macos-latest), which failed in its Set up job step while downloading the actions/setup-java action (429 Too Many Requests from codeload.github.com): a transient GitHub infrastructure issue unrelated to this change. Gradle Check on ubuntu/windows and Codacy pass for the same commit; Maven Verify was skipped because of the failed sibling job. I cannot re-run the failed job on a fork PR, please re-run it when convenient.

manticore-projects:
Thanks!

PR #2482, opened 2026-08-18
feat(parser): add support for ClickHouse [LEFT] ARRAY JOIN syntax

### LINKED ISSUE #2468: [BUG] JSQLParser 5.4-SNAPSHOT : ClickHouse : ARRAY JOIN / LEFT ARRAY JOIN not supported in SELECT
**Failing SQL Feature**

  ClickHouse supports `ARRAY JOIN` and `LEFT ARRAY JOIN`, which unnest an array-typed column of a table in the `FROM` clause into one row per element (comparable to `LATERAL UNNEST`). Unlike a regular join it has no right-hand table and no `ON`/`USING` condition, its operand is an array expression with an optional alias. `LEFT ARRAY JOIN` additionally keeps rows whose array is empty, emitting the element type's default value. JSQLParser fails to parse both forms.
  
  See the ClickHouse docs:
  https://clickhouse.com/docs/en/sql-reference/statements/select/array-join

  **SQL Example**
  
  ```sql
  CREATE TABLE t (s String, arr Array(UInt32)) ENGINE = MergeTree ORDER BY s;

  INSERT INTO t VALUES ('a', [1, 2]), ('b', []), ('c', [3]);

  SELECT s, x FROM t LEFT ARRAY JOIN arr AS x;
  ```
  
  Parsing error

  `ParseException: Encountered: <K_ARRAY_LITERAL> / "ARRAY", at line 1, column 25, in lexical state DEFAULT.`

Tested with Clickhouse 25.8.

watashi-00:
### Summary
Implements support for ClickHouse `ARRAY JOIN` and `LEFT ARRAY JOIN` clauses in `SELECT` statements, enabling unnesting array columns into separate rows.

Fixes #2468

### Changes
- **Grammar (`JSqlParserCC.jjt`)**: Added grammar support for `[LEFT] ARRAY JOIN` and adjusted `isAliasAhead()` / `isReservedKeywordSafeByFollower()` to prevent `LEFT` / `ARRAY` from being wrongly recognized as table aliases.
- **AST (`Join.java`)**: Added `isArray()` / `setArray()` and updated join formatting.
- **Tests (`SelectTest.java`)**: Added `testClickHouseArrayJoin()` with parse/deparse assertions and AST validation.
- **Documentation**: Updated `README.md` under supported extensions.

### Checklist
- [x] `assertSqlCanBeParsedAndDeparsed()` executed for new syntax
- [x] Full test suite passing (0 failures)
- [x] `mvn checkstyle:check` passed with 0 violations
- [x] Clean commit history rebased on `upstream/master`

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -829,9 +829,12 @@ public class CCJSqlParser extends AbstractJSqlParser<CCJSqlParser> {
             case K_CONNECT:  return nextKind != K_BY;
             case K_START:    return nextKind != K_WITH;
             case K_LEFT:     return nextKind != K_JOIN && nextKind != K_OUTER
-                                    && nextKind != K_SEMI;
+                                    && nextKind != K_SEMI && nextKind != K_ARRAY_LITERAL
+                                    && (next2.image == null || !next2.image.equalsIgnoreCase("ARRAY"));
+
             case K_RIGHT:    return nextKind != K_JOIN && nextKind != K_OUTER
-                                    && nextKind != K_SEMI;
+                                    && nextKind != K_SEMI && nextKind != K_ARRAY_LITERAL
+                                    && (next2.image == null || !next2.image.equalsIgnoreCase("ARRAY"));
             case K_ALL:      return nextKind != K_JOIN;
             case K_ANY:      return nextKind != OPENING_BRACKET;
             case K_SOME:     return nextKind != OPENING_BRACKET;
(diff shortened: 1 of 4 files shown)


src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
```diff
@@ -993,29 +995,17 @@ public class CCJSqlParser extends AbstractJSqlParser<CCJSqlParser> {
         // String-literal alias:  SELECT col 'myAlias'
         if (kind == S_CHAR_LITERAL) return true;
 
-        // Base identifier tokens
```
manticore-projects, line 996: Why changing the logical order here?

manticore-projects: also, run at least `gradle check` before committing.

manticore-projects (approved):

manticore-projects:

src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
watashi-00: Thanks for the tip...

about the order, since K_ARRAY_LITERAL is inside the non-reserved word range, having the range check first was eagerly eating ARRAY as a table alias for t (in FROM t ARRAY JOIN ...) before the switch even had a chance to check getToken(2)
Moving the switch up catches combinations like ARRAY JOIN and LEFT ARRAY early, while still falling back to the range check when ARRAY is actually just an alias

PR #2479, opened 2026-08-17
fix: complete TablesNamesFinder traversal for piped queries, DML side clauses and analytic functions (#2478)

fudianchn:
## What

Completes the visitor traversal of `TablesNamesFinder` for piped queries (`FromQuery` + all pipe operators), the `WITH` list of `DELETE`, the `ON` condition and `WHEN` operations of `MERGE`, the `SET` list, `ON DUPLICATE KEY UPDATE` / `ON CONFLICT ... DO UPDATE` actions and `OUTPUT` / `RETURNING` clauses of the DML statements, data modifying CTE payloads and the `ORDER BY` / `FILTER` parts of analytic functions.

## Why

`TablesNamesFinder` is the documented entry point for SQL auditing / firewall use cases; four of the gaps silently return an incomplete table list and two throw. Fixes #2478.

## How

- `visit(FromQuery)` traverses the `WITH` list, the from item and all pipe operators; `TablesNamesFinder` now implements `PipeOperatorVisitor` (mirroring `SelectDeParser`).
- `visit(Delete)` traverses `withItemsList` like `visit(Update)` / `visit(Insert)` already do.
- `visit(Merge)` traverses `onCondition` and all operations; `TablesNamesFinder` now implements `MergeOperationVisitor`.
- `visit(Insert)` traverses `setUpdateSets`, `duplicateAction`, `conflictAction` (via the existing `ExpressionVisitor.visitUpdateSets` default); all three DML visits traverse `outputClause` (overriding the `SelectVisitor` default) and `returningClause`.
- `visit(WithItem)` dispatches the payload as any `ParenthesedStatement` instead of casting via `getSelect()`, which fixes data modifying CTEs.
- `visit(AnalyticExpression)` traverses `funcOrderBy` and `orderByElements` independently and adds `filterExpression`.

The parser, grammar and deparser are untouched, so there is no impact on parsing performance.

## Root cause

Visitor coverage in `TablesNamesFinder` was incomplete: newer AST families (piped queries, merge operations, insert actions, output / returning clauses) were never wired up, and two pre-existing guards were wrong (`getSelect()` cast, `getFuncOrderBy()` checked but `getOrderByElements()` iterated).

## Testing

- 14 new tests in `TablesNamesFinderTest`; all 14 fail on master `f41c0b8` (missing tables, `ClassCastException`, `NullPointerException`) and pass with this change: `./gradlew test --tests net.sf.jsqlparser.util.TablesNamesFinderTest` -> 85 tests, 0 failures.
[...]

Issue #1514, opened 2022-04-19
Major Issue in 4.4, FROM Keyword is not Mandatory


bala28:
**Describe the bug**
FROM Keyword is removed from SQL and JSQLParser is not complaining about syntax.

**To Reproduce**
1. sqlStatement = "SELECT SYSDATE  DUAL"
2. Parsing this SQL using JSqlParser with this statements (  final Statement statement =  CCJSqlParserUtil.parse(sqlStatement);)
3. No Exception

**Expected behavior**
Exception must be thrown for invalid syntax as FROM is missing.

**System**
 - Oracle 
- Java 17
- JSqlParser 4.4

Note:: "FROM" keyword validation is working in 4.3

manticore-projects:
Greetings.
 
Your statement matches `SELECT expression alias` and is valid Standard SQL (although not Oracle compliant).

manticore-projects:
> thanks @manticore-projects for reply.
> 
>     1. is this new feature in 4.4 ? if not , why it is working in 4.3 ?

Yes, prior 4.4 JSQLParser supported only `SELECT SimpleExpression`, which has been improved now into `SELECT Expression`.

> 
>     2. Secondly , i tried with validation framework , but the validation errors is 0 , do i miss something below ?
> 
> 
> ```
>  Validation validation = new Validation(Arrays.asList(DatabaseType.ORACLE), "SELECT SYSDATE  DUAL WHERE 1=1");
>   List<ValidationError> errors = validation.validate();
> ```

I can not say much about the Validation Framework, because I do not maintain or use it.
Although I do know that it has holes and did not fully keep up with the latest JSQLParser developments.

So if you want to use it, then rather understand it as a template, which you may want to add to in order to validate your particular concerns.

> 
> from below article , i can see some databases support "SELECT without FROM"
> 
> https://modern-sql.com/use-case/select-without-from ,
> 
> In that case , my suggestion will be adding database type during parse in `CCJSqlParserUtil.parse(sqlStatement);`

Your consideration is correct and in fact I have had similar thoughts in the past.
However:

a) JSQLParser's main objective is to parse RDBMS agnostic and to retrieve the AST from any SQL
b) We would have to maintain a Grammar for each dialect, while JavaCC does not support any `include` directive. So we would need to engage a template-engine for the Grammars -- a lot of work not justified by any added value

Your provided statement is parsed correctly.
[...]

PR #2484, opened 2026-08-18
fix: keep all clauses when deparsing SELECT without FROM (#2483)

### LINKED ISSUE #2483: [BUG] JSQLParser 5.4-SNAPSHOT : RDBMS : SELECT without FROM silently drops GROUP BY / HAVING / WINDOW / QUALIFY / CONNECT BY / PREFERRING in toString()
### Failing SQL Feature:
`SELECT` without a `FROM` clause parses fine (supported Standard SQL, see #1514), but `PlainSelect.toString()` silently drops every clause after the select items except `WHERE` / `PREWHERE`: `GROUP BY`, `HAVING`, `WINDOW`, `QUALIFY`, `START WITH` / `CONNECT BY` and `PREFERRING` are all stored on the AST yet disappear from the output. A parse -> toString -> reparse round trip silently rewrites the statement.

The `StatementDeParser` path already renders all of them (`SelectDeParser` prints the whole clause chain outside its `getFromItem() != null` block), so the two render paths disagree.

Root cause: `PlainSelect.appendSelectBodyTo` (`src/main/java/net/sf/jsqlparser/statement/select/PlainSelect.java`): the shared clause chain lives inside the `fromItem != null` branch, while the `else // without from` branch prints `preWhere` / `where` only.

### SQL Example:
```sql
SELECT 1 GROUP BY 1 HAVING 1 = 1
SELECT 1 QUALIFY 1 = 1
SELECT 1 WINDOW w AS (ORDER BY 1)
SELECT 1 START WITH 1 = 1 CONNECT BY LEVEL <= 1
SELECT 1 PREFERRING HIGH 1
```

All five parse on current master (`ad69ecc`), and each `toString()` returns `SELECT 1`:

```
IN : SELECT 1 GROUP BY 1 HAVING 1=1
OUT: SELECT 1

IN : SELECT 1 WINDOW w AS (ORDER BY 1)
OUT: SELECT 1

IN : SELECT 1 START WITH 1 = 1 CONNECT BY LEVEL <= 1
OUT: SELECT 1
```

`WHERE` survives (`SELECT 1 WHERE 1 = 1` renders correctly), which shows the asymmetry.

### Software Information:
- JSqlParser 5.4-SNAPSHOT, master `ad69ecc`
- JDK 17
- any RDBMS (renderer-level, dialect independent)

fudianchn:
## What
`PlainSelect.toString()` silently dropped every clause except `WHERE` / `PREWHERE` for a `SELECT` without `FROM`: `GROUP BY`, `HAVING`, `WINDOW`, `QUALIFY`, `START WITH` / `CONNECT BY` and `PREFERRING` parsed fine and were stored on the AST, but disappeared from the rendered SQL.

## Why
FROM-less SELECT is supported Standard SQL (#1514). A parse -> toString -> reparse round trip silently rewrites the statement, the worst failure mode for audit / rewriting tools. The `StatementDeParser` path already renders all these clauses unconditionally, so the two render paths disagreed.

## How
`PlainSelect.appendSelectBodyTo`: keep only the FROM-adjacent parts (`FROM`, lateral views, joins, `FINAL`, ksql `WINDOW`) inside the `fromItem != null` branch, move the shared clause chain out, and delete the truncated `else // without from` branch. This mirrors `SelectDeParser`, which already prints the whole chain outside its FROM block. Output for SELECTs with FROM is byte-identical (same checks in the same order).

## Root cause
The `else // without from` branch of `PlainSelect.appendSelectBodyTo` rendered `preWhere` / `where` only, while the whole clause chain sat inside the `fromItem != null` branch.

## Testing
- New regression tests `SelectTest#testSelectWithoutFromRetainsWhereGroupByHaving`, `#testSelectWithoutFromRetainsWindowAndQualify`, `#testSelectWithoutFromRetainsHierarchicalAndPreferring`: on master the round-trip assertions fail (`toString()` returns `SELECT 1`), with the fix all pass.
- `./gradlew spotlessApply check`: 4858 tests, 0 failures, 0 errors.
- jmh not applicable: deparse-side change only, the parse path and grammar are untouched (same reasoning as #2479).

On master `ad69ecc` (jshell, `CCJSqlParserUtil.parse(...).toString()`):

```
IN : SELECT 1 GROUP BY 1 HAVING 1=1
OUT: SELECT 1

IN : SELECT 1 WINDOW w AS (ORDER BY 1)
OUT: SELECT 1

IN : SELECT 1 START WITH 1 = 1 CONNECT BY LEVEL <= 1
OUT: SELECT 1
```

With the fix, all statements from the issue round-trip through both `toString()` and `StatementDeParser` (both are asserted by the new tests).

Fixes #2483

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/select/PlainSelect.java
+++ b/src/main/java/net/sf/jsqlparser/statement/select/PlainSelect.java
@@ -607,43 +607,35 @@ public class PlainSelect extends Select {
             if (ksqlWindow != null) {
                 builder.append(" WINDOW ").append(ksqlWindow);
             }
-            if (preWhere != null) {
-                builder.append(" PREWHERE ").append(preWhere);
-            }
-            if (where != null) {
-                builder.append(" WHERE ").append(where);
-            }
-            if (oracleHierarchical != null) {
-                builder.append(oracleHierarchical);
-            }
-            if (preferringClause != null) {
-                builder.append(" ").append(preferringClause);
-            }
-            if (groupBy != null) {
-                builder.append(" ").append(groupBy);
-            }
-            if (having != null) {
-                builder.append(" HAVING ").append(having);
-            }
-            if (qualify != null) {
-                builder.append(" QUALIFY ").append(qualify);
-            }
-            if (windowDefinitions != null) {
-                builder.append(" WINDOW ");
-                builder.append(windowDefinitions.stream().map(WindowDefinition::toString)
-                        .collect(joining(", ")));
-            }
-            if (emitChanges) {
-                builder.append(" EMIT CHANGES");
-            }
-        } else {
-            // without from
-            if (preWhere != null) {
-                builder.append(" PREWHERE ").append(preWhere);
-            }
-            if (where != null) {
-                builder.append(" WHERE ").append(where);
-            }
+        }
+        if (preWhere != null) {
+            builder.append(" PREWHERE ").append(preWhere);
+        }
+        if (where != null) {
+            builder.append(" WHERE ").append(where);
+        }
+        if (oracleHierarchical != null) {
+            builder.append(oracleHierarchical);
+        }
+        if (preferringClause != null) {
+            builder.append(" ").append(preferringClause);
+        }
+        if (groupBy != null) {
+            builder.append(" ").append(groupBy);
+        }
+        if (having != null) {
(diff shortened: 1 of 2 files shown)


manticore-projects:
Thank you, all of this came from a time where `SELECT * FROM DUAL` was the standard and nothing else.

PR #2486, opened 2026-08-18
feat: complete SQL Server OPTION query hint support

fudianchn:
## Description

Completes two SQL Server `OPTION` follow-ups from #2472:

1. Parse, retain, validate and deparse `OPTION (...)` after `MERGE`, including after `OUTPUT`.
2. Parse `OPTIMIZE FOR (@variable UNKNOWN)` as an existing table-less `Column`, avoiding a new expression type and Visitor methods.

Also reject `OPTION` after standalone `VALUES`, including `INSERT ... VALUES`, because SQL Server permits query hints on `INSERT` only in its nested `SELECT`. `INSERT ... SELECT ... OPTION` remains supported.

## Testing

`OptionClauseTest`: 19/19 passed. New cases cover `MERGE`, `OUTPUT`, `OPTIMIZE FOR`, invalid `INSERT ... VALUES` and valid `INSERT ... SELECT` forms.

## Performance

`JSQLParserBenchmark.parseSQLStatements` on `performance.sql`, `version=latest`, JDK 17.0.20, 10 forks x (3 warmup + 10 measurement iterations), 100 samples per run, two interleaved runs per build on a 32-thread Intel Core i9-13900KS host:

| build | run 1 (ms/op) | run 2 (ms/op) |
|---|---:|---:|
| base `7013909` | 3.724 ± 0.026 | 3.717 ± 0.023 |
| this PR `bf17921` | 3.714 ± 0.022 | 3.738 ± 0.025 |

Mean delta: +0.15%. All confidence intervals overlap -> no regression.

### DIFF
--- a/src/test/java/net/sf/jsqlparser/statement/select/OptionClauseTest.java
+++ b/src/test/java/net/sf/jsqlparser/statement/select/OptionClauseTest.java
@@ -13,7 +13,10 @@ import static net.sf.jsqlparser.test.TestUtils.assertSqlCanBeParsedAndDeparsed;
 
 import net.sf.jsqlparser.JSQLParserException;
 import net.sf.jsqlparser.parser.CCJSqlParserUtil;
+import net.sf.jsqlparser.schema.Column;
 import net.sf.jsqlparser.statement.Statement;
+import net.sf.jsqlparser.statement.insert.Insert;
+import net.sf.jsqlparser.statement.merge.Merge;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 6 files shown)


src/main/java/net/sf/jsqlparser/expression/UnknownVariable.java
```diff
@@ -0,0 +1,54 @@
...
+
+import net.sf.jsqlparser.parser.ASTNodeAccessImpl;
+
+/**
+ * A user variable followed by the bare {@code UNKNOWN} keyword, as used in the SQL Server (T-SQL)
+ * {@code OPTIMIZE FOR (@variable_name UNKNOWN)} query hint parameter form, see
+ * <a href="https://learn.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-query">Hints
+ * (Transact-SQL) - Query Hints</a>.
+ */
+public class UnknownVariable extends ASTNodeAccessImpl implements Expression {
```
manticore-projects: Do we really need an extra class for this? Why not just carry a normal "Identifier" or "String" along? I would love to avoid especially the extra methods in the Visitors.

src/test/java/net/sf/jsqlparser/statement/select/OptionClauseTest.java
```diff
@@ -147,6 +150,55 @@ public void testOptionInUpdateAndDelete() throws JSQLParserException {
...
     }
 
+    @Test
+    public void testOptionOptimizeForUnknownParameter() throws JSQLParserException {
+        String sql = "SELECT * FROM t WHERE c = @p OPTION (OPTIMIZE FOR (@p UNKNOWN))";
+        PlainSelect plainSelect =
+                (PlainSelect) assertSqlCanBeParsedAndDeparsed(sql, true);
+        OptionHint optimizeFor = plainSelect.getOption().getOptionHints().get(0);
+        Assertions.assertEquals(1, optimizeFor.getParameters().size());
+        Assertions.assertTrue(optimizeFor.getParameters().get(0) instanceof UnknownVariable);
```
manticore-projects: Assertions should provide `assertInstanceOf` directly.

manticore-projects (changes requested):

src/main/java/net/sf/jsqlparser/expression/UnknownVariable.java
fudianchn: Replaced `UnknownVariable` with a table-less `Column` and removed the class plus every Visitor addition. `Column#getColumnName()` retains `@p UNKNOWN`.

src/test/java/net/sf/jsqlparser/statement/select/OptionClauseTest.java
fudianchn: Replaced both `instanceof` assertions with `Assertions.assertInstanceOf(...)` and used the returned `Column` for the value assertions.

manticore-projects:
Thank you, I like this more.

PR #2487, opened 2026-08-18
Separate JSON operators with spaces so #> deparses correctly

dngr2:
`JsonExpression.toString()` appends each operator directly onto its operand, so
`a #> '{b}'` deparses to `a#>'{b}'`. Because `#` is a legal identifier character,
that output re-parses as `a#` `>` `'{b}'` -- a GreaterThan comparison rather than
the original JSON expression, so deparse+reparse is not round-trip safe.

Emit a space on each side of the operator, matching how the JsonOperator is printed
elsewhere. Added a test that parses `#>` and `#>>`, deparses, and checks the result
still parses back to a JsonExpression.


manticore-projects:
Thank you much!

PR #2489, opened 2026-08-19
avoid quadratic backtracking when parsing bracketed server names

dxbjavid:
Server's SERVER_PATTERN lets the first capturing group consume backslashes, so a bracketed name that is full of backslashes but never closes (for example `[a\a\a...`) makes the matcher explore every possible split between the server and instance groups, and the matching cost grows roughly quadratically with the length of the string, which can stall a thread that builds a Server from an untrusted qualified name. Excluding the backslash separator from the first group removes that ambiguity so matching stays linear, while `[server\instance]` and plain `[server]` still split exactly as before. The added test builds a Server from a long crafted value and would previously time out.


dxbjavid:
Thanks for the merge. Appreciated.

PR #2492, opened 2026-08-22
feat(parser): support MySQL PROCEDURE ANALYSE clause

### LINKED ISSUE #2428: [BUG] JSQLParser 5.4-SNAPSHOT : MySQL : PROCEDURE ANALYSE() not supported in SELECT statements
### Failing SQL Feature
MySQL `PROCEDURE ANALYSE()` clause used in `SELECT` statements.

MySQL (notably in older versions such as 5.5) supports `PROCEDURE ANALYSE()` as a query modifier that analyzes a result set and suggests optimal column types and sizes. It is appended directly to a `SELECT` statement and returns metadata instead of modifying data.

Although seems to be deprecated and removed in modern MySQL versions, it is still valid syntax in legacy mysql databases and may appear in migration scenarios or older production systems.

### SQL Example
```sql
CREATE TABLE heavy_table (
    id INT AUTO_INCREMENT PRIMARY KEY,
    col1 INT,
    col2 VARCHAR(50)
);

INSERT INTO heavy_table (col1, col2) VALUES
(10, 'alpha'),
(20, 'beta'),
(10, 'gamma');

SELECT col1, col2
FROM heavy_table
PROCEDURE ANALYSE(10, 256);
```

### Parsing error
`ParseException: Encountered: <S_IDENTIFIER> / "ANALYSE", at line 1, column 46, in lexical state DEFAULT.`

dengliming:
Closes [https://github.com/JSQLParser/JSqlParser/issues/2428](https://github.com/JSQLParser/JSqlParser/issues/2428)

- add ANALYSE as a non-reserved keyword
- parse zero, one, or two integer parameters
- prevent PROCEDURE ANALYSE from being consumed as an alias
- add AST, deparser, and visitor support
- add parsing and regression tests

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/select/MySqlProcedureAnalyse.java
+++ b/src/main/java/net/sf/jsqlparser/statement/select/MySqlProcedureAnalyse.java
@@ -1,0 +1,75 @@
+/*-
+ * #%L
+ * JSQLParser library
+ * %%
+ * Copyright (C) 2004 - 2026 JSQLParser
+ * %%
+ * Dual licensed under GNU LGPL 2.1 or Apache License 2.0
+ * #L%
+ */
+package net.sf.jsqlparser.statement.select;
+
+import java.io.Serializable;
+import net.sf.jsqlparser.expression.LongValue;
+import net.sf.jsqlparser.parser.ASTNodeAccessImpl;
+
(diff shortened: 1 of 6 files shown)


manticore-projects:
Thank you!

PR #2495, opened 2026-08-22
avoid quadratic backtracking in the oracle hint comment pattern

dxbjavid:
OracleHint checks every comment that precedes a select/insert/update/delete/merge against the block-hint pattern with an unanchored find, so a single line comment that happens to contain a /*+ marker followed by a long run of stars and no closing */ makes the greedy match backtrack quadratically (a comment with 100k stars takes around twelve seconds), and this is reachable straight through CCJSqlParserUtil.parse. A hint only ever begins the comment though, so anchoring the pattern at the start of the input lets it reject that line-comment case immediately while still matching every real /*+ ... */ hint exactly as before. I have added a regression test that pins the crafted comment under a short preemptive timeout.


manticore-projects:
Thank you much!

dxbjavid:
TY. Appreciated

PR #2496, opened 2026-08-22
feat(parser): support DuckDB MAP literals and PIVOT queries

dengliming:
Support and closes [https://github.com/JSQLParser/JSqlParser/issues/2423](https://github.com/JSQLParser/JSqlParser/issues/2423)

## Summary

Adds support for DuckDB MAP literals and simplified PIVOT queries.

Supported examples:

```sql
SELECT MAP {'key': 'value', 'key2': 'other'} AS m FROM t;

CREATE VIEW v AS
SELECT MAP {'category': name, 'count': name} AS summary
FROM products;

PIVOT sales ON region USING SUM(amount);

SELECT * FROM (PIVOT sales ON region USING SUM(amount));
```

## Changes

- Added `MapExpression` for DuckDB `MAP {key: value}` literals.
- Added `PivotQuery` for DuckDB simplified PIVOT syntax.
- Integrated both nodes with visitors, deparsers, validators, and `TablesNamesFinder`.
- Added support for PIVOT as a top-level query and subquery.
- Preserved the existing Oracle/SQL Server-style `PIVOT (...)` syntax.

## Tests

Added tests covering all four reported DuckDB cases, additional PIVOT clauses, table-name extraction, and existing PIVOT compatibility.

Relevant parser and regression tests pass without introducing new JavaCC warnings.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitor.java
+++ b/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitor.java
@@ -621,6 +621,14 @@ public interface ExpressionVisitor<T> {
         this.visit(arrayConstructor, null);
     }
 
+    default <S> T visit(MapExpression mapExpression, S context) {
+        return null;
+    }
+
+    default void visit(MapExpression mapExpression) {
+        this.visit(mapExpression, null);
+    }
+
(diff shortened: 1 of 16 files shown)


manticore-projects:
This one is awesome! Thank you so much!

PR #2497, opened 2026-08-22
fix(parser): support CHARACTER SET in CAST CHAR types

### LINKED ISSUE #2298: [BUG] JSQLParser 5.4-SNAPSHOT : MySQL : Failed to parse CAST as CHAR with CHARACTER SET
### Failing SQL Feature:
It seems that a combination of casting with character set definition doesn't work, while it's a valid syntax in MySQL.

Error:
```
ParseException: Encountered: <K_SET> / "SET", at line 1, column 37, in lexical state DEFAULT.
Was expecting one of these terminals within expansion starting at 6381:5:
    <K_FORMAT> (inside 6379:5) ...
....
```

### SQL Example:

```
CREATE TABLE demo_strings (
    id INT AUTO_INCREMENT PRIMARY KEY,
    txt VARCHAR(50)
) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;

INSERT INTO demo_strings (txt) VALUES
('hello'),
('abc'),
('こんにちは');

SELECT CAST('abc' AS CHAR CHARACTER SET utf8mb4);
```

### Software Information:
Query is valid with recent MySQL versions

dengliming:
Fixes [https://github.com/JSQLParser/JSqlParser/issues/2298](https://github.com/JSQLParser/JSqlParser/issues/2298)


manticore-projects:
Thank you!

Issue #2499, opened 2026-08-23
[BUG] JSQLParser 5.4-SNAPSHOT : MySQL : # line comments are not supported


fudianchn:
AI disclosure: this issue was prepared with AI coding agents, reviewed and revised line by line by me.

### Failing SQL Feature:

MySQL `#` line comments are not parsed. MySQL supports `#` to end of line as a comment style ([MySQL 8.0 Manual, Comments](https://dev.mysql.com/doc/refman/8.0/en/comments.html)). JSqlParser handles `-- ` and `/* */`, but `#` fails.

Context: `#` is currently lexed for the PostgreSQL JSON operators `#>` and `#>>` (`JSqlParserCC.jjt`, token handling around line 8403). A line comment rule must not swallow those: with javacc longest-match, a plain `#` to end of line token would also consume `#>` / `#>>`, so the comment rule needs to exclude a `>` immediately after `#`.

### SQL Example:

```sql
SELECT 1 # comment
```

```
net.sf.jsqlparser.JSQLParserException: net.sf.jsqlparser.parser.ParseException: Encountered: <K_COMMENT> / "comment", at line 1, column 12, in lexical state DEFAULT.
```

A standalone comment line fails the same way:

```sql
SELECT 1
# trailing comment
```

```
net.sf.jsqlparser.parser.ParseException: Encountered: <K_TRAILING> / "trailing", at line 2, column 3, in lexical state DEFAULT.
```

`SELECT 1 -- comment` parses fine.

### Software Information:

- JSqlParser version: master `1e4e92b` (also present in 5.1)
- Database: MySQL 8.0

manticore-projects:
Do we really need/want that? I have not problem with `\n# this is a comment` but `SELECT 1 # comment` adds lots of complexity, when at the same time there is SQL:2016. Why can't we expect MySQL doing things in a sane way?

fudianchn:
@manticore-projects lexically both forms are the same rule and the trailing one adds no complexity beyond the own-line one: a `#` to end of line token that does not match when `#` is immediately followed by `>` keeps `#>` / `#>>` intact through longest match, the same mechanism that already separates `/` from `/*` (`SELECT 4 / 2` and `SELECT 1 /* c */ + 2` both parse on master). Restricting the comment to line-initial `#` would take more machinery than covering both, since javacc token regexes have no line anchor.

[...]

Issue #1695, opened 2022-12-15
MySQL "#..." comments style not recognized 


TearsOfJupiter:
The following MySQL query will produce the following exception:
`#Here's a MySQL comment\nselect 1`

`Exception in thread "main" net.sf.jsqlparser.JSQLParserException
	at net.sf.jsqlparser.parser.CCJSqlParserUtil.parse(CCJSqlParserUtil.java:65)
	at net.sf.jsqlparser.parser.CCJSqlParserUtil.parse(CCJSqlParserUtil.java:40)
	at Scratch.main(scratch_3.java:11)
Caused by: net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: "#Here" <S_IDENTIFIER>
    at line 1, column 1.`

It would be nice if JSqlParser was able to parse MySQL `#`-style comments.

manticore-projects:
Greetings.

As far as I understand it. the `#` is used as an Operator in other supported SQL Dialects.

Since Comments would be discarded by the parser in any, the best approach will be to mangle your SQL before handing it to the parser and to suppress any of those comments. (e. g. using a `String.replaceAll()`  and a proper Regex).

TearsOfJupiter:
Ah, ok. Thank you!

Issue #1197, opened 2021-05-18
Single-line comment in mysql statement does not support # symbol


cxfly:
**Describe the bug**
      Single-line comment in mysql statement does not support # symbol

**To Reproduce**
1. Example SQL
       #tbds_sales_room\r\nalter table tbds_sales_room add `sale_type` tinyint(4) NOT NULL DEFAULT '0'
2. Parsing this SQL using JSqlParser with this statements
        String sql = "--tbds_sales_room\r\nalter table tbds_sales_room add `sale_type` tinyint(4) NOT NULL DEFAULT '0'";
        CCJSqlParserManager parserManager = new CCJSqlParserManager();
        Statement stmt =parserManager.parse(new StringReader(sql));
        System.out.println( stmt);
3. Exception
     net.sf.jsqlparser.JSQLParserException: Encountered unexpected token: "#tbds_sales_room" <S_IDENTIFIER>

**System**
 - Database you are using
            MySQL
- Java Version
     1.8
- JSqlParser version
   jsqlparser-4.0.jar

I found a solution, modify file
/jsqlparser/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt

from

SPECIAL_TOKEN:
{
   < LINE_COMMENT: "--"(~["\r","\n"])*>
|  < MULTI_LINE_COMMENT: "/*" (~["*"])* "*" ("*" | (~["*","/"] (~["*"])* "*"))* "/">
}

to

SPECIAL_TOKEN:
{
   < LINE_COMMENT: ("--" | "#")(~["\r","\n"])*>
|  < MULTI_LINE_COMMENT: "/*" (~["*"])* "*" ("*" | (~["*","/"] (~["*"])* "*"))* "/">
}

manticore-projects:
Unfortunately, your sample is inconsistent regarding the #.
However, you statement will work fine:
```sql
--#tbds_sales_room
ALTER TABLE tbds_sales_room
    ADD sale_type TINYINT (4) NOT NULL
;
```
You can [test it here interactively](http://217.160.215.75:8080/jsqlformatter/demo.html?args=-c%20LTDEBcCMBMGcH1YEMA2BTBAnA9tgtgFACCAMgCoCiASgARlEBCJFNUciqG8O+BN-NIgBEhNZOnjgAngAc0dAJIA5AJrKyNABQAWAJQ0lAeQ1KAqiRIEA3EA+)

The proper Java String representation was:
```java
[...]

Issue #677, opened 2018-09-08
should bracket quotation (like sqlserver supports it) be removed to be able to support array constructs


wumpz:
At the moment JSqlParser supports quotation like [colname]. This is the reason it cannot support simple array constructs using brackets as well. 

As the bracket quotation, IMHO, should be replaced by double quotes. 

Should the bracket quotation be removed to get the way free to arrays?

wumpz:
So no opinion from JSqlParsers users so far.

I tend to do extactly this, removing bracket quotation for version 2.0 of JSqlParser, to do array processing.

CodingFabian:
since we have seen arrays but not this quotation, I would vote for yes

wumpz:
So be it. Starting with version 2 JSqlParser drops support for bracket quotation.

yeswaydude:
There is a huge history of SQL that has continued to run against SQL Server for decades that uses the square bracket-quoted identifiers.  If I recall correctly, some SQL Server syntax can only be accessed via the square brackets (i.e. FOR XML directives).

Does someone have an example of array syntax with an explanation of how support for that precludes support for square bracket-quoted identifiers?

wumpz:
Some would be (from Postgresqls manual (http://www.postgresqltutorial.com/postgresql-array/) ):

```
SELECT
 id,
 name,
 phones [ 2 ]
FROM
 contacts
WHERE
 id = 3;
```

SqlServer would interpret `[2]` as an alias, which is perfectly fine. Nevertheless array access using brackets is more or less standardized as well within most programming languages.

gordthompson:
IMO it certainly would be a shame if "drops support for bracket quotation" meant that v2.x would be completely unable to parse T-SQL like

```sql
SELECT [column name] FROM [table name];
```

As mentioned in a previous comment, like it or not there is an enormous amount of T-SQL out there that uses `[column name]` instead of `"column name"`, despite what ANSI (or others) might say.

wumpz:
I will happily reopen this issue. I as well was surprised, that so little seemed to be interested. I want a discussion. I want suggestions how to implement it. I would keep both, if it works, but I assume is collides to much: 

select a [b] 

is using sqlserver an alias and using e.g. postgres an array access.

gordthompson:
[...]

PR #2502, opened 2026-08-23
fix: support MySQL # line comments (#2499)

fudianchn:

## What

Support MySQL `#` line comments, fixing #2499:

```sql
SELECT 1 + 2 # note
, 3
# leading comment
SELECT 1
```

Both forms currently fail with a `ParseException`: `#` lexes as an identifier and the comment text then fails as keywords (`SELECT 1 # comment` fails on `comment`).

## Why / Root cause

`LINE_COMMENT` only knows `--` and `//`. A `#` falls through to the identifier token (`#` is a legal identifier start and part character), so the comment text itself has to parse and does not.

## How

One additional alternation in the `LINE_COMMENT` token: a `#` followed by a blank runs to end of line. The blank gate keeps every existing `#` lexing intact:

- `#temp`, `##global`, `#$tab1#`, `a#b` stay identifiers (#1197, #1695)
- `#>` / `#>>` JSON operators are untouched (`>` is not a blank; `-#` and `<#>` never start at the `#`)

## Scope

- `SELECT 1 #comment` (no blank) still lexes as an identifier (alias), as before: treating it as a comment would collide with `#temp` style identifiers.
- A bare `#` column followed by a blank (`SELECT # FROM t`) now starts a comment instead of parsing; no dialect defines such an unquoted column name.
- `SELECT 5 # 3` (PostgreSQL bitwise XOR, never supported) now parses as `SELECT 5` plus comment, the same way `SELECT 1 -- 2` already behaves.

## Testing

`CCJSqlParserUtilTest`: 5 new tests. The 3 comment form tests were verified failing on master; the 2 guard tests pin the identifier and JSON operator families (they fail when the blank gate is removed). Full suite green.

## Performance

`gradle jmh`, `JSQLParserBenchmark.parseSQLStatements` on `performance.sql`, `version=latest`, 10 forks × 10 iterations (100 samples) on a 32-core host:

| build | ms/op |
| --- | --- |
| master `1e4e92b` | 3.755 ± 0.024 |
| branch `eaf9884` | 3.789 ± 0.029 |

Δ +0.9% with overlapping 99.9% CIs → no regression.

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -1845,7 +1845,11 @@ TOKEN : /* Numeric Constants */
 
 SPECIAL_TOKEN:
 {
-   < LINE_COMMENT: ("--" | "//") (~["\r","\n"])*>
+   // The `#` comment requires a following blank: `#` without one stays an
+   // identifier start (SQL Server `#temp`, `##global`) and `#>` / `#>>`
+   // JSON operators must keep their usual lexing (issue #1197, #1695).
+   < LINE_COMMENT: ("--" | "//") (~["\r","\n"])*
+       | "#" [" ", "\t"] (~["\r","\n"])*>
 }
 
 // Nested block comments:  /* ... /* ... */ ... */
--- a/src/test/java/net/sf/jsqlparser/parser/CCJSqlParserUtilTest.java
+++ b/src/test/java/net/sf/jsqlparser/parser/CCJSqlParserUtilTest.java
@@ -222,6 +222,40 @@ public class CCJSqlParserUtilTest {
     }
 
     @Test
+    public void testLineCommentHashTrailing() throws Exception {
+        assertEquals("SELECT 1", CCJSqlParserUtil.parse("SELECT 1 # trailing comment").toString());
+    }
+
+    @Test
+    public void testLineCommentHashLeading() throws Exception {
+        assertEquals("SELECT 1", CCJSqlParserUtil.parse("# leading comment\nSELECT 1").toString());
+    }
+
+    @Test
+    public void testLineCommentHashWithinSelect() throws Exception {
+        assertEquals("SELECT 1 + 2, 3",
+                CCJSqlParserUtil.parse("SELECT 1 + 2 # note\n, 3").toString());
+    }
+
+    @Test
+    public void testHashWithoutBlankStaysIdentifier() throws Exception {
+        // without a following blank `#` is an identifier start, not a comment (#1197, #1695)
+        TestUtils.assertSqlCanBeParsedAndDeparsed("SELECT * FROM #temp");
+        TestUtils.assertSqlCanBeParsedAndDeparsed("SELECT * FROM ##global");
+        TestUtils.assertSqlCanBeParsedAndDeparsed("SELECT * FROM #$tab1#");
+        TestUtils.assertSqlCanBeParsedAndDeparsed("SELECT * FROM $#tab1#");
+        TestUtils.assertSqlCanBeParsedAndDeparsed("SELECT * FROM #$tab#tab1");
+        TestUtils.assertSqlCanBeParsedAndDeparsed("SELECT a#b FROM t");
+        TestUtils.assertSqlCanBeParsedAndDeparsed("SELECT 1 #note");
+    }
+
+    @Test
+    public void testHashJsonOperatorsStayOperators() throws Exception {
+        TestUtils.assertSqlCanBeParsedAndDeparsed("SELECT js #> '{a}' FROM t");
+        TestUtils.assertSqlCanBeParsedAndDeparsed("SELECT js #>> '{a,b}' FROM t");
+    }
+
+    @Test
     public void testParseStatementsFail() throws Exception {
         String sqlStr = "select * from dual;WHATEVER!!";
 


manticore-projects:
> SELECT 5 # 3 (PostgreSQL bitwise XOR, never supported) now parses as SELECT 5 plus comment, the same way SELECT 1 -- 2 already behaves.

Ooofffff!
This is evil syntax (which I did not even know exists) -- but so is the MySQL syntax. And my heart is much more on PostgreSql than it is on MySQL. Let me think about this and sorry for being dramatic here, I just feel like there is no good solution to this.

fudianchn:
> This is evil syntax (which I did not even know exists) -- but so is the MySQL syntax.

For completeness on the lexeme: besides the bitwise XOR row in Table 9.4, Mathematical Operators (`integral_type # integral_type → integral_type`, "Bitwise exclusive OR", `17 # 5 → 20`), the current documentation (PostgreSQL 18) gives `#` three more meanings, all in Table 9.36, Geometric Operators:

- `# geometric_type → integer`: "Returns the number of points. Available for path, polygon." (`# path '((1,0),(0,1),(-1,0))' → 3`)
- `geometric_type # geometric_type → point`: "Computes the point of intersection, or NULL if there is none. Available for lseg, line." (`lseg '[(0,0),(1,1)]' # lseg '[(1,0),(0,1)]' → (0.5,0.5)`)
- `box # box → box`: "Computes the intersection of two boxes, or NULL if there is none." (`box '(2,2),(-1,-1)' # box '(1,1),(-2,-2)' → (1,1),(-1,-1)`)

Sources: [Mathematical Functions and Operators](https://www.postgresql.org/docs/current/functions-math.html), [Geometric Functions and Operators](https://www.postgresql.org/docs/current/functions-geometry.html)

manticore-projects:
Honestly a tough call, we are doomed when we do and also when we don't.
Nobody cared about those Postgres Operators so far. So I see three choices:

1) we implement/add the Postgres Operators just to have a strong and sound argument why the MySQL Comment syntax won't be supported
2) we implement the MySQL Comment syntax "for the time being", just to throw it out when there is a solid demand for the Postgres Operators
3) Or we just close the cursed PR and work on something else 

I am in favor of 1) or 3), but I won't oppose when you prefer 2). You have provided so much good stuff that I trust your good judgement.

fudianchn:
My position up front: either leave it as is (closing this PR is fine with me), or fix it properly with a switch, so that no reading silently loses data. The proper fix is an architecture decision that is yours to make; I will not push it unilaterally.

> Honestly a tough call, we are doomed when we do and also when we don't.

**The simplest decision.** Put "never silently change semantics" first, and current master is not only acceptable but safer than this PR. First-hand runs on both sides (master `1e4e92b` vs PR branch `eaf9884`, each built locally; re-checked on current master `4c6a4fb`):

| SQL | master | this PR |
| --- | --- | --- |
| `SELECT 42 # 24` | loud error | parses -> `SELECT 42` (`# 24` silently dropped) |
| `SELECT 1 # comment` | loud error | parses -> `SELECT 1` |
| `SELECT 1 # comment, 2` | loud error | parses -> `SELECT 1` (the `, 2` column silently dropped) |

All three are correct under real MySQL semantics, so for the MySQL reading this PR fixes a real gap. The problem is the other reading: data silently lost, no error at all. Master is the only state where neither reading gets silently changed.

**The "I want it all" route.** My gut reaction to such conflicts is "I want it all :)", so I searched the code for switch-like mechanisms and found the earlier cases:

- #677 (2018-2019): the square bracket deadlock (SQL Server quoting vs PG arrays), settled in v3.0 as arrays by default plus the `withSquareBracketQuotation` switch;
- the backslash switch (`allowBackslashEscapeCharacter`, in place since 4.6, 2023): the same kind of lexer-level switch;
- the Dialect enum + EXASOL gating: existing precedent for dialect-specific syntax behind a switch.

This class of problem has come up more than once, so for maintaining these switches we could consider a **Feature set + Dialect enum extension**, in three phases as I expect it:

1. a preset layer, on the order of a hundred lines, classifying the historical cases;
2. afterwards each new lexeme conflict remains an incremental one-Feature change hooked to the matching preset;
3. rebuilding the tokenizer itself into multi-mode (whole tokenization rule sets switched per dialect) is where effort and risk get large. This was debated in 2019: in #677 SerialVelocity sketched a full builder API (`.withTSQLBracketNotation().withPostgresArraySyntax()...`), and wumpz's response was: "Productions you can disable, but tokens? How would you do that using JavaCC? The brackets are part of a token." The answer that landed in v3.0 was to **not touch the tokenizer structure**: tokens match as usual, then the token action rewrites the token and backs up the input stream for re-tokenization when the switch is off. The square bracket and backslash switches still work exactly this way today. So for "one sign, two dialects, split by switch", following that historical decision (match-then-rewrite, no multi-mode rework) seems the more reasonable path.

**If it is not worth it.** The flag mechanism has existed for over six years and one more flag is cheap, so the real question is not feasibility but **which side the default takes**. My preference is **PG syntax by default**, for two reasons:

1. PG XOR would become usable: on master `SELECT 42 # 24` is a ParseException, with a PG default it would parse as XOR;
2. fail-loudly mostly survives: a MySQL `#` comment would still error as long as its text is not a valid expression (real comments are mostly natural language, most would still error). The exception is text that happens to be a valid expression: `42 # 24` would silently parse as XOR 50, and `1 # note` as XOR against the column `note`. That has to be stated as a known limitation.

With PG as the default, if real demand shows up on the MySQL side, one flag (off by default) enables `#` comments. And whether that single flag or the dialect presets above, once a switch exists the dividing line is the same: on the MySQL side `#` is an unconditional line comment (in real MySQL semantics `42#24` is a comment too; in that mode there is no competing `#temp`, `#>` or XOR need, so not even the space gate is needed); on the PG side `#` is all operators (`#>`/`#>>`/`#-` JSON family plus bare `#`), with no `#` comment. The two modes do not leak into each other.

manticore-projects:
Greetings,

this would be the best indeed and you are right: token manipulation should work although there is one particular challenge here. SPECIAL_TOKEN vs. TOKEN, so far we have manipulated only TOKEN vs. TOKEN. JavaCC is very poorly equipped for such use-cases and we won't get any help from anyone.

But if you want to do this, you have my full support. I would suggest starting to implement the Postgres `INTERSECT` operator, because this has some merit on its own. Once this works, we try to bend the token into a Special Token so it finds MySQL comments.

fudianchn:
> I would suggest starting to implement the Postgres `INTERSECT` operator, because this has some merit on its own. Once this works, we try to bend the token into a Special Token so it finds MySQL comments.

First step up for review: #2507, the Postgres `#` binary operator (bitwise XOR, docs Table 9.4; geometric intersection of lseg / line / box, Table 9.36). A dedicated token declared before `S_IDENTIFIER` wins the length tie on a lone `#`; longest match keeps every other `#` lexing untouched (`#temp`, `##global`, `#$tab1#`, `a#b`, `#>`, `#>>`). The operator assertions were verified failing on master and passing on the branch; the guard test pins the identifier and JSON families.

One behavior delta disclosed there: a lone `#` can no longer be an identifier, so bare names (`SELECT # FROM t`, `SELECT 1 #`, ...) fail loudly instead of parsing silently; quoted `"#"` still parses.

Second step: `Feature.allowHashLineComments` (off by default) routing `#` into a Special Token per lexer state, so MySQL line comments work under the switch. A spike says MORE + SwitchTo handles it; `#word` adjacency and `SELECT 1 #` at EOF are the open edges to disclose in that PR. I will start on it once this lands.

manticore-projects:
I assume, this is obsolete after #2508?

fudianchn:
Yes. #2508 landed the same feature behind `allowHashLineComments` and closed #2499. Closing this one.

PR #2504, opened 2026-08-23
fix(parser): parse qualified columns in GROUP BY

### LINKED ISSUE #2485: [BUG] JSQLParser Version 5.3.218: RDBMS Oracle: failing group by parsing
### Failing SQL Feature:
- `GROUP BY` parsing error when referencing a column inside a parenthesis with its fully qualified name followed by a comma. 

### SQL Example:
- Simplified Query Example, focusing on the failing feature
    ```sql
      select sys.dual.dummy, count(*) from sys.dual
      group by (sys.dual.dummy), (sys.dual.dummy)
    ```
Message is:  `Encountered: <EOF>, at line 2, column 43, in lexical state DEFAULT.  `
### Software Information:
- JSqlParser Manticore version 5.3.218 
- Database:  Oracle

jianjindream:
Fixes #2485

## Summary

Parenthesized fully qualified columns in a `GROUP BY` list could be parsed as
a lambda expression after a comma.

The fixed `LOOKAHEAD(7)` accepted the seven-token `(sys.dual.dummy)` prefix
before verifying the required `->`. This caused the parser to enter
`LambdaExpression` and fail at EOF.

The lookahead now matches the explicit unparenthesized lambda form,
`RelObjectName() "->"`. Parenthesized lambdas continue to be handled through
the regular `Expression()` path.

## Tests

Coverage includes:

- the original Oracle `GROUP BY` regression from #2485;
- a four-part qualified-column boundary case;
- unparenthesized and four-parameter lambda AST assertions;
- JSON `->` and `->>` AST assertions.

Validation:

- Gradle check with JDK 17
- 4,920 tests, 0 failures

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -7966,7 +7966,7 @@ ExpressionList SimpleExpressionList():
     (
         LOOKAHEAD(2, {!interrupted} ) ","
         (
-            LOOKAHEAD( 7 ) expr=LambdaExpression()
+            LOOKAHEAD( RelObjectName() "->" ) expr=LambdaExpression()
             |
             expr=SimpleExpression()
         )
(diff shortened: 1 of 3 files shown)


manticore-projects:
Thank you much for fixing this, it was broken indeed (and I caused that myself).

Issue #2490, opened 2026-08-21
[BUG] 5.3 / master : MySQL : valid index DDL fails to parse (prefix+direction key parts, CREATE/DROP INDEX options, CREATE FULLTEXT/SPATIAL, multi-valued indexes)


minleejae:
### Failing SQL Feature:

Several **valid MySQL index DDL statements** fail to parse. Every statement listed below was verified to **execute successfully on MySQL 8.4.11** (official Docker image), and then parsed with `CCJSqlParserUtil.parse()` against both release **5.3** and **master @ 4c6a4fb** (2026-08-19). The failures group into 6 root causes.

For contrast, closely related forms that already parse fine on master are listed per group, which should help localize each gap.

---

#### 1. Key part with both prefix length **and** direction: `col(len) [ASC|DESC]`

```sql
CREATE INDEX i03 ON t (c1(20) DESC);
CREATE INDEX i04 ON t (c1(20) ASC, c2(10) DESC);
CREATE UNIQUE INDEX i25 ON t (c1(10) DESC);
ALTER TABLE t ADD INDEX i05 (c1(20) DESC);
ALTER TABLE t ADD INDEX i33 (c1(20) ASC);
```
Error: `Encountered unexpected token: "DESC" "DESC"` (resp. `"ASC"`).

Works on master: `c1(20)` (prefix only), `c1 DESC` (direction only, since #2387), `(LOWER(c1)) DESC` (expression key part + direction), and `ADD UNIQUE INDEX ux (c1(10) DESC)` (unique-key path). Only the *combination* of prefix length + direction on a named column fails.

#### 2. `CREATE INDEX` with `KEY_BLOCK_SIZE`

```sql
CREATE INDEX i08 ON t (c1) KEY_BLOCK_SIZE=8;
CREATE INDEX i09 ON t (c1) KEY_BLOCK_SIZE 8;
CREATE INDEX i14 ON t (c1) USING BTREE KEY_BLOCK_SIZE=8 COMMENT 'combo' INVISIBLE;
```
Error: `Encountered unexpected token: "KEY_BLOCK_SIZE"`.

Works on master: `ALTER TABLE t ADD INDEX i (c1) KEY_BLOCK_SIZE = 8` — the ALTER path already supports it, only the `CREATE INDEX` statement rejects it.

#### 3. `CREATE INDEX` with `algorithm_option` / `lock_option`

```sql
[...]

PR #2505, opened 2026-08-24
fix(parser): support MySQL index options and prefix-length key parts in index DDL

minleejae:
Refs #2490

## Description

Fixes five of the six groups of MySQL index DDL reported in #2490. These now parse:

```sql
-- 1. key part with a prefix length *and* a direction (CREATE and ALTER)
CREATE INDEX i03 ON t (c1(20) DESC);
ALTER TABLE t ADD INDEX i05 (c1(20) DESC);

-- 2/3. CREATE INDEX index_option / algorithm_option / lock_option
CREATE INDEX i08 ON t (c1) KEY_BLOCK_SIZE = 8;
CREATE INDEX i10 ON t (c1) ALGORITHM = INPLACE LOCK = NONE;

-- 4. DROP INDEX algorithm_option / lock_option
DROP INDEX i15 ON t ALGORITHM = INPLACE LOCK = NONE;

-- 5. FULLTEXT / SPATIAL (previously fell back to UnsupportedStatement)
CREATE FULLTEXT INDEX i18 ON t (body) WITH PARSER ngram;
CREATE SPATIAL INDEX i19 ON t (g);
```

Two causes:

- `IndexColumnWithParams()` allowed only **one** `CreateParameter()` per key part, so `(20)` consumed it and the trailing `ASC`/`DESC` could not be matched. It now collects parameters in a loop. `IndexColumnsWithParamsList()` is shared by the CREATE and ALTER paths, so one change fixes both.
- The option keywords (`KEY_BLOCK_SIZE`, `ALGORITHM`, `LOCK`, `NONE`, `PARSER`, `FULLTEXT`, `SPATIAL`) were simply not reachable from the option lists that `CREATE INDEX` and `DROP INDEX` use.

## No new grammar ambiguity

Those keywords all live in `NonReservedWord()`, so adding them as new alternatives to the trailing `(...)*` loops pushed JavaCC from 13 to 15 choice-conflict warnings, and explicit syntactic `LOOKAHEAD` did not help. They are instead added to the **existing flat token lists** of `CreateParameter()` and `Drop()`, which adds no new choice point and leaves the warning count unchanged.

`LOCK` is the one real ambiguity, since it also starts a `LOCK TABLE` statement. It is guarded so it is only taken as a `DROP` argument when it cannot begin the next statement:

```java
LOOKAHEAD({ getToken(1).kind == K_LOCK && getToken(2).kind != K_TABLE })
tk=<K_LOCK> { dropArgs.add(tk.image); }
```

## Scope

Group 6 (`CAST(... AS UNSIGNED ARRAY)` for multi-valued indexes) is **not** included: it needs an `ARRAY` marker on the public `ColDataType` model, and `K_ARRAY_LITERAL` is already used in six other places. Better as its own PR, so #2490 stays open.

## Testing

- `./gradlew check` passes (checkstyle, PMD, spotless, spotbugs, JaCoCo).
- **4923 tests, 0 failures, 0 errors.**
- JavaCC reports `0 errors and 13 warnings`, identical to `master` — verified by building both states and diffing the warning list.
- Seven new tests in `CreateIndexTest`, `AlterTest` and `DropTest`, including one that pins `DROP TABLE t1; LOCK TABLE t2 IN SHARE MODE;` still parsing as two statements.

Deparsing follows the existing conventions: `c1 (20) DESC` (matching the `mycol2 (75)` output already asserted in `CreateIndexTest`) and `KEY_BLOCK_SIZE = 8`. All statements above were verified against MySQL 8.4.11 while writing #2490.

## PR Checklist

- [x] I have read the contribution guidelines and the governance document on PR expectations.
- [x] Minimal changes to code not directly related to your change
- [x] One feature/change per PR unless tightly coupled
- [x] Do a rebase on upstream `master`

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -10980,6 +10980,7 @@ List<Index.ColumnParams> ColumnNamesWithParamsList() : {
 Index.ColumnParams IndexColumnWithParams(): {
     String columnName = null;
     List<String> parameter = null;
+    List<String> param = null;
     Expression expression = null;
     Index.ColumnParams column = null;
 }
@@ -10987,14 +10988,30 @@ Index.ColumnParams IndexColumnWithParams(): {
     (
         columnName=RelObjectName()
         { parameter = null; }
-        [ parameter = CreateParameter() ]
+        // MySQL allows a key part to carry a prefix length and a direction, e.g. "c1(20) DESC",
+        // so more than one parameter has to be collected here.
+        ( LOOKAHEAD(2) param = CreateParameter()
+            {
+                if (parameter == null) {
+                    parameter = new ArrayList<String>();
+                }
+                parameter.addAll(param);
+            }
+        )*
         {
             column = new Index.ColumnParams(columnName, parameter);
         }
         |
         "(" expression=Expression() ")"
         { parameter = null; }
-        [ LOOKAHEAD(2) parameter = CreateParameter() ]
+        ( LOOKAHEAD(2) param = CreateParameter()
+            {
+                if (parameter == null) {
+                    parameter = new ArrayList<String>();
+                }
+                parameter.addAll(param);
+            }
+        )*
         {
             column = new Index.ColumnParams(expression, parameter);
         }
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -10979,41 +10979,30 @@ List<Index.ColumnParams> ColumnNamesWithParamsList() : {
 
 Index.ColumnParams IndexColumnWithParams(): {
     String columnName = null;
+    // the options collected for this key part, and the result of a single CreateParameter()
+    List<String> columnParams = new ArrayList<String>();
     List<String> parameter = null;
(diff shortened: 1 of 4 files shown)


src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
```diff
@@ -10980,21 +10980,38 @@ List<Index.ColumnParams> ColumnNamesWithParamsList() : {
 Index.ColumnParams IndexColumnWithParams(): {
     String columnName = null;
     List<String> parameter = null;
+    List<String> param = null;
```
manticore-projects: sorry, why do we have now `parameter` and `param`? Can you please name it in a way so we know which one is what for?

manticore-projects (changes requested):

src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
minleejae: Good catch — the names were the wrong way round. The grammar already spells this pattern twice as createOptions.addAll(parameter) / columnSpecs.addAll(parameter), so the accumulator is now columnParams, parameter stays the per-call value, and the loop body becomes the same one-liner. [ec61e7f](https://github.com/JSQLParser/JSqlParser/pull/2505/commits/ec61e7f1c261827791c0077d44afbb82e8da3e0d)

manticore-projects:
Much nicer now, thank you for work and effort!

PR #2507, opened 2026-08-25
fix(parser): support PostgreSQL # binary operator (bitwise XOR / geometric intersection)

fudianchn:

## What

Support the PostgreSQL `#` binary operator, the first step agreed in #2502 (see #2499):

```sql
SELECT 17 # 5                       -- bitwise exclusive OR (docs Table 9.4)
SELECT lseg1 # lseg2                -- geometric intersection of lseg / line / box (Table 9.36)
SELECT a # (b + 1) FROM t
SELECT * FROM t WHERE a # b = 1
```

All currently fail with a `ParseException`.

## Why / Root cause

`#` is a legal identifier start and part character, so a lone `#` lexes as a
one-character identifier and never reaches an operator production.

## How

A dedicated `S_HASH_OPERATOR` token declared before `S_IDENTIFIER` (it wins
the length tie on a lone `#`) plus one case pair in `prattArithRest`
(precedence 5, with `+` `-` `&`, the existing mapping of PostgreSQL's
"any other operator" tier). Longest match keeps every other `#` lexing
untouched: `#temp`, `##global`, `#$tab1#`, `a#b` stay identifiers
(#1197, #1695), `#>` / `#>>` JSON operators are unchanged.

The AST node is the dedicated `Intersects` operator, sibling of
`GeometryDistance`, wired into `ExpressionVisitor`, the adapter, the
deparser, `TablesNamesFinder` and the validator.

One behavior change: a lone `#` can no longer be an identifier. Bare
column / table / alias names written as a single `#` (`SELECT # FROM t`,
`SELECT 1 FROM #`, `SELECT a AS # FROM t`, `SELECT 1 #`, `SELECT t.# FROM t`,
`INSERT INTO # VALUES (1)`) previously parsed silently and now fail loudly
as an operator without operands; quoted forms (`"#"`, `` `#` ``) still parse.

Follow-up to #2502: with this in place, the MySQL `#` line comment can be
attached behind a feature switch by routing the token into a special token
per lexer state (the SPECIAL_TOKEN vs TOKEN point).

## Testing

`PostgresTest`: 2 new tests. The operator assertions were verified failing
on master; the guard test pins the identifier (incl. Oracle `#$tab1#`) and
JSON operator families and fails when `#` is dropped from the identifier
start set. Full suite green (4926 tests).

## Performance

`gradle jmh`, `JSQLParserBenchmark.parseSQLStatements` on `performance.sql`,
`version=latest`, 10 forks × 10 iterations (100 samples) on a 32-core host,
interleaved master/branch × 2:

| build | run 1 | run 2 |
| --- | --- | --- |
| master `9fc38bd` | 3.773 ± 0.021 | 3.799 ± 0.030 |
| branch (this PR) | 3.757 ± 0.024 | 3.777 ± 0.024 |

Δ within each window ≤ 0.6% with overlapping 99.9% CIs → no regression.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitor.java
+++ b/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitor.java
@@ -40,6 +40,7 @@ import net.sf.jsqlparser.expression.operators.relational.GreaterThan;
 import net.sf.jsqlparser.expression.operators.relational.GreaterThanEquals;
 import net.sf.jsqlparser.expression.operators.relational.InExpression;
 import net.sf.jsqlparser.expression.operators.relational.IncludesExpression;
+import net.sf.jsqlparser.expression.operators.relational.Intersects;
 import net.sf.jsqlparser.expression.operators.relational.IsBooleanExpression;
 import net.sf.jsqlparser.expression.operators.relational.IsDistinctExpression;
 import net.sf.jsqlparser.expression.operators.relational.IsNullExpression;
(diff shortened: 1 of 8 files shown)


manticore-projects:
Nice and clean! Thank you for providing this, its definitely a win.

PR #2511, opened 2026-08-26
feat(parser): turn on the dialect's lexer feature preset in withDialect

### LINKED ISSUE #2510: [FEATURE] dialect presets for the lexer-level feature switches

Not a single grammar production, but the switch layer below it. JSqlParser now has three lexer-level switches, each added when one lexeme carries different meanings in two dialects:

- `Feature.allowSquareBracketQuotation` (v3.0, #677: SQL Server quoting vs Postgres arrays)
- `Feature.allowBackslashEscapeCharacter` (v4.6: backslash escapes)
- `Feature.allowHashLineComments` (#2508: MySQL `#` line comments vs the Postgres `#` operator from #2507)

Every lexeme conflict settled the same way adds one more switch, and every user of a dialect then has to know and wire the full set by hand:

```java
CCJSqlParserUtil.parse(sql, parser -> parser.withBackslashEscapeCharacter(true)
        .withHashLineComments(true));
```

Proposal: a preset layer so one call selects the dialect's full switch set. Prototyping found the parser already owns a narrow `Dialect` enum (`ORACLE`, `EXASOL`) behind `withDialect(Dialect)` for grammar gating, so extending that enum keeps a single entry point:

```java
CCJSqlParserUtil.parse(sql, parser -> parser.withDialect(Dialect.MYSQL));
```

Pure data mapping, no mechanism changes: individual switches keep working and override the preset.

### SQL Example

One statement, MySQL reading, today vs proposed:

```sql
-- MySQL: `col` is a quoted identifier, 'x\'yz' keeps the escaped quote, 42#24 is 42 plus a line comment
SELECT `col` FROM t WHERE a = 'x\'yz' AND b = 42#24
```

today: two switches wired manually per parse call (backticks already parse by default, brackets are not MySQL syntax); proposed: `withDialect(Dialect.MYSQL)` once.

### Additional context

- Current master `177dbd7`.
- Follow-up sketched at the end of #2508: "The dialect presets sketched in the #2502 discussion would be a small standalone follow-up if you want them: a mapping from DatabaseType to the existing Feature set, no mechanism changes."
- #2502 holds the full history of the one-lexeme-two-dialects class of problem (square brackets, backslash, `#`).
- Open design points: which switches each preset turns on (POSTGRESQL may need none, since the defaults already follow the PG reading), whether the mapping lives on the parser's `Dialect` enum or beside the validation-side `DatabaseType`, and whether the validation feature sets and the parse-time switches should share one table.
- #2511 answers the first two points (empty presets for the PG-default dialects, mapping on the parser's `Dialect` enum); the shared validation table stays open as a follow-up.

fudianchn:

## What

`withDialect(Dialect.MYSQL)` now also enables that dialect's lexer feature preset, on top of the grammar gating it already performs (#2510):

```java
CCJSqlParserUtil.parse("SELECT `col` FROM t WHERE a = 'x\\'yz' AND b = 42#24",
        p -> p.withDialect(Dialect.MYSQL))
// -> SELECT `col` FROM t WHERE a = 'x\'yz' AND b = 42
```

## Why

Each one-lexeme-two-dialects conflict settled so far left a switch behind: square bracket quotation (v3.0, #677), backslash escapes (v4.6), `#` line comments (#2508). Every user of a dialect has to know and wire the full set by hand, and the set grows by one on every future conflict.

## How

`Dialect` gains `ANSI_SQL`, `MYSQL`, `MARIADB`, `SQLSERVER`, `POSTGRESQL` and `H2` next to `ORACLE` and `EXASOL`, and carries its preset as an `EnumSet<Feature>`; `withDialect` enables the preset after setting `Feature.dialect`. The existing EXASOL grammar gating (`Import()/Export()`, `SubImport()`) compares enum names and is untouched; the EXASOL preset stays empty.

Presets per dialect, from the vendor docs:

| dialect | preset |
| --- | --- |
| MYSQL, MARIADB | backslash escapes, `#` line comments |
| SQLSERVER | square bracket quotation |
| ANSI_SQL, ORACLE, POSTGRESQL, H2, EXASOL | none (the defaults) |

MySQL 8 manual ([Comments](https://dev.mysql.com/doc/refman/8.0/en/comments.html), [String Literals](https://dev.mysql.com/doc/refman/8.0/en/string-literals.html)), MariaDB KB ([Comment Syntax](https://mariadb.com/kb/en/comment-syntax/), [String Literals](https://mariadb.com/kb/en/string-literals/)), Microsoft Learn ([Database Identifiers](https://learn.microsoft.com/en-us/sql/relational-databases/databases/database-identifiers)). Brackets are not MySQL syntax and `#` stays reserved for SQL Server identifiers (temp tables), so the presets do not cross.

The preset only turns features on, never off: an explicit switch after `withDialect` wins, and a switch enabled before it survives (MYSQL does not disable bracket quotation set earlier). Pure data mapping, no lexer changes.

Design alternative: a separate `withDatabaseType(DatabaseType)` entry point beside the validation-side `DatabaseType` enum would leave two dialect enums and two builders; extending `Dialect` keeps one entry point. Merging the validation feature sets and the parse-time switches into one table looked like a larger refactor of the validation API and is not attempted here. If you prefer the separate entry point, the rework is small.

## Testing

`CCJSqlParserUtilTest`: `testDialectPresets` (the MySQL family, SQLSERVER, and the empty presets keeping the #2507 operator while rejecting `\'` escapes), `testDialectPresetSwitchOverrides` (a switch after the preset wins, the preset never disables one). All verified failing with the preset loop removed, with MYSQL missing the backslash feature, and with the preset turning foreign switches off; full suite green.

## Performance

No benchmark: `withDialect` runs once per parse call at configuration time (the same `FeatureConfiguration` writes the individual switches make) and the token loop is untouched, so default parses execute identical code.

Closes #2510

### DIFF
--- a/src/main/java/net/sf/jsqlparser/parser/AbstractJSqlParser.java
+++ b/src/main/java/net/sf/jsqlparser/parser/AbstractJSqlParser.java
@@ -13,7 +13,10 @@ import net.sf.jsqlparser.parser.feature.Feature;
 import net.sf.jsqlparser.parser.feature.FeatureConfiguration;
 
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.EnumSet;
 import java.util.List;
+import java.util.Set;
 
 public abstract class AbstractJSqlParser<P> {
 
@@ -22,7 +25,21 @@ public abstract class AbstractJSqlParser<P> {
     protected List<ParseException> parseErrors = new ArrayList<>();
 
     public enum Dialect {
-        ORACLE, EXASOL
+        ANSI_SQL, ORACLE, MYSQL(Feature.allowBackslashEscapeCharacter,
+                Feature.allowHashLineComments), MARIADB(Feature.allowBackslashEscapeCharacter,
+                        Feature.allowHashLineComments), SQLSERVER(
+                                Feature.allowSquareBracketQuotation), POSTGRESQL, H2, EXASOL;
+
+        private final Set<Feature> lexerFeatures;
+
+        Dialect(Feature... lexerFeatures) {
+            this.lexerFeatures = lexerFeatures.length == 0 ? EnumSet.noneOf(Feature.class)
+                    : EnumSet.copyOf(Arrays.asList(lexerFeatures));
+        }
+
+        public Set<Feature> getLexerFeatures() {
+            return lexerFeatures;
+        }
     }
 
     public P withSquareBracketQuotation() {
@@ -62,7 +79,11 @@ public abstract class AbstractJSqlParser<P> {
     }
 
     public P withDialect(Dialect dialect) {
-        return withFeature(Feature.dialect, dialect.name());
+        withFeature(Feature.dialect, dialect.name());
+        for (Feature lexerFeature : dialect.getLexerFeatures()) {
+            withFeature(lexerFeature, true);
+        }
+        return me();
     }
 
     public P withAllowedNestingDepth(int allowedNestingDepth) {
--- a/src/site/sphinx/usage.rst
+++ b/src/site/sphinx/usage.rst
@@ -318,3 +318,15 @@ Additionally there are Features to control the Parser's effort at the cost of th
             , parser -> parser
                 .withBackslashEscapeCharacter(true)
     );
+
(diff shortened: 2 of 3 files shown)


manticore-projects:
I like this very much! Thank you for working this out!
We still will need to update the documentation/website regarding this new paradigm.

fudianchn:
@manticore-projects Documentation added to the parser features section of the usage page (also covering withHashLineComments, which was not documented yet).

manticore-projects:
Awesome! Thank you much again!

PR #2514, opened 2026-08-26
feat(parser): concatenate adjacent string literals behind adjacentStringLiterals

fudianchn:

## What

Adjacent string literals concatenate behind a mode, item 3 of #2512:

```java
CCJSqlParserUtil.parse("SELECT 'a'\n'b'", p -> p.withAdjacentStringLiterals(AdjacentStringLiterals.NEWLINE))
// -> SELECT 'ab' (one merged StringValue)
```

## Why

Implementing this turned out considerably simpler than the #2512 discussion suggested, so it seems worth having after all: the newline rule needs no lexer work, the token line numbers carry it, and one guarded loop in `PrimaryExpression` is the whole grammar change. The default is OFF and byte-identical to today, where the second literal stays the MySQL / SQL Server alias.

Three readings, one String valued Feature (the shape `Feature.dialect` already has):

| mode | reading | dialects |
| --- | --- | --- |
| OFF (default) | alias in the select list, error in expressions | MySQL, SQL Server (today's behavior) |
| NEWLINE | concatenate when the whitespace contains a newline | the SQL standard, PostgreSQL |
| WHITESPACE | concatenate across any whitespace | GoogleSQL, Spark/Databricks |

The result is a single merged `StringValue`, matching how the engines describe it ("treated as if the string had been written as one constant"), not a Concat expression, so deparsing never invents a `||`. The `ANSI_SQL` and `POSTGRESQL` presets carry NEWLINE; H2 concatenates too but its separator rule is unverified, so its preset stays off for now. Item 2 (the dollar quoting) is independent and its timing is yours.

## How

One guarded loop after the string literal branch in `PrimaryExpression`; the semantic lookahead reads the mode, checks the next token kind, and for NEWLINE compares `getToken(1).beginLine > getToken(0).endLine`. Same-line input keeps the alias reading under NEWLINE, exactly the split the engines have. Combined with `allowDoubleQuotedStrings`, `"1" "2"` concatenates under WHITESPACE, the GoogleSQL chunking shape. Literals consumed by dedicated productions (interval and friends) concatenate only where they go through ordinary expressions, same as today; a continuation part carrying a prefix (say `E'a'\n'b'`) merges its value and keeps the first part's prefix.

## Testing

`CCJSqlParserUtilTest`: `testAdjacentStringLiteralsNewline` (off = alias and error unchanged, newline merge incl. three parts and expression positions, same line stays alias, ANSI/Postgres presets on, MySQL preset off), `testAdjacentStringLiteralsWhitespace` (same-line merge, the BigQuery `"1" "2"` shape combined with double quoted strings). All verified failing with the loop removed, with the newline gate removed, and with ANSI_SQL missing the preset; full suite green.

## Performance

`gradle jmh`, `JSQLParserBenchmark.parseSQLStatements` on `performance.sql`, `version=latest`, 10 forks × 10 iterations (100 samples) on a 32-core host, interleaved master/branch:

| build | run 1 | run 2 |
| --- | --- | --- |
| master `c86cf6a` | 3.795 ± 0.027 | 3.787 ± 0.026 |
| branch (this PR) | 3.791 ± 0.029 | 3.789 ± 0.024 |

Same-window deltas are -0.1% and +0.05% with fully overlapping CIs: no regression. The lookahead is one mode comparison returning immediately while OFF, paid after a string literal only.

Implements item 3 of #2512.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/parser/AbstractJSqlParser.java
+++ b/src/main/java/net/sf/jsqlparser/parser/AbstractJSqlParser.java
@@ -100,7 +100,7 @@ public abstract class AbstractJSqlParser<P> {
     public P withDialect(Dialect dialect) {
         withFeature(Feature.dialect, dialect.name());
         if (dialect.getAdjacentStringLiterals() != AdjacentStringLiterals.OFF) {
-            withFeature(Feature.adjacentStringLiterals, dialect.getAdjacentStringLiterals().name());
+            withAdjacentStringLiterals(dialect.getAdjacentStringLiterals());
         }
         for (Feature lexerFeature : dialect.getLexerFeatures()) {
             withFeature(lexerFeature, true);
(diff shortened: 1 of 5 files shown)


manticore-projects:
Thank you much!

manticore-projects:
Thank you much!

PR #2515, opened 2026-08-27
feat(parser): add BIGQUERY, DATABRICKS and SNOWFLAKE dialect presets

fudianchn:

## What

Three warehouse presets, following up on the four warehouses named in #2512:

```java
CCJSqlParserUtil.parse("SELECT `col` FROM t WHERE a = 'x\\'yz' AND b = 42#24",
        p -> p.withDialect(Dialect.BIGQUERY))
// -> SELECT `col` FROM t WHERE a = 'x\'yz' AND b = 42
```

| preset | `"` | `#` | `\` | adjacent literals |
| --- | --- | --- | --- | --- |
| BIGQUERY | string literal | line comment | escape sequences | WHITESPACE |
| DATABRICKS | string literal | the #2507 operator | escape sequences | WHITESPACE |
| SNOWFLAKE | quoted identifier | the operator | escape sequences | OFF (alias) |

## Why

Every row is taken from the vendor docs, and Redshift, the fourth warehouse surveyed, matches the defaults on every switch and gains no entry.

- GoogleSQL, lexical structure: "Both string and bytes literals must be quoted, either with single (`'`) or double (`\"`) quotation marks", `#` single-line comments, "Backslashes (`\`) introduce escape sequences", literal chunking where "the literal value is the concatenation of all these parts" across any whitespace ("Quoted identifiers don't concatenate"); identifiers are backticked.
- Spark, literals: the syntax `[ r ] { 'char [ ... ]' | "char [ ... ]" } [ ... ]`, the escape table, "Chains of string literals are coalesced into a single string literal". The Databricks KB adds that double quoted identifiers need the non-default `doubleQuotedIdentifiers` conf, so the default reading is the string one, the same preset decision as MySQL and its ANSI_QUOTES mode in #2513. `#` is not a Spark comment.
- Snowflake, string and binary data types: "A backslash escape sequence is a sequence of characters that begins with a backslash (`\`)", the double quote "is used (as needed) for delimiting object identifiers"; no adjacent-literal concatenation documented.
- Redshift: delimited identifiers for `"`, PG-style `--` comments, and no AWS docs sentence found for backslash escapes or adjacent literals (community evidence points to escapes, unverified), so every row stays with the default and no empty preset entry is added.

Sources: [GoogleSQL lexical structure](https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical), [Spark literals](https://spark.apache.org/docs/latest/sql-ref-literals.html), [Databricks KB, double quoted identifiers](https://kb.databricks.com/dbsql/using-double-quotes-in-a-query-causes-a-syntax-error), [Snowflake string and binary data types](https://docs.snowflake.com/en/sql-reference/data-types-text), [Redshift names](https://docs.aws.amazon.com/redshift/latest/dg/r_names.html).

## How

Enum entries only; `withDialect` is data driven since #2511 and unchanged. Stacked on #2514 for the WHITESPACE mode, the diff shrinks automatically once it merges.

## Testing

`CCJSqlParserUtilTest`: `testDialectPresetsWarehouses` (BigQuery backticks + escape + `42#24` swallowed as comment, Databricks double quoted string + escape + `#` stays the operator, Snowflake escape + `#` stays the operator), `testDoubleQuotedStringsPreset` extended (string reading for BigQuery and Databricks, identifier guard for Snowflake), `testAdjacentStringLiteralsWhitespace` extended (the warehouse presets carry the mode, Snowflake stays the alias reading). All verified failing with a wrong row added or a right row removed (Databricks `#` on, Snowflake `"` on, BigQuery `#` off, WHITESPACE turned into NEWLINE); full suite green.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/parser/AbstractJSqlParser.java
+++ b/src/main/java/net/sf/jsqlparser/parser/AbstractJSqlParser.java
@@ -37,7 +37,15 @@ public abstract class AbstractJSqlParser<P> {
                         Feature.allowHashLineComments,
                         Feature.allowDoubleQuotedStrings), SQLSERVER(AdjacentStringLiterals.OFF,
                                 Feature.allowSquareBracketQuotation), POSTGRESQL(
-                                        AdjacentStringLiterals.NEWLINE), H2, EXASOL;
+                                        AdjacentStringLiterals.NEWLINE), H2, EXASOL, BIGQUERY(
+                                                AdjacentStringLiterals.WHITESPACE,
+                                                Feature.allowDoubleQuotedStrings,
+                                                Feature.allowHashLineComments,
+                                                Feature.allowBackslashEscapeCharacter), DATABRICKS(
+                                                        AdjacentStringLiterals.WHITESPACE,
+                                                        Feature.allowDoubleQuotedStrings,
+                                                        Feature.allowBackslashEscapeCharacter), SNOWFLAKE(
+                                                                Feature.allowBackslashEscapeCharacter);
 
         private final Set<Feature> lexerFeatures;
         private final AdjacentStringLiterals adjacentStringLiterals;
--- a/src/site/sphinx/usage.rst
+++ b/src/site/sphinx/usage.rst
@@ -319,7 +319,7 @@ Additionally there are Features to control the Parser's effort at the cost of th
                 .withBackslashEscapeCharacter(true)
     );
 
-Instead of turning the individual Parser Features on one by one, a ``Dialect`` preset selects the features of that database dialect: ``withDialect(Dialect.MYSQL)`` turns on ``withBackslashEscapeCharacter``, ``withHashLineComments`` and ``withDoubleQuotedStrings`` (MySQL and MariaDB syntax, the latter for the default sql_mode), ``withDialect(Dialect.SQLSERVER)`` turns on ``withSquareBracketQuotation``. ``withDialect(Dialect.POSTGRESQL)`` and ``withDialect(Dialect.ANSI_SQL)`` turn on the newline rule for adjacent String Literals. Features set explicitly after the dialect preset win over the preset.
+Instead of turning the individual Parser Features on one by one, a ``Dialect`` preset selects the features of that database dialect: ``withDialect(Dialect.MYSQL)`` turns on ``withBackslashEscapeCharacter``, ``withHashLineComments`` and ``withDoubleQuotedStrings`` (MySQL and MariaDB syntax, the latter for the default sql_mode), ``withDialect(Dialect.SQLSERVER)`` turns on ``withSquareBracketQuotation``. ``withDialect(Dialect.POSTGRESQL)`` and ``withDialect(Dialect.ANSI_SQL)`` turn on the newline rule for adjacent String Literals. ``withDialect(Dialect.BIGQUERY)`` and ``withDialect(Dialect.DATABRICKS)`` turn on ``withDoubleQuotedStrings`` and ``withBackslashEscapeCharacter`` plus the any-whitespace rule for adjacent String Literals, the BigQuery preset additionally ``withHashLineComments``; ``withDialect(Dialect.SNOWFLAKE)`` turns on ``withBackslashEscapeCharacter`` only, keeping double quotes as quoted identifiers. Features set explicitly after the dialect preset win over the preset.
 
 .. code-block:: java
 
--- a/src/test/java/net/sf/jsqlparser/parser/CCJSqlParserUtilTest.java
+++ b/src/test/java/net/sf/jsqlparser/parser/CCJSqlParserUtilTest.java
@@ -642,18 +642,50 @@ public class CCJSqlParserUtilTest {
     }
 
     @Test
+    public void testDialectPresetsWarehouses() throws Exception {
+        // BIGQUERY: double quoted strings, # line comments and backslash
+        // escapes (GoogleSQL "Lexical structure and syntax", identifiers are
+        // backticked), backticks stay unconditional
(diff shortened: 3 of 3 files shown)


src/main/java/net/sf/jsqlparser/parser/AbstractJSqlParser.java
```diff
@@ -82,12 +107,19 @@ public P withTimeOut(long timeOutMillSeconds) {
...
+        if (dialect.getAdjacentStringLiterals() != AdjacentStringLiterals.OFF) {
+            withFeature(Feature.adjacentStringLiterals, dialect.getAdjacentStringLiterals().name());
+        }
         for (Feature lexerFeature : dialect.getLexerFeatures()) {
             withFeature(lexerFeature, true);
         }
         return me();
     }
 
+    public P withAdjacentStringLiterals(AdjacentStringLiterals adjacentStringLiterals) {
```
manticore-projects, line 129: I fully understand why `"OFF"` vs. `false`, however this is a more exotic behavior and so I wonder if an overloaded method accepting  `BOOLEAN` would make sense here with `false` mapped to `OFF` and `true` mapped to `WHITESPACE`. Just to stay with the commonly used convention so far.

Though it is not the hill I want to die on.

manticore-projects:
Nice work, please consider my one comment on the Feature and then I would merge prompt. Thank you much for your contribution!

fudianchn:
@manticore-projects The comment on the Feature has not arrived on my side: the review still shows as pending (unsubmitted). Could you submit it? I will address it right away.

manticore-projects (changes requested):

src/main/java/net/sf/jsqlparser/parser/AbstractJSqlParser.java
fudianchn: The String default is the part that looked odd to me as well; it follows `dialect`, the one other feature carrying a String, and the setting is 3-state anyway, so I left it at that. Since you raised it, let's line the options up:

| | current | typed value | + boolean overload (your suggestion) |
|---|---|---|---|
| public API | one method, takes the mode | same | two methods |
| reachable states | OFF / NEWLINE / WHITESPACE | same | boolean reaches OFF / WHITESPACE, NEWLINE only via the enum |
| stored value | enum name as String | the enum constant, typed comparison | same |
| precedent | `withDialect(Dialect)` signature, `dialect` stores a String name | the rest are boolean / numeric | the 9 `withXxx(boolean)` pairs, each a 2-state toggle |
| wrong setting | visible at the call site | same | `true` silently means WHITESPACE (GoogleSQL); the standard mode is NEWLINE, a different tree without an error |
| a later 4th mode | new constant | same | the `true` mapping is frozen by compatibility |

Each column holds up in its own scenario: the boolean matches the familiar 2-state toggles, the enum matches multi-valued settings like `withDialect`, and both on-modes ship as presets anyway (ANSI_SQL / POSTGRESQL = NEWLINE, BIGQUERY / DATABRICKS = WHITESPACE), so the boolean would mainly serve hand-built configurations. I am not attached to either; if I had to pick, the middle one: store the enum constant itself and keep the single enum-typed method. Whichever you prefer, I will push it; the overload can also sit on top of the typed value.

manticore-projects: I am most interested in switching it "Off" w/o remembering the details. If Standard Mode is `Newline` then `true` should map to `Newline` of course. I don't ever expect (or want to see) a 4th mode on that :-D either.

manticore-projects:
Sorry for asking, are those 2 PRs not additive or redundant?

src/main/java/net/sf/jsqlparser/parser/AbstractJSqlParser.java
fudianchn: Agreed on `false` = OFF, `true` = NEWLINE. The overload landed in #2514 together with the typed value, and this PR is rebased on top of it. Also restored the javadoc of `isNamedExprListAhead` that inserting `isAdjacentStringConcat` had orphaned.

fudianchn:
@manticore-projects Additive by design: #2514 carries the feature, this PR only the presets on top (the first four commits here are #2514's), and the description notes the same. Once #2514 merges, this diff shrinks to the presets commit. The boolean overload landed in #2514, with this PR rebased on it.

manticore-projects:
Thank you much!

manticore-projects:
Please I will need a break for a few days to catch up with the massive wave of improvements and changes you have provided.
I have fixed already JavaCC r/ the Block Comments. I need to update/test JSQLFormatter and JSQLTranspiler as well covering the newly introduced dialects. There is not a problem with publishing the WebSite as well.

fudianchn:
Rebased onto master `84d23e7`; the branch now carries only the presets commit. The previous merge of master into the branch had kept the #2514 test methods from both sides, defining them twice and breaking compilation; the rebase removes the duplicates. `./gradlew spotlessApply check` green locally.

Enjoy the break, there is no rush on this one; it will be ready whenever you are back.

manticore-projects:
We are back!
Your Hashtag rewrite broke the `XMLDoclet` (the chosen Unicode is invalid for XML)  which broke the `publishing` task. I went down all the way and fixed it one by one. We can continue now.

fudianchn:
Oh! XML 1.0 and `\u0001`... I need to go read the spec, this never even occurred to me.

fudianchn:
Oh! XML 1.0 and `\u0001`... I need to go read the spec, this never even occurred to me.

manticore-projects:
> Oh! XML 1.0 and `\u0001`... I need to go read the spec, this never even occurred to me.

No worries, the problem was caused in `XMLDoclet` which we use to render our Website JavaDoc. You only provided the reproducer :-D

PR #2518, opened 2026-08-31
fix: prevent bare partition columns from inheriting the previous item's value

### LINKED ISSUE #2500: [BUG] JSQLParser 5.4-SNAPSHOT : Hive : INSERT PARTITION bare column inherits the previous item's value

### Failing SQL Feature:

In `INSERT ... PARTITION`, a bare (dynamic) partition column that follows an item with a value inherits that item's value instead of keeping `null`. The deparsed statement silently turns the dynamic column into a constant assignment, changing the statement's semantics.

Verified combinations (master `1e4e92b` and the 5.3 release jar):

| PARTITION spec | parsed items |
|---|---|
| `(dtime = '20220403', region)` | `dtime = '20220403'`, `region = '20220403'` (wrong) |
| `(a, b = '1', c)` | `a = null`, `b = '1'`, `c = '1'` (wrong) |
| `(dtime, hour = '10')` | `dtime = null`, `hour = '10'` (correct) |
| `(dtime = '20220403', hour = '10')` | both correct |
| `(dtime)` | `dtime = null` (correct) |

The trigger is: an item without a value following an item with one. Affects both `INSERT OVERWRITE TABLE ... PARTITION` and `INSERT INTO TABLE ... PARTITION` (shared `Insert()` grammar).

Root cause in `Partitions()` (`src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt`, around line 3924, introduced with #2323):

```
List<Partition> Partitions():
{
    List<Partition> partitions = new ArrayList<Partition>();
    Column tableColumn;
    Expression valueExpression = null;
}
{
    (
        tableColumn=Column() [ "=" valueExpression=Expression() ]
        { partitions.add( new Partition (tableColumn, valueExpression)); }
    )
    ( LOOKAHEAD(2) (
            ","
            tableColumn=Column() [ "=" valueExpression=Expression() ]
            { partitions.add( new Partition (tableColumn, valueExpression)); }
    ) )*
    ...
```

`valueExpression` is declared once for the whole production and reused for every item. When the optional `"=" Expression()` is skipped for a bare column, `valueExpression` still holds the value of an earlier item, which then lands in `new Partition(tableColumn, valueExpression)`.

Suggested fix: reset `valueExpression = null` at the start of each item. `Partition` already renders a bare column when the value is `null`, so no deparser change should be needed.

### SQL Example:

```java
Insert insert = (Insert) CCJSqlParserUtil.parse(
        "INSERT OVERWRITE TABLE t PARTITION (dtime = '20220403', region) SELECT code, region FROM src");

for (Partition p : insert.getPartitions()) {
    System.out.println(p.getColumn() + " -> " + p.getValue());
}
```

Output on master:

```
dtime -> '20220403'
region -> '20220403'
```

Expected:

```
dtime -> '20220403'
region -> null
```

The deparsed statement is `INSERT OVERWRITE TABLE t PARTITION (dtime = '20220403', region = '20220403') SELECT code, region FROM src`, so `region` no longer takes its value dynamically from the SELECT list.

### Software Information:

- JSqlParser version: master `1e4e92b`, also verified on the 5.3 release jar (`Partitions()` exists since #2323; 5.0 does not parse these forms at all)
- Database: Hive (mixed static and dynamic partition columns, see "Dynamic Partition Inserts" in the Hive LanguageManual DML)

fudianchn:

## What

A bare (dynamic) partition column no longer inherits the value expression of the preceding item in a mixed `INSERT ... PARTITION (...)` list; it keeps `null` again, as before #2323.

## Why

With a static item followed by a bare one, the deparser output silently turned the dynamic column into a constant assignment, changing the statement's semantics for Hive mixed static/dynamic partition writes (#2500).

## How

`Partitions()` declares one `valueExpression` for the whole production and the optional `"=" Expression()` unit leaves the previous item's value in place when skipped. The list loop arm now resets `valueExpression = null` before the optional unit, the same idiom already used by `DeclareType`, `Alias` and `ColumnsNamesList` for their list items. No AST, deparser or visitor change: `Partition` already renders a bare column when the value is `null`.

## Root cause

Production-level `valueExpression` in `Partitions()` is shared across list items and was not reset when an item skipped the optional value (introduced with #2323).

## Testing

- New `InsertTest#testInsertMixedStaticAndDynamicPartitions` covers the full #2500 combination table: static -> bare, bare -> static, static -> bare -> bare, bare -> static -> bare, and `INSERT INTO` with a numeric value. All wrong rows fail on master and pass with the change; the two correct rows are pinned as guards. Two mutants verified: removing the reset and misplacing it after the optional unit both turn the test red.
- Local run: `gradlew test --tests InsertTest` -> 79 passed.
- No jmh run: `performance.sql` contains no `PARTITION` clause, the changed action is not reached by the benchmark.

```java
Insert insert = (Insert) CCJSqlParserUtil.parse(
        "INSERT OVERWRITE TABLE t PARTITION (dtime = '20220403', region) SELECT code, region FROM src");
// before: dtime -> '20220403', region -> '20220403'  (wrong)
// after:  dtime -> '20220403', region -> null        (correct)
```

Fixes #2500

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -4065,7 +4065,7 @@ List<Partition> Partitions():
     (
         LOOKAHEAD(2) (
             ","
-            tableColumn=Column() [ "=" valueExpression=Expression() ]
+            tableColumn=Column() { valueExpression = null; } [ "=" valueExpression=Expression() ]
             { partitions.add( new Partition (tableColumn, valueExpression)); }
         )
     )*
--- a/src/test/java/net/sf/jsqlparser/statement/insert/InsertTest.java
+++ b/src/test/java/net/sf/jsqlparser/statement/insert/InsertTest.java
@@ -966,6 +966,58 @@ public class InsertTest {
         assertFalse(insert.isOverwrite());
     }
 
+    @Test
+    // a dynamic partition column without a value must not inherit the value of a
+    // preceding static partition column (#2500)
+    void testInsertMixedStaticAndDynamicPartitions() throws JSQLParserException {
+        String sqlStr =
+                "INSERT OVERWRITE TABLE t PARTITION (dtime = '2024-04-03', region) SELECT * FROM a";
+        Insert insert = (Insert) assertSqlCanBeParsedAndDeparsed(sqlStr);
+        assertEquals(2, insert.getPartitions().size());
+        assertEquals("dtime", insert.getPartitions().get(0).getColumn().getColumnName());
+        assertEquals("'2024-04-03'", insert.getPartitions().get(0).getValue().toString());
+        assertEquals("region", insert.getPartitions().get(1).getColumn().getColumnName());
+        assertNull(insert.getPartitions().get(1).getValue());
+
+        sqlStr =
+                "INSERT OVERWRITE TABLE t PARTITION (region, dtime = '2024-04-03') SELECT * FROM a";
+        insert = (Insert) assertSqlCanBeParsedAndDeparsed(sqlStr);
+        assertEquals(2, insert.getPartitions().size());
+        assertEquals("region", insert.getPartitions().get(0).getColumn().getColumnName());
+        assertNull(insert.getPartitions().get(0).getValue());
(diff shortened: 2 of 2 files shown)


manticore-projects:
Thank you much!

Issue #2519, opened 2026-08-31
[BUG] Parse time grows super-linearly on long dotted-name chains and nested parentheses (phase-2 lookahead), StackOverflowError at nesting depth ~2000


fudianchn:
AI disclosure: this issue was prepared with AI coding agents, reviewed and revised line by line by me.

Two adversarial statements of around 4 KB each make `CCJSqlParserUtil.parse` burn seconds of CPU on current master. These shapes do not occur in hand-written or ORM-generated SQL; the exposure is services that parse SQL they do not fully control (query gateways, SQL auditing and firewalling, BI endpoints, data-lineage scanners) plus tools that open SQL files from untrusted sources. Public GitHub alone shows ~3000 files calling `CCJSqlParserUtil.parse`, so that consumer class is not hypothetical.

Measured impact on master `9a32ff5`:

- The default `Feature.timeOut` (8000 ms) applies per attempt: a statement that times out in the simple pass is retried in the complex pass when its nesting depth allows it, so one crafted chain burns two full budgets (16 s wall clock, measured; 8 s with `withAllowComplexParsing(false)`, measured).
- `parse(String)` runs each statement on a fresh single-thread executor, so concurrent crafted statements each hold their own core (measured: four 8k-part chains, 1.88 s each serial, completed in 1.88 s wall total).
- Callers supplying a shared executor get head-of-line blocking: a `SELECT 1` submitted 100 ms behind a crafted statement waited 15.9 s (measured).
- Callers invoking the parser directly (`new CCJSqlParser(...).Statement()`) get no timeout at all; with quadratic growth a few hundred KB of chain means minutes of CPU per statement.

### 1. Long dotted-name chain: quadratic

`SELECT a` + `.b` x N + ` FROM t` (a single column, parses fine):

| N | parse time |
|---|---|
| 1000 | 124 ms |
| 2000 | 144 ms |
| 4000 | 500 ms |
| 8000 | 1847 ms |

Clean x3.5-x4 per doubling. Control: a comma-separated list of the same length parses flat (74 ms at 8000 items), so it is the dotted chain, not the statement size.

### 2. Nested parentheses: super-quadratic + StackOverflowError

`SELECT ` + `(` x N + `1` + `)` x N:

| N | parse time | outcome |
|---|---|---|
| 500 | 96 ms | ok |
| 1000 | 648 ms | ok |
[...]

PR #2520, opened 2026-08-31
fix: memoize long-chain predicate walks behind a size threshold

fudianchn:

## What

Dotted-name chains longer than 8 parts are now memoized in `isFunctionAhead()` / `isAllTableColumnsAhead()`: the walk aborts past the threshold, restarts through a per-start-token cache, and one filling walk records the answer for every eligible chain part. Chains within the threshold keep the plain single walk, with no map access and no allocation.

## Why

`SELECT a.b.b...` with 8000 parts parses in ~1.8 s on master and stays super-linear; the predicates walk the whole remaining chain unbounded and the phase-2 lookahead re-evaluates them at every position. Addresses the predicate-layer share of #2519; the engine-layer share (phase-2 re-simulation, also behind the nested-parentheses family) stays open there.

## How

Threshold gate merged into the walk itself: the non-filling pass returns a `LONG_CHAIN` sentinel past 8 delimiter/part pairs and the caller retries through the cache, so short chains cost exactly one plain walk plus one counter compare. The filling pass records the result for every chain part that would pass the same first-token guards (each part walks the identical suffix). Caches die with the parser instance.

## Root cause

Unbounded O(chain) walks x O(positions) re-evaluation = quadratic predicate cost, isolated as one of two layers in #2519.

## Testing

- New `LongChainPredicateTest`: AST shapes pinned on both sides of the threshold (8/9 delimiters), a 40-part chain, `t.*` across the threshold, and the Oracle `column(+)` exclusion on the memoized path. All green before and after (behavior guard); a head-result mutant turns all 7 red.
- Local run: `gradlew test --tests LongChainPredicateTest` -> 7 passed; full suite 4979/0.
- No CI timing assertion: the improvement is constant-factor (~2.4x at 8000 parts) and no machine-independent budget separates it.
- jmh, `JSQLParserBenchmark.parseSQLStatements` on performance.sql, version=latest, 10 forks x 10 iterations (100 samples) per run, interleaved master/branch x2 on a 32-core host: master 3.769/3.767 vs branch 3.783/3.799 ms/op, CIs overlap -> no regression.

```java
CCJSqlParserUtil.parse("SELECT a" + ".b".repeat(8000) + " FROM t");
// before: 1847 ms   after: 775 ms  (still super-linear overall: the engine layer remains, see #2519)
```

If the fork-side phase-2 fix lands and makes this redundant, closing this PR is perfectly fine.

Related to #2519 (no Fixes: the issue tracks both families and the engine layer).

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -347,15 +347,55 @@ public class CCJSqlParser extends AbstractJSqlParser<CCJSqlParser> {
      * from column references like col, schema.col, a.b.c.col.
      *
      * Replaces LOOKAHEAD(16) on Function() with a targeted O(chain-length) check.
+     * Long chains are memoized per start token: the phase-2 lookahead routines
+     * re-evaluate the same walk many times per position and each chain part
+     * would walk the very same suffix again, which makes long chains parse in
+     * quadratic time. Chains up to CHAIN_CACHE_THRESHOLD parts skip the cache
+     * entirely, keeping the normal short-name path allocation- and map-free.
      */
+    private static final int CHAIN_CACHE_THRESHOLD = 8;
+
+    // Sentinel returned by the chain walks when the non-filling pass ran past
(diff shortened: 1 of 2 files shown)


manticore-projects:
Thank you much!

Issue #2525, opened 2026-09-02
[BUG] 5.4-SNAPSHOT: PostgreSQL CREATE INDEX key and trailing clauses are missing


minleejae:
### Failing SQL feature

Several PostgreSQL index-key and trailing clauses either fail parsing or are available only as unstructured strings.

### SQL examples

```sql
CREATE INDEX pg_idx_sort
    ON pg_index_test (id DESC NULLS FIRST);

CREATE UNIQUE INDEX pg_idx_nulls
    ON pg_index_test (email) INCLUDE (payload) NULLS NOT DISTINCT;

CREATE INDEX pg_idx_with
    ON pg_index_test (id)
    WITH (fillfactor = 80) TABLESPACE pg_default WHERE active;

CREATE INDEX pg_idx_collate
    ON pg_index_test (email COLLATE "C" text_pattern_ops);
```

### Actual behavior

Tested with JSqlParser `5.4-SNAPSHOT` at `9a32ff568f580178d9a27b101596aa43eb45d0b7`:

- `NULLS FIRST` fails at `NULLS`.
- index-level `NULLS NOT DISTINCT` fails at `NULLS`.
- the storage/tablespace/predicate example fails at predicate identifier `active`.
- COLLATE and the operator class parse, but are flattened into `Index.ColumnParams.getParams()` as `[COLLATE, "C", text_pattern_ops]`. Existing `INCLUDE` and other suffixes are similarly exposed through `CreateIndex.tailParameters` strings.

### Expected behavior

Parse and deparse the documented clauses and expose structured properties for key collation, operator class/options, sort/null ordering, included columns, null-distinct behavior, storage parameters, tablespace, and predicate expression.

### Software information and server validation

- JSqlParser: `5.4-SNAPSHOT`
- Database: PostgreSQL 18.6 (`postgres:18`)
- All four statements executed successfully. `pg_indexes.indexdef` confirmed the key expression, INCLUDE columns, NULLS NOT DISTINCT, fillfactor, and predicate.
- Grammar reference: https://www.postgresql.org/docs/18/sql-createindex.html

The `CONCURRENTLY`, optional-name, and `ON ONLY` header forms are tracked in #2524.

### AI assistance

OpenAI Codex was used to assemble and execute the parser/server comparison. The cases were checked against the PostgreSQL 18 documentation and an actual PostgreSQL 18.6 server.

PR #2527, opened 2026-09-02
Support PostgreSQL CREATE INDEX header variants

### LINKED ISSUE #2524: [BUG] 5.4-SNAPSHOT: PostgreSQL CREATE INDEX header variants fail to parse
### Failing SQL feature

PostgreSQL `CREATE INDEX` header variants are rejected:

- `CONCURRENTLY` between `INDEX` and the index name;
- an omitted index name;
- `ON ONLY table_name`.

### SQL examples

```sql
CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS pg_idx_email
    ON pg_index_test USING btree (email);

CREATE INDEX ON ONLY pg_index_test USING btree (id DESC);
```

### Actual behavior

Tested with JSqlParser `5.4-SNAPSHOT` at `9a32ff568f580178d9a27b101596aa43eb45d0b7`:

- The first statement treats `CONCURRENTLY` as the index name and fails at `IF`.
- The second requires an index name and fails at `ON`.

### Expected behavior

Both statements should return `CreateIndex`, deparse losslessly, and expose `concurrently`, `only`, and the optional index name as structured properties rather than tail strings.

### Software information and server validation

- JSqlParser: `5.4-SNAPSHOT`
- Database: PostgreSQL 18.6 (`postgres:18`)
- Both forms executed successfully against a real table. The concurrent command was executed outside a transaction block.
- Grammar reference: https://www.postgresql.org/docs/18/sql-createindex.html

This issue is limited to the command header. Index key and trailing clauses are tracked separately.

### AI assistance

OpenAI Codex was used to assemble and execute the parser/server comparison. The cases were checked against the PostgreSQL 18 documentation and an actual PostgreSQL 18.6 server.

minleejae:
## What

- Parse PostgreSQL `CREATE INDEX CONCURRENTLY`.
- Allow the PostgreSQL index name to be omitted.
- Parse `ON ONLY table_name`.
- Expose `concurrently` and `only` as explicit `CreateIndex` properties.
- Preserve all variants in both `CreateIndex.toString()` and `CreateIndexDeParser`.

## Why

These server-valid PostgreSQL header forms currently fail before the index key is reached. They are part of the documented PostgreSQL 18 `CREATE INDEX` grammar: https://www.postgresql.org/docs/18/sql-createindex.html

## Testing

- `./gradlew check --no-daemon --no-build-cache -Dorg.gradle.vfs.watch=false`
- Added three parser/deparser and AST tests in `PostgreSQLCreateIndexTest`.
- Executed the same three statements successfully on PostgreSQL 18.6, including the concurrent build outside a transaction block.

Fixes #2524. The separate key and trailing-clause scope remains in #2525.

## AI assistance

OpenAI Codex was used to audit the grammar, draft the implementation, and run validation. The changes and tests were reviewed against the PostgreSQL 18 documentation and a PostgreSQL 18.6 server.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/create/index/CreateIndex.java
+++ b/src/main/java/net/sf/jsqlparser/statement/create/index/CreateIndex.java
@@ -23,6 +23,8 @@ public class CreateIndex implements Statement {
     private List<String> tailParameters;
     private boolean indexTypeBeforeOn = false;
     private boolean usingIfNotExists = false;
+    private boolean concurrently;
+    private boolean only;
 
     public boolean isIndexTypeBeforeOn() {
         return indexTypeBeforeOn;
@@ -39,6 +41,22 @@ public class CreateIndex implements Statement {
     public CreateIndex setUsingIfNotExists(boolean usingIfNotExists) {
         this.usingIfNotExists = usingIfNotExists;
         return this;
+    }
+
+    public boolean isConcurrently() {
+        return concurrently;
+    }
+
+    public void setConcurrently(boolean concurrently) {
+        this.concurrently = concurrently;
+    }
+
+    public boolean isOnly() {
+        return only;
+    }
+
+    public void setOnly(boolean only) {
+        this.only = only;
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


manticore-projects:
Nice! Thank you for this.

PR #2529, opened 2026-09-03
fix(parser): model ALTER COLUMN nullability actions

minleejae:
## What

- Add a dedicated `AlterExpression.ColumnSetNotNull` AST model for `ALTER COLUMN ... SET NOT NULL`.
- Match the existing `ColumnDropNotNull` production before the generic column-data-type production.
- Preserve both actions through `AlterExpression.toString()`.
- Validate the affected column name in `AlterValidator`.
- Add AST, parse/deparse, validation, and reflection coverage.

## Why

Both statements currently parse, but their ASTs are not reliable:

```sql
ALTER TABLE test_table ALTER COLUMN name SET NOT NULL;
ALTER TABLE test_table ALTER COLUMN name DROP NOT NULL;
```

`SET NOT NULL` is represented as a column data type named `SET` with `[NOT, NULL]` in its raw column specs. `DROP NOT NULL` is also consumed by the generic column-data-type branch even though a dedicated `ColumnDropNotNull` model already exists. Consumers therefore have to inspect and reconstruct raw tokens instead of using the AST.

This change gives both PostgreSQL nullability actions explicit AST representations without changing unrelated column option parsing.

PostgreSQL grammar reference: https://www.postgresql.org/docs/current/sql-altertable.html

## Testing

- `./gradlew check`
- Added parse/deparse and AST assertions for both actions.
- Added validator coverage for `SET NOT NULL`.
- Added the new model to `ReflectionModelTest`.

## AI assistance

OpenAI Codex was used to trace the grammar branch selection, prepare the AST/deparser/validator changes, and run validation. The generated change and tests were reviewed against the PostgreSQL documentation.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java
+++ b/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java
@@ -41,6 +41,7 @@ public class AlterExpression implements Serializable {
     private String columnOldName;
     private List<ColumnDataType> colDataTypeList;
     private List<ColumnDropNotNull> columnDropNotNullList;
+    private List<ColumnSetNotNull> columnSetNotNullList;
     private List<ColumnDropDefault> columnDropDefaultList;
     private List<ColumnSetDefault> columnSetDefaultList;
     private List<ColumnSetVisibility> columnSetVisibilityList;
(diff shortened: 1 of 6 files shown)


manticore-projects:
Thanks!

PR #2531, opened 2026-09-03
feat(parser): add column data type fragment parser

minleejae:
## What

- Add `CCJSqlParserUtil.parseColDataType(String)`.
- Add a configurable overload accepting `Consumer<CCJSqlParser>` for dialect and feature settings.
- Require the complete input to be consumed, rejecting trailing column constraints or modifiers instead of returning a partial type.
- Cover PostgreSQL array dimensions, MySQL `ENUM` arguments, parser configuration, null/empty input, and partial-input rejection.

## Why

`ColDataType()` is available on the generated parser, but there is no public `CCJSqlParserUtil` entry point for applications that receive a type fragment rather than a complete DDL statement. Such applications currently need to instantiate the generated parser directly or wrap the fragment in synthetic DDL.

The strict EOF check is important for type fragments: returning `varchar(255)` from `varchar(255) NOT NULL` would silently discard meaningful input.

## API example

```java
ColDataType type = CCJSqlParserUtil.parseColDataType(
        "enum('small','medium')",
        parser -> parser.withDialect(AbstractJSqlParser.Dialect.MYSQL));
```

## Testing

- `./gradlew check`
- Added focused tests in `CCJSqlParserUtilTest`.

## AI assistance

OpenAI Codex was used to review existing parser utility conventions, prepare the API and tests, and run validation. The generated change was reviewed for strict full-input behavior and compatibility with parser configuration.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/parser/CCJSqlParserUtil.java
+++ b/src/main/java/net/sf/jsqlparser/parser/CCJSqlParserUtil.java
@@ -30,6 +30,7 @@ import net.sf.jsqlparser.expression.Expression;
 import net.sf.jsqlparser.parser.feature.Feature;
 import net.sf.jsqlparser.statement.Statement;
 import net.sf.jsqlparser.statement.Statements;
+import net.sf.jsqlparser.statement.create.table.ColDataType;
 
 /**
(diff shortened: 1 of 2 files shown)


manticore-projects:
Thank you!

PR #2517, opened 2026-08-31
fix: collect tables from all SELECT and DML trailing clauses in TablesNamesFinder

fudianchn:
AI disclosure: this change was prepared with AI coding agents, reviewed and revised line by line by me.

## What

`TablesNamesFinder` now traverses all Expression- and Table-bearing clauses of the SELECT family and the DML trailing clauses it previously skipped, so tables located in those clauses are collected instead of silently dropped.

## Why

`TablesNamesFinder` is the public helper for table extraction (SQL analysis, auditing, firewalls); a silently incomplete result is the worst failure mode for those consumers. All shapes in #2516 parse fine today and return incomplete table sets. This continues the family fixed by #2478 / #2479 (piped queries, DML side clauses, analytic clauses).

## How

- `PlainSelect`: DISTINCT ON items, INTO tables, LATERAL VIEW generator functions, PREFERRING, GROUP BY (incl. GROUPING SETS), QUALIFY, WINDOW definitions, PIVOT / UNPIVOT, ORDER BY, LIMIT / LIMIT BY, OFFSET, FETCH, SETTINGS, INTO TEMP target.
- `ParenthesedSelect`, `SetOperationList`, `Values`: the ORDER BY / LIMIT / OFFSET / FETCH clauses they carry themselves.
- `Delete`, `Update`: their ORDER BY and LIMIT clauses (MySQL single-table DML).
- Attached PIVOT / UNPIVOT / PIVOT XML (incl. the `PIVOT XML ... IN (SELECT ...)` subquery) is reached via the newly implemented `PivotVisitor`, the same mechanism as `MergeOperationVisitor` / `PipeOperationVisitor` in #2479.

All traversals reuse the `ExpressionVisitor` / `FromItemVisitor` default helpers (`visitOrderBy`, `visitLimit`, `visit(GroupByElement)`, `visitUpdateSets`, `visitPreferringClause`, `visitTables`, `visitFromItem`) already used by `SelectVisitorAdapter` and `StatementVisitorAdapter`. Two `SelectVisitorAdapter` `@todo`s (`windowDefinitions`, `lateralViews`) are covered too, because both can carry subqueries.

## Root cause

The finder's visit methods traversed only a subset of the Expression- and Table-bearing children; the adapters already covered most of these clauses, the finder did not.

## Testing

[...]

Issue #2478, opened 2026-08-17
[BUG] TablesNamesFinder misses tables or throws for piped queries, DML statements and analytic clauses


fudianchn:
### Failing SQL Feature:

`TablesNamesFinder` misses tables silently or throws for six statement / expression families. Since it is the documented entry point for SQL auditing / firewall use cases, a silently incomplete table list is a security relevant defect.

| # | SQL family | Result on master |
|---|---|---|
| 1 | Piped query | `[]`, every table missing |
| 2 | `DELETE` with a `WITH` list | CTE tables missing, the CTE alias leaks as a phantom table |
| 3 | `MERGE` | tables inside the `ON` condition and the `WHEN` branches missing |
| 4 | `INSERT` with `SET` list, `ON DUPLICATE KEY UPDATE` or `ON CONFLICT ... DO UPDATE`; `OUTPUT` / `RETURNING` clauses of `INSERT` / `UPDATE` / `DELETE` | clause subqueries missing |
| 5 | Data modifying CTE (`WITH del AS (DELETE FROM ...)`) | `ClassCastException` |
| 6 | Analytic function with function level `ORDER BY`, window `ORDER BY` or `FILTER` clause | `NullPointerException` / silently skipped |

Root causes in `TablesNamesFinder.java` @ f41c0b8:

- `visit(FromQuery)` returns `null` without traversing anything: piped queries were never wired up
- `visit(Delete)` does not traverse `withItemsList` (Update and Insert do)
- `visit(Merge)` does not traverse `onCondition` and `operations`
- `visit(Insert)` does not traverse `setUpdateSets`, `duplicateAction`, `conflictAction`, `outputClause`, `returningClause`; `visit(Update)` / `visit(Delete)` skip `outputClause` / `returningClause` as well
- `visit(WithItem)` dispatches the payload through `getSelect()`, which assumes a Select, while data modifying CTE payloads (`ParenthesedDelete` / `ParenthesedUpdate` / `ParenthesedInsert`) are legal
- `visit(AnalyticExpression)` guards on `getFuncOrderBy()` but iterates `getOrderByElements()`, and skips `filterExpression`; #2443 fixed a sibling NPE (window element range / offset) in the same method, this guard went unnoticed

### SQL Example:

```java
// 1. piped query: returns []
TablesNamesFinder.findTables("FROM MY_TABLE1 "
        + "|> WHERE id IN (SELECT id FROM MY_TABLE2) "
[...]

PR #2538, opened 2026-09-03
test: fail the build when TablesNamesFinder misses a visitor dispatch method

fudianchn:

## What
Adds `TablesNamesFinderCompletenessTest`: it fails the build when a visitor interface declared by `TablesNamesFinder` gains a dispatch method `visit(Node, context)` that is not implemented in the `TablesNamesFinder` class hierarchy (it resolves to an interface default instead).

## Why
`TablesNamesFinder` dispatches through 8 visitor interfaces. Most `visit(Node, context)` methods are abstract, so a missing implementation fails compilation. But the interfaces progressively gain default methods (`SelectVisitor#visit(PivotQuery, context)` returns null, `ExpressionVisitor#visit(GroupByElement, context)` traverses), and for a default declared by a single interface a missing implementation compiles silently and loses the tables of that node type without any test failing.

Demonstrated on master by mutation: turning `StatementVisitor#visit(Analyze, context)` into a `return null` default and dropping the finder override compiles, keeps the whole test suite green (no existing test collects tables from `ANALYZE`), and `findTables("ANALYZE mytab")` silently returns `[]`.

After #2479 and #2517 the finder implements all 214 dispatch methods; this test pins that state for the interfaces it declares.

## How
- Enumerates the declared visitor interfaces and their `visit(Node, context)` methods by reflection and asserts each resolves to an implementation in the `TablesNamesFinder` class hierarchy (not to an interface default); interfaces declared later are covered automatically.
- Dispatch helpers carrying a `java.util` collection (e.g. `MergeOperationVisitor#visit(Collection<MergeOperation>, context)`) are skipped: bulk dispatch, not an AST node.
- `INHERITED_DISPATCH_NODES` documents the deliberate exception: `ExpressionVisitor`'s `visit(GroupByElement, context)` default already forwards the group-by expressions and the grouping sets, the finder relies on it from `visit(PlainSelect)`.
- The walk guards itself against becoming vacuous: it fails when a declared interface exposes no guarded dispatch method at all (filters too narrow) and when the finder stops declaring one of the 8 visitor families.

## Testing
- New test `TablesNamesFinderCompletenessTest#testImplementsEveryDispatchMethodOfEveryDeclaredVisitorInterface`: green on master (214 dispatch methods verified).
- Mutation 1 (`visit(Analyze, context)` turned into a default + finder override dropped): `findTables("ANALYZE mytab")` returns `[]`, the existing suite stays green, the new test fails naming `StatementVisitor#visit(Analyze, context)`.
- Mutation 2 (allowlist entry removed): the test fails naming `ExpressionVisitor#visit(GroupByElement, context)`, the allowlist is load-bearing.
- Mutation 3 (collection filter removed): the test fails naming `MergeOperationVisitor#visit(Collection, context)`, the filter is load-bearing.
- Mutation 4 (walk matches no dispatch method): the test fails through the vacuousness assertion naming all declared interfaces, the guard cannot silently degrade to always-green.
- 本地验证：`./gradlew check`（5081 tests, 0 failures）。
- 未覆盖的边界/风险：interface-level gaps only. Empty-shell implementations (an override that visits nothing, the #2478 shape) and visitor families the finder does not declare at all are not detected: both remain manual review duties. Methods declared by two interfaces at once (e.g. `visit(PivotQuery, context)`) are compiler-enforced through the unrelated-defaults rule and only belt-and-braces here.

### DIFF
--- a/src/test/java/net/sf/jsqlparser/util/TablesNamesFinderCompletenessTest.java
+++ b/src/test/java/net/sf/jsqlparser/util/TablesNamesFinderCompletenessTest.java
@@ -1,0 +1,82 @@
+package net.sf.jsqlparser.util;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+import org.junit.jupiter.api.Test;
+
+/**
+ * TablesNamesFinder collects tables by dispatching through the visitor interfaces it declares. Most
+ * dispatch methods are abstract, so a missing implementation fails the build already. But the
+ * interfaces progressively gain default methods (e.g. SelectVisitor#visit(PivotQuery, context)
+ * returns null), and for those a missing TablesNamesFinder implementation compiles silently and
+ * loses the tables of that node type without any test failing.
+ *
+ * This test fails the build when a declared visitor interface has a dispatch method visit(Node,
+ * context) for an AST node that is not implemented in the TablesNamesFinder class hierarchy (it
+ * resolves to an interface default instead). It does not catch empty-shell implementations (an
+ * override that visits nothing) nor visitor families the finder does not declare at all: both
+ * remain manual review duties.
+ */
+public class TablesNamesFinderCompletenessTest {
+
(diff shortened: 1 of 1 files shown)


manticore-projects:
Thanks!

PR #2539, opened 2026-09-04
feat(parser): model ColDataType precision and scale as structured properties

minleejae:
## Summary

- add structured `precision` and `scale` state to `ColDataType`, populated on every grammar path
- keep `getDataType()`, `toString()`, `equals()` and `hashCode()` unchanged, so deparsing and existing consumers are unaffected

`ColDataType` currently exposes type parameters in three different shapes depending on which grammar path produced it:

| SQL | `getDataType()` | `getArgumentsStringList()` |
|---|---|---|
| `VARCHAR(255)` | `"VARCHAR (255)"` | `null` |
| `MEDIUMINT(9)` | `"MEDIUMINT"` | `["9"]` |
| `TIMESTAMP(3) WITH TIME ZONE` | `"TIMESTAMP(3) WITH TIME ZONE"` | `null` |

Keyword types fold the parameters into the type string (inserting a space), identifier types keep them in the arguments list, and zoned types embed them in the lexer token image. A consumer that needs the precision has to regex the string, with a different pattern per shape. The new accessors return the parameters uniformly:

```
VARCHAR(255)                -> precision=255  scale=null
DECIMAL(10, 2)              -> precision=10   scale=2
mediumint(9)                -> precision=9    scale=null
TIMESTAMP(3) WITH TIME ZONE -> precision=3    scale=null
ENUM('x', 'y')              -> precision=null scale=null   (non-numeric arguments)
VARCHAR(MAX)                -> precision=Integer.MAX_VALUE
```

The string-based behavior stays untouched for compatibility. The hand-built expected object trees in `DeclareStatementTest` were extended with the new state, since `assertEqualsObjectTree` compares all fields reflectively.

## Testing

- `./gradlew check` passes (checkstyle, PMD, spotless, spotbugs, JaCoCo)
- Full test suite: **5059 tests, 0 failures, 0 errors**
- JavaCC reports `0 errors and 11 warnings`, identical to `master` — no new choice conflicts
- New tests in `ColDataTypeTest` pin `getPrecision()` / `getScale()` for the three shapes above, plus the no-parameter, `DECIMAL(10, 2)`, `VARCHAR(MAX)` and `ENUM(...)` cases

## AI assistance

AI tooling assisted with the implementation and test coverage. The changes were manually reviewed and verified against MySQL and PostgreSQL syntax.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/create/table/ColDataType.java
+++ b/src/main/java/net/sf/jsqlparser/statement/create/table/ColDataType.java
@@ -29,6 +29,8 @@ public class ColDataType implements Serializable {
     private String characterSet;
     private IntervalQualifier intervalQualifier;
     private List<Integer> arrayData = new ArrayList<Integer>();
+    private Integer precision;
+    private Integer scale;
 
     public ColDataType() {
         // empty constructor
@@ -38,8 +40,10 @@ public class ColDataType implements Serializable {
         this.dataType = dataType;
 
         if (precision >= 0) {
+            this.precision = precision;
             this.dataType += " (" + (precision == Integer.MAX_VALUE ? "MAX" : precision);
             if (scale >= 0) {
+                this.scale = scale;
                 this.dataType += ", " + scale;
             }
             this.dataType += ")";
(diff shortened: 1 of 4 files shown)


manticore-projects:
Thank you much!

PR #2543, opened 2026-09-04
Support MySQL ALTER TABLE rename variants

minleejae:
## Summary

- parse `ALTER TABLE old RENAME new` and `RENAME AS new` in addition to `RENAME TO new`
- expose the exact table-rename keyword as `NONE`, `AS`, or `TO`
- preserve `RENAME TO` as the default for programmatically constructed `AlterExpression` instances
- keep column and index rename parsing distinct

## Validation

- `./gradlew test spotlessCheck checkstyleMain checkstyleTest`
- parser generation remains at the baseline 11 warnings
- all three table-rename forms were executed successfully against MySQL 8.4

## PR checklist

- [x] One independently reviewable syntax feature
- [x] Structured AST access; no token reparsing
- [x] Existing rename output remains backward compatible


manticore-projects:
Thanks!

PR #2544, opened 2026-09-04
Model MySQL CREATE SCHEMA options

minleejae:
## Summary

- parse MySQL `CREATE SCHEMA` character-set, collation, and encryption options
- introduce a shared typed `DatabaseOption` model for both `CREATE SCHEMA` and `CREATE DATABASE`
- expose option kind, value, `DEFAULT`, and equals-sign usage without requiring token reparsing
- retain the existing flat `CreateDatabase#getDatabaseOptions()` representation for compatibility

## Validation

- `./gradlew test spotlessCheck checkstyleMain checkstyleTest`
- parser generation remains at the baseline 11 warnings
- combined `CHARACTER SET`, `COLLATE`, and `ENCRYPTION` schema options were executed successfully against MySQL 8.4

## PR checklist

- [x] One independently reviewable feature
- [x] Shared structured AST instead of duplicated raw-token handling
- [x] Existing `CREATE DATABASE` access remains compatible

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/create/database/CreateDatabase.java
+++ b/src/main/java/net/sf/jsqlparser/statement/create/database/CreateDatabase.java
@@ -14,6 +14,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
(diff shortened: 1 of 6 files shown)


manticore-projects:
Thanks!

PR #2549, opened 2026-09-05
feat(parser): enhance PostgreSQL CREATE POLICY support

minleejae:
## Summary

This extends the existing PostgreSQL CREATE POLICY implementation to cover the complete policy mode clause and expose its finite options through typed AST values.

* Parse AS PERMISSIVE and AS RESTRICTIVE without introducing globally reserved keyword tokens
* Add PolicyMode and PolicyCommand AST types while retaining the existing string command accessors
* Distinguish omitted clauses from their effective PostgreSQL defaults through dedicated getters
* Traverse policy tables and expressions in StatementVisitorAdapter and StatementValidator
* Add a dedicated CreatePolicyDeParser that honors custom expression visitors
* Cover restrictive and permissive policies, typed commands, defaults, invalid modes, and deparser integration

PostgreSQL syntax reference: https://www.postgresql.org/docs/current/sql-createpolicy.html

## Tests

* ./gradlew test with the complete test suite
* ./gradlew check -x spotlessJavaCheck
* ./gradlew spotlessJavaCheck reports only a pre-existing formatting difference in CCJSqlParserTest.java caused by the local fork master being behind upstream; all files in this PR pass Spotless

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/StatementVisitorAdapter.java
+++ b/src/main/java/net/sf/jsqlparser/statement/StatementVisitorAdapter.java
@@ -329,6 +329,9 @@ public class StatementVisitorAdapter<T> implements StatementVisitor<T> {
 
     @Override
     public <S> T visit(CreatePolicy createPolicy, S context) {
+        fromItemVisitor.visitFromItem(createPolicy.getTable(), context);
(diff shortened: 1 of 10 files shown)


manticore-projects:
Thank you!

PR #2550, opened 2026-09-05
feat(parser): parse navigation chains after array subscripts

fudianchn:

## What

Parse navigation chains that alternate field accesses and subscripts after any primary expression: `SELECT recommendations[0].language_id FROM recs` now parses and round-trips. Previously the dot after a subscript was unreachable, because field access was only wired inside the parenthesized branch of PrimaryExpression.

## Why

Redshift SUPER navigation and the same family in Spark/Databricks (`arr[0].field`) failed to parse. Feeding the jsqltranspiler test corpus (900 provided segments) to master `ff2bc8d` shows this as the single remaining parse gap: 899/900, the one failure being `recommendations[0].language_id` (redshift 233/234, all other dialect groups clean).

## How

A postfix navigation loop after the existing ArrayExpression slot in PrimaryExpression, via a helper production that reuses the existing `RowGetExpression` and `ArrayExpression` nodes. No new AST classes, no visitor expansion.

1. field access accepts `RelObjectNameExt()`, matching the acceptance set of ColumnIdentifier's dotted continuation, so `a[0].from` parses like `a.from` already does
2. chains like `a.b[0].c[1].d` work; subscripts after a field go through `ArrayExpression`
3. the parenthesized branch keeps its existing row-get loop on `RelObjectName()` (so `(a).from` still fails, as before); untouched as pre-existing behavior

## Root cause

A subscript consumed by Column's ArrayConstructor left no production able to consume a following dot; navigation was reachable only for parenthesized primaries.

## Testing

Two states and mutants, all executed locally on clean rebuilds:

| state | result |
|---|---|
| master `ff2bc8d` | the 5 new encoder tests fail |
| this branch | 8/8 pass |
| mutant: drop dot arm | the 5 dot-arm tests fail |
| mutant: drop subscript arm | only the 2 chain tests fail |
| mutant: dot arm accepts `*` | only the rejection test fails |

`./gradlew check`: 5271 tests, 0 failures. javacc warnings stay at 11, no new choice conflicts.

Performance (gradle jmh, JSQLParserBenchmark.parseSQLStatements on performance.sql, version=latest, 10 forks x 10 iterations, 100 samples per run, interleaved master->branch->master->branch on a 32-core host):

| build | ms/op |
|---|---|
| master `ff2bc8d` | 3.853 ± 0.034 / 3.842 ± 0.029 |
| this branch | 3.870 ± 0.034 / 3.868 ± 0.040 |

Delta inside the confidence intervals -> no regression. A second interleaved batch against the final commit ran in a polluted window (wider CIs, disclosed for honesty): branch <= master in both windows, same conclusion.

## Verification of the original report

No upstream issue; the gap came from the corpus run above. With this branch the same corpus parses 900/900 (redshift 234/234 included).

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -8535,9 +8535,24 @@ Expression ArrayExpression(Expression obj): {
     { return expr; }
 }
 
+Expression NavigationStep(Expression base):
+{
+    Expression step = null;
+    String nm = null;
+}
+{
+    (
+        "." nm=RelObjectNameExt() { step = new RowGetExpression(base, nm); }
+        |
+        step = ArrayExpression(base)
+    )
+    { return step; }
+}
+
 Expression PrimaryExpression() #PrimaryExpression:
 {
     Expression retval = null;
+    Expression nxt = null;
     Expression expression = null;
     CastExpression castExpr = null;
     TimezoneExpression timezoneExpr = null;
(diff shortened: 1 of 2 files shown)


manticore-projects:
Thank you much!

PR #2556, opened 2026-09-06
fix(parser): parse stacked NOT prefixes as nested NotExpression (#1170)

### LINKED ISSUE #1170: NotExpression parsing error
**Describe the bug**
NotExpression is not parsed correctly.

**To Reproduce**
1. `CCJSqlParserUtil.parseCondExpression("not not 1 = 1")` generates following objects: `NotExpression ( EqualsTo ( NotExpression ( 1 ) = 1 ) )`. Instead, the result should be: `NotExpression ( NotExpression ( EqualsTo (1 = 1) ) )`
2. `CCJSqlParserUtil.parseCondExpression("not not not 1 = 1")` yields Exception: `net.sf.jsqlparser.JSQLParserException: Encountered unexpected token: "not" "NOT"`

**Expected behavior**
Based on SQL operator precedence and associativity, NOT operator should be evaluated after equal (=), and NOT operator should be evaluated from right to left.

**System**
 - Database you are using - irrelative
- Java Version - 1.8
- JSqlParser version - 4.0

fudianchn:

## What
Parse stacked leading `NOT` / `!` prefixes as nested `NotExpression`s (one wrap per prefix token) instead of mis-nesting the second `NOT` onto the left operand or failing outright on the third.

## Why
Fixes #1170. `NOT` has the lowest operator precedence and is right-associative, so `not not 1 = 1` must be `NOT (NOT (1 = 1))`. Today the second `NOT` leaks into `PrimaryExpression()` and attaches to the left operand only, silently producing `NOT (NOT 1) = 1` (an `EqualsTo` between a boolean and a number), and a third prefix raises a `ParseException` (`NOT NOT EXISTS (...)` fails the same way).

## How
`Condition()` consumed at most one leading `NOT` / `!` token. The optional prefix is now a loop that consumes all consecutive leading `NOT` / `!` tokens (guarded by the same token check the previous `LOOKAHEAD` used, so the generated parser keeps 11 warnings / 0 choice conflicts) and wraps the parsed condition once per token, leftmost token outermost. The value-position `NOT` prefix in `PrimaryExpression()` (`a = NOT 1`, `1 + NOT 2`) is untouched.

## Root cause
`JSqlParserCC.jjt` `Condition()`: single optional `<K_NOT>` / `"!"` prefix instead of a repetition.

## Testing
- `ExpressionPrecedenceTest`: 17 new tests. On master, 13 fail (wrong tree for double `NOT` with `=` / `IN` / `LIKE` / `IS NULL` / `SIMILAR TO` / inside an `AND` chain / mixed case, and `ParseException` for triple prefix, `NOT NOT EXISTS`, deep stacking); 4 guards pass on both (single prefix stays flat, value-position `a = NOT 1` / `1 + NOT 2`, parenthesized `NOT NOT (...)`, round-trip in subquery / `UPDATE` / `DELETE` `WHERE`). Mutation check: reversing the wrap order fails exactly the mixed `NOT !` test, dropping the wrap loop fails all 16 `NOT`-family tests, removing the value-position `NOT` prefix in `PrimaryExpression()` fails exactly the value-position guard.
- Differential probe over 20+ adjacent forms (single `NOT` with `=` / `LIKE` / `IN` / `BETWEEN` / `IS NULL` / `EXISTS` / `SIMILAR TO`, `a NOT LIKE`, value-position `NOT`, parenthesized, `AND` / `OR` chains, `SELECT` item, `CASE WHEN`, `HAVING`, `JOIN ON`, `CHECK`): master and this branch produce byte-identical trees, so the change only affects stacked-prefix inputs.
- Token audit for the touched tokens: leading `NOT` / `!` uniformly mean logical negation in every supported dialect; `!=` is its own token (`OP_NOTEQUALSBANG`), so the loop can never swallow it; the suffix forms (`NOT IN` / `NOT BETWEEN` / `NOT LIKE` / `IS NOT NULL`) and DDL `NOT` (`NOT NULL`, `NOT DEFERRABLE`) go through other productions and are untouched.
- `./gradlew check`: 5330 tests, 0 failures.
- `gradle jmh`, `JSQLParserBenchmark.parseSQLStatements` on `performance.sql`, `version=latest`, 10 forks × 10 iterations (100 samples), interleaved M→B→M→B on a 32-core host:

| build | ms/op |
|---|---|
| master | 3.895 ± 0.032 |
| branch | 3.916 ± 0.032 |
| master | 3.906 ± 0.031 |
| branch | 3.932 ± 0.040 |

Paired deltas +0.021 / +0.026 ms/op (+0.5 % / +0.7 %) stay inside the run-level confidence intervals in both interleaved pairs → no regression.

```
// master
CCJSqlParserUtil.parseCondExpression("not not 1 = 1")
  → NotExpression(EqualsTo(NotExpression(1), 1))        // wrong: NOT sits on the left operand
CCJSqlParserUtil.parseCondExpression("not not not 1 = 1")
  → ParseException: Encountered unexpected token: "not" "NOT"

// branch
CCJSqlParserUtil.parseCondExpression("not not 1 = 1")
  → NotExpression(NotExpression(EqualsTo(1, 1)))        // NOT NOT 1 = 1
CCJSqlParserUtil.parseCondExpression("not not not 1 = 1")
  → NotExpression(NotExpression(NotExpression(EqualsTo(1, 1))))  // NOT NOT NOT 1 = 1
```

Fixes #1170

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -7812,13 +7812,16 @@ Expression Condition():
 {
     Expression result;
     Expression left;
-    boolean not = false;
-    boolean exclamationMarkNot = false;
+    // Leading NOT / ! prefixes, leftmost token wraps outermost (issue #1170).
+    final Deque<Boolean> exclamationMarks = new ArrayDeque<Boolean>();
     int oraclePrior = EqualsTo.NO_ORACLE_PRIOR;
     int oracleJoin = EqualsTo.NO_ORACLE_JOIN;
 }
 {
-    [ LOOKAHEAD(2, { getToken(1).kind == K_NOT || "!".equals(getToken(1).image) }) (<K_NOT> { not=true; } | "!" { not=true; exclamationMarkNot=true; })]
+    (
+        LOOKAHEAD({ getToken(1).kind == K_NOT || "!".equals(getToken(1).image) })
+        ( <K_NOT> { exclamationMarks.addFirst(Boolean.FALSE); } | "!" { exclamationMarks.addFirst(Boolean.TRUE); } )
+    )*
     (
         result=ExistsExpression()
         |
@@ -7863,7 +7866,10 @@ Expression Condition():
                 && result instanceof SupportsOldOracleJoinSyntax) {
             ((SupportsOldOracleJoinSyntax) result).setOraclePriorPosition(oraclePrior);
         }
-        return not ? new NotExpression(result, exclamationMarkNot) : result;
+        while (!exclamationMarks.isEmpty()) {
+            result = new NotExpression(result, exclamationMarks.pollFirst());
+        }
+        return result;
     }
 }
 
--- a/src/test/java/net/sf/jsqlparser/expression/ExpressionPrecedenceTest.java
+++ b/src/test/java/net/sf/jsqlparser/expression/ExpressionPrecedenceTest.java
@@ -10,9 +10,20 @@
 package net.sf.jsqlparser.expression;
 
 import net.sf.jsqlparser.JSQLParserException;
+import net.sf.jsqlparser.expression.operators.arithmetic.Addition;
 import net.sf.jsqlparser.expression.operators.arithmetic.BitwiseAnd;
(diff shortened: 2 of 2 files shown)


manticore-projects:
Good job! Thank you for fixing this!

PR #2567, opened 2026-09-10
Complete MySQL index options and share CREATE INDEX rendering

minleejae:
MySQL DROP INDEX rejects DEFAULT option values, and CREATE/DROP INDEX reject LOCK EXCLUSIVE. Parse their algorithm/lock options through a shared production while retaining the existing token-list APIs, option order and optional equals signs. A CREATE INDEX tail production keeps these options separate from the generic DDL parameter parser and preserves a following LOCK TABLE statement boundary.

CreateIndex.appendTo now supplies the model and CreateIndexDeParser output, removing their duplicate CREATE INDEX renderers. Header, target and key-list rendering are split into focused helpers. PostgreSQL structured tails and legacy raw tails remain supported.

The existing nested-bracket timing check failed intermittently during validation on Linux and macOS. It now compares five-sample median durations from a monotonic clock, preserving its growth threshold and parse/deparse assertions.

Validation:

- Full Gradle check and Maven verify.
- 96 CREATE/DROP syntax combinations covering algorithms, lock modes, equals signs and both orders; missing-value and statement-boundary regressions.
- Shared-renderer checks for MySQL, PostgreSQL and legacy Oracle tails.
- 88 CREATE/DROP statements executed on MySQL 8.4 (excluding the engine-incompatible COPY/LOCK NONE combination from execution testing).

References: [MySQL CREATE INDEX](https://dev.mysql.com/doc/refman/8.4/en/create-index.html), [MySQL DROP INDEX](https://dev.mysql.com/doc/refman/8.4/en/drop-index.html).


manticore-projects:
Thank you!

PR #2568, opened 2026-09-10
Fix parseStatements failure propagation and executor cleanup

minleejae:
parseStatements("SELECT FROM", parser -> parser.withAllowComplexParsing(false)) currently returns null after a parse failure. The same happens when the configured nesting limit prevents the complex retry, including a swallowed timeout exception.

Return each successful parse directly and rethrow the original JSQLParserException when retry is unavailable. Wrap the convenience overload's internally created executor in try/finally so failure also shuts it down. Caller-owned executors remain open; null/empty-input behavior and unsupported-statement recovery are preserved.

Validation:

- Full Gradle check and Maven verify.
- Regression tests for disabled complex parsing, nesting limits, deterministic timeout, caller-executor reuse, empty input and unsupported-statement fallback.


manticore-projects:
Thank you much!

PR #2572, opened 2026-09-10
Support PostgreSQL tagged dollar strings and preserve literal bodies

### LINKED ISSUE #2233: [BUG] JSQLParser 5.1: PostgreSQL: fail to parse dollar-quoted string constants with tags
The parser fails to parse dollar-quoted string constants with tags (see https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING)
SQL example:
```sql
select
$json$
[
{"some":"json","with":"properties"},
{"other":"json","with":"properties"}
]
$json$::jsonb
```
The error I get:
```
net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: "::" "::"
    at line 7, column 7.

Was expecting one of:

    <EOF>
    <ST_SEMICOLON>
```

However, replacing dollar-quotes `$json$` with single quotes `'` or removing the tag (to leave only double dollar `$$`) works.

### Software Information:
- JSqlParser 5.1
- PostgreSQL

### Additional information
It used to be supported in JSQLParser 4.5
Connected issue in DBeaver dbeaver/dbeaver#37028

minleejae:
Tagged PostgreSQL strings such as `$json$…$json$` currently fail to parse or become column references. Recognize them as `StringValue` while preserving the exact delimiter, whitespace and literal contents:

```sql
SELECT $json$[{"some":"json","with":"properties$"}]$json$::jsonb;
```

Reuse a shared delimiter recognizer in the lexer, `StringValue` and routine-body boundary detection. The lexer reads the original character stream with a linear delimiter matcher, handles longest-match identifier tokens, and updates token end positions. Dollar-quoted values keep doubled apostrophes unchanged in `getNotExcapedValue()`. A tagged routine body remains one token, so a following SQL statement is preserved.

Tagged quoting is disabled by default to preserve identifier parsing. Enable it with `withDialect(Dialect.POSTGRESQL)` or `withDollarQuotedStringTags(true)`; an explicit `withDollarQuotedStringTags(false)` after the PostgreSQL preset disables it again. Ordinary dollar-containing identifiers, positional parameters and untagged `$$` literals remain supported.

Validation:

- Full Gradle `check` and Maven `verify`.
- Re-enabled the existing tagged-string regression test.
- Empty/Unicode tags, whitespace and CRLF preservation, embedded dollar signs and other tags, casts, bind parameters, quoted identifiers, default/non-PostgreSQL identifier parsing, explicit opt-in and PostgreSQL preset overrides.
- Model/deparser round-trips, token positions, long Reader/InputStream inputs with overlapping delimiter prefixes, unterminated tags, and routine bodies followed by another statement.
- Syntax follows the [PostgreSQL dollar-quoting documentation](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING).

Fixes #2233.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/parser/feature/Feature.java
+++ b/src/main/java/net/sf/jsqlparser/parser/feature/Feature.java
@@ -809,10 +809,10 @@ public enum Feature {
     allowDoubleQuotedStrings(false),
 
     /**
-     * Recognizes PostgreSQL $tag$...$tag$ literals. Disable for dialects where these spellings are
-     * unquoted identifiers. Untagged $$ literals are unaffected.
+     * Recognizes PostgreSQL $tag$...$tag$ literals; disabled by default to preserve unquoted
+     * identifiers, enabled by the PostgreSQL dialect preset. Untagged $$ literals are unaffected.
      */
-    allowDollarQuotedStringTags(true),
+    allowDollarQuotedStringTags(false),
 
     /**
      * concatenates adjacent String Literals: NEWLINE when separated by whitespace with at least one
--- a/src/main/java/net/sf/jsqlparser/parser/feature/Feature.java
+++ b/src/main/java/net/sf/jsqlparser/parser/feature/Feature.java
@@ -809,6 +809,12 @@ public enum Feature {
     allowDoubleQuotedStrings(false),
 
     /**
+     * Recognizes PostgreSQL $tag$...$tag$ literals. Disable for dialects where these spellings are
+     * unquoted identifiers. Untagged $$ literals are unaffected.
+     */
+    allowDollarQuotedStringTags(true),
+
+    /**
      * concatenates adjacent String Literals: NEWLINE when separated by whitespace with at least one
      * newline (the SQL standard and PostgreSQL), WHITESPACE across any whitespace (GoogleSQL,
      * Spark/Databricks); OFF by default, where the second literal stays an alias (MySQL, SQL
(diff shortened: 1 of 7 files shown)


src/main/java/net/sf/jsqlparser/parser/feature/Feature.java
```diff
@@ -808,6 +808,12 @@ public enum Feature {
      */
     allowDoubleQuotedStrings(false),
 
+    /**
+     * Recognizes PostgreSQL $tag$...$tag$ literals. Disable for dialects where these spellings are
+     * unquoted identifiers. Untagged $$ literals are unaffected.
+     */
+    allowDollarQuotedStringTags(true),
```
manticore-projects: Question: would we not expected `false` here until Dialect=Postgres is set?

manticore-projects (changes requested):
Good work, just minor German nit picking.

src/main/java/net/sf/jsqlparser/parser/feature/Feature.java
minleejae: Agreed, thanks. Changed the default to `false` and enabled it in the `POSTGRESQL` dialect preset. Explicit opt-in remains available. Updated the tests and documentation; Gradle `check` and Maven `verify` pass.

manticore-projects:
고맙습니다!

manticore-projects:
I miss my time in Seoul! I have been working for Woori Bank in 2009/10 and it was one of the best times of my life! You people are just amazing and your hospitality and sophistication is outstanding! Cheers my friend!

PR #2577, opened 2026-09-10
Fix DISTINCT traversal and share optional select-item visits

### LINKED ISSUE #2573: Statement.getFeatures() throws NullPointerException on plain SELECT DISTINCT
`Statement.getFeatures()` throws a `NullPointerException` for a plain `SELECT DISTINCT`, so a read-only guard that calls it has to fail closed on every `DISTINCT` query.

**Reproduce**

```java
CCJSqlParserUtil.parse("SELECT DISTINCT id FROM foo").getFeatures();
```

```
java.lang.NullPointerException: Cannot invoke "java.util.List.iterator()" because the return value of "net.sf.jsqlparser.statement.select.Distinct.getOnSelectItems()" is null
      at net.sf.jsqlparser.statement.select.SelectVisitorAdapter.visit(SelectVisitorAdapter.java:149)
      at net.sf.jsqlparser.statement.StatementFeatureVisitor$FeatureSelectVisitor.visit(StatementFeatureVisitor.java:801)
      at net.sf.jsqlparser.statement.StatementFeatureVisitor$FeatureSelectVisitor.visit(StatementFeatureVisitor.java:769)
```

`SELECT DISTINCT ON (id) id FROM foo` and `SELECT ALL id FROM foo` classify fine; only a `Distinct` without an `ON` list is affected.

**Cause**

`SelectVisitorAdapter.visit(PlainSelect)` (`SelectVisitorAdapter.java:149`) iterates `plainSelect.getDistinct().getOnSelectItems()` without a null check; `Distinct.getOnSelectItems()` is null unless `DISTINCT ON (..)` was written. `SelectDeParser` (`:531`) guards the same list with `!= null`.

**Suggestion**

Guard the loop in `SelectVisitorAdapter` (`if (onSelectItems != null)`), and add `SELECT DISTINCT a FROM t` to `StatementFeatureVisitorTest` / `AdapterRegressionTest`. Setting an empty list from the outside is not a workaround: the deparser then prints `DISTINCT ON ()`.

minleejae:
`SELECT DISTINCT id FROM foo` throws a NullPointerException when visited or classified with `getFeatures()`, because ordinary DISTINCT has no ON list.

Share the null-safe select-item traversal used by DISTINCT ON, the SELECT list and OUTPUT. Preserve callback order and context, and leave the AST unchanged so ordinary DISTINCT cannot become DISTINCT ON ().

Validation: full Gradle `check` and Maven `verify`; plain, nested and CTE DISTINCT, DISTINCT ON/ALL controls, feature classification, model/deparser output, and custom SELECT/OUTPUT callbacks.

Fixes #2573.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/select/SelectVisitorAdapter.java
+++ b/src/main/java/net/sf/jsqlparser/statement/select/SelectVisitorAdapter.java
@@ -9,6 +9,7 @@
  */
 package net.sf.jsqlparser.statement.select;
 
+import java.util.List;
 import net.sf.jsqlparser.expression.ExpressionVisitor;
 import net.sf.jsqlparser.expression.ExpressionVisitorAdapter;
 import net.sf.jsqlparser.expression.Function;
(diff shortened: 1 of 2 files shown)


manticore-projects:
Thank you for fixing this so fast!

PR #2578, opened 2026-09-10
Validate data-modifying CTEs through statement validators

### LINKED ISSUE #2574: Validation.validate throws ClassCastException on a data-modifying CTE (WithItem.getSelect on ParenthesedDelete)
`Validation.validate(..)` throws a `ClassCastException` on a data-modifying CTE instead of reporting a validation error, so a caller using `FeaturesAllowed.SELECT` to keep writes out gets an exception rather than a verdict.

**Reproduce**

```java
Validation.validate(List.of(FeaturesAllowed.SELECT),
        "WITH x AS (DELETE FROM foo RETURNING id) SELECT * FROM x");
```

```
java.lang.ClassCastException: class net.sf.jsqlparser.statement.delete.ParenthesedDelete cannot be cast to class net.sf.jsqlparser.statement.select.ParenthesedSelect
      at net.sf.jsqlparser.statement.select.WithItem.getSelect(WithItem.java:229)
      at net.sf.jsqlparser.util.validation.validator.SelectValidator.visit(SelectValidator.java:393)
      at net.sf.jsqlparser.util.validation.validator.SelectValidator.visit(SelectValidator.java:57)
```

Same for `WITH x AS (UPDATE foo SET id = 1 RETURNING id) SELECT * FROM x` and the `INSERT .. RETURNING` form. `Statement.getFeatures()` handles all three (`MODIFIES_DATA` is certain), because `StatementFeatureVisitor` overrides `visit(WithItem)` to look at `getParenthesedStatement()`; the validator still calls `withItem.getSelect()`.

**Cause**

`SelectValidator.visit(WithItem)` (`SelectValidator.java:393`): `withItem.getSelect().accept(..)`; `WithItem.getSelect()` (`WithItem.java:229`) casts the body to `ParenthesedSelect`, and the body is a `ParenthesedDelete`.

**Suggestion**

Dispatch on `getParenthesedStatement()` as `StatementFeatureVisitor` does: a `ParenthesedSelect` continues into the select validator, a `ParenthesedInsert/Update/Delete` into the matching statement validator (so `Feature.delete` etc. are reported as not allowed), anything else as an unsupported feature. A `ValidationTest` case with `FeaturesAllowed.SELECT` and a DML CTE, expecting `Feature.delete not allowed`, would pin it.

minleejae:
Validating `WITH x AS (DELETE FROM foo RETURNING id) SELECT * FROM x` throws ClassCastException because the CTE validator assumes every body is a SELECT. INSERT and UPDATE bodies fail the same way.

Route the body through `WithItem.accept(StatementVisitor, context)` and the existing statement validators. This removes the SELECT-only cast and reuses the validation and error collection already implemented for each statement type. CTE feature checks and column-list traversal remain in the select validator.

Validation: full Gradle `check` and Maven `verify`; INSERT/UPDATE/DELETE CTEs under restricted and permissive capabilities, ordinary/recursive/nested SELECT CTEs, multiple bodies, following statements and RETURNING expression validation. Restricted DML produces validation errors instead of runtime exceptions.

Fixes #2574.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/util/validation/validator/SelectValidator.java
+++ b/src/main/java/net/sf/jsqlparser/util/validation/validator/SelectValidator.java
@@ -390,7 +390,7 @@ public class SelectValidator extends AbstractValidator<SelectItem<?>>
         if (isNotEmpty(withItem.getWithItemList())) {
             withItem.getWithItemList().forEach(wi -> wi.accept(this, context));
         }
-        withItem.getSelect().accept((SelectVisitor<?>) this, context);
+        withItem.accept(getValidator(StatementValidator.class), context);
         return null;
     }
 
--- a/src/test/java/net/sf/jsqlparser/util/validation/DataModifyingCteValidationTest.java
+++ b/src/test/java/net/sf/jsqlparser/util/validation/DataModifyingCteValidationTest.java
@@ -1,0 +1,97 @@
+/*-
+ * #%L
+ * JSQLParser library
+ * %%
+ * Copyright (C) 2004 - 2026 JSQLParser
+ * %%
+ * Dual licensed under GNU LGPL 2.1 or Apache License 2.0
+ * #L%
+ */
+package net.sf.jsqlparser.util.validation;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.List;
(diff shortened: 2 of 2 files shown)


manticore-projects:
Thank you much!

PR #2579, opened 2026-09-10
Analyse analytic and conversion functions with complete window traversal

### LINKED ISSUE #2575: StatementFeatureVisitor purity check misses AnalyticExpression (f(x) OVER ..) and TranscodingFunction (CONVERT .. USING)
`Statement.getFeatures(pureFunctions)` is meant to report every function the caller cannot prove pure, so a read-only guard can refuse it (`usage.rst`, "Proven, possible, excluded"). Two expression kinds never reach that check, so a guard built on `getUnresolvedReferences()` / `mayModifyData()` lets them through.

**Version:** `com.manticore-projects.jsqlformatter:jsqlparser:5.4.351-SNAPSHOT` (build 5.4.351-20260910.100601-1), same on `master` at `b2115ac8`.

**Reproduce**

```java
Predicate<String> nothingIsPure = name -> false;

// window-form call: the name pg_sleep is never tested
StatementFeatures f = CCJSqlParserUtil.parse("SELECT pg_sleep(1) OVER () FROM t").getFeatures(nothingIsPure);
f.getUnresolvedReferences();   // []      expected [pg_sleep]
f.mayModifyData();             // false   expected true

// CONVERT .. USING: same
f = CCJSqlParserUtil.parse("SELECT CONVERT(name USING utf8) FROM t").getFeatures(nothingIsPure);
f.getUnresolvedReferences();   // []      expected [convert]

// for comparison, the plain call is reported
f = CCJSqlParserUtil.parse("SELECT pg_sleep(1) FROM t").getFeatures(nothingIsPure);
f.getUnresolvedReferences();   // [pg_sleep]
```

**Cause**

`StatementFeatureVisitor.FeatureExpressionVisitor` overrides `visit(Function)` (`StatementFeatureVisitor.java:836-848`) and the table-function case, but not `visit(AnalyticExpression)` (`getName()`) and `visit(TranscodingFunction)`. An `AnalyticExpression` wraps the call itself; its `getExpression()` is the argument, not a `Function`, so the inherited traversal never produces a `Function` node for the name.

**Suggestion**

Treat `AnalyticExpression.getName()` like a `Function` name in the purity check, and `TranscodingFunction` as a function named `convert` (or its `getTranscodingName()`), so `mayModifyData()` and `getUnresolvedReferences()` cover `f(x) OVER (..)` and `CONVERT(x USING y)`. A negative test in `StatementFeatureVisitorTest` next to `unprovenFunctionIsPossibleNotCertain` would pin it.

minleejae:
`getFeatures(pureFunctions)` misses the function name in an analytic call or CONVERT/TRY_CONVERT expression, so unproven calls can be classified as unable to modify data. Window PARTITION BY, ORDER BY and FILTER expressions are also skipped, and an aggregate's internal ORDER BY can trigger a null-list exception.

Share function-name normalization and purity checks across the three expression forms. Use the conversion keyword as the function name, and preserve qualified analytic names through the existing Function name renderer. Unproven effects remain possible rather than certain.

WindowDefinition now supplies its partition, ordering and frame expressions to both inline and named-window traversal. ExpressionVisitorAdapter visits both ordering lists, filters and function modifiers through its existing child-visitor hook. Ordinary and analytic functions share modifier collection, preserving HAVING, keyword arguments and LIMIT expressions when a Function becomes an AnalyticExpression.

Validation: full Gradle `check` and Maven `verify`; pure/unproven function controls, qualified names, nested calls, inline/named windows, both ordering lists, FILTER, frame bounds, offsets/defaults, modifier conversion, callback order/context and existing feature/visitor regression tests.

Fixes #2575.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/expression/AnalyticExpression.java
+++ b/src/main/java/net/sf/jsqlparser/expression/AnalyticExpression.java
@@ -58,7 +58,7 @@ public class AnalyticExpression extends ASTNodeAccessImpl implements Expression 
     public AnalyticExpression() {}
 
     public AnalyticExpression(Function function) {
-        this.name = String.join(" ", function.getMultipartName());
+        this.name = function.getName();
         this.allColumns = function.isAllColumns();
         this.distinct = function.isDistinct();
         this.unique = function.isUnique();
--- a/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitorAdapter.java
+++ b/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitorAdapter.java
@@ -12,8 +12,8 @@ package net.sf.jsqlparser.expression;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import net.sf.jsqlparser.expression.operators.arithmetic.Addition;
 import net.sf.jsqlparser.expression.operators.arithmetic.BitwiseAnd;
 import net.sf.jsqlparser.expression.operators.arithmetic.BitwiseLeftShift;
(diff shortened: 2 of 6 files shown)


manticore-projects:
Thank you very much for fixing this so fast!

PR #2580, opened 2026-09-10
Preserve EXEC bind arguments and support OUTPUT parameters

### LINKED ISSUE #268: Support for OUTPUT variable arguments to procedure calls
Microsoft SQLServer has support for output variable arguments to procedure calls.
Suppose I have a procedure `myProc` which has two arguments: an input variable and an output variable. This procedure can be called like this:
`EXECUTE myProc 'foo', @outputVar OUTPUT`

More about this can be found here:
- https://technet.microsoft.com/nl-nl/library/ms187004.aspx
- http://stackoverflow.com/questions/1589466/execute-stored-procedure-with-an-output-parameter

The above query gives a ParseException in JSqlParser.

I am not quite sure how this should be implemented. I guess `net.sf.jsqlparser.expression.UserVariable` could have a boolean flag `isOutputParameter` but this wouldn't be exactly right, since it only applies to procedure call arguments and not to variables in general.
Maybe better would be to introduce an `net.sf.jsqlparser.expression.ExecuteArgument` object which implements `net.sf.jsqlparser.expression.Expression`, and which contains an expression and a `isOutputParameter` boolean flag.
In addition, `net.sf.jsqlparser.statement.execute.Execute.getExprList()` could then return a `List<ExecuteArgument>` rather than an `ExpressionList`.
What do you think about this, Tobias?

I am using JSqlParser version 0.9.5.

### LINKED ISSUE #2192: [BUG] JSQLParser 5.0-5.1: SQL Server: Stored procedure with placeholders being parsed wrong when it was working in v1
### Failing SQL Feature:
Stored procedures with placeholders are failing to parse. Multiple placeholders is failing parsing which is better.

Previously this was being parsed successfuly in 1.x but with the latest version it's not.

### SQL Example:
- Simplified Query Example, focusing on the failing feature

```sql
EXEC PCK_ACTION_BY :USER_ID 
```
is being parsed into

```sql
EXEC PCK_ACTION_BY.USER_ID 
```

### Software Information:
- JSqlParser 5.0-5.1
- MS SQL Server

minleejae:
`EXEC PCK_ACTION_BY :USER_ID` absorbs the bind argument into the procedure name, and `EXECUTE myProc 'foo', @outputVar OUTPUT` cannot be parsed. Keep the procedure-name boundary separate from colon-prefixed arguments and represent OUTPUT/OUT modifiers with an ExecuteArgument expression.

Unmarked arguments retain their existing expression types and Execute's argument-list API. The new visitor method forwards to the wrapped expression by default; the deparser and table-name visitor also dispatch it explicitly. Execute and ExecuteDeParser now share argument rendering, including parentheses, and setting a multipart procedure name replaces the previous name while preserving omitted qualifiers such as `db..p`.

Validation: full Gradle `check` and Maven `verify`; single/multiple/qualified binds, positional and named OUTPUT arguments, OUT normalization, invalid output values, ordinary CALL/EXEC arguments, following statements, custom deparsers, visitor context and existing visitor-completeness checks.

Syntax reference: [Microsoft EXECUTE documentation](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/execute-transact-sql).

Fixes #2192.
Fixes #268.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitor.java
+++ b/src/main/java/net/sf/jsqlparser/expression/ExpressionVisitor.java
@@ -10,6 +10,7 @@
 package net.sf.jsqlparser.expression;
 
 import java.util.List;
+import net.sf.jsqlparser.statement.execute.ExecuteArgument;
 import net.sf.jsqlparser.expression.operators.arithmetic.Addition;
 import net.sf.jsqlparser.expression.operators.arithmetic.BitwiseAnd;
 import net.sf.jsqlparser.expression.operators.arithmetic.BitwiseLeftShift;
@@ -71,6 +72,10 @@ import net.sf.jsqlparser.statement.select.Select;
 import net.sf.jsqlparser.statement.update.UpdateSet;
 
 public interface ExpressionVisitor<T> {
+
+    default <S> T visit(ExecuteArgument argument, S context) {
+        return argument.getExpression().accept(this, context);
+    }
 
     default <S> T visitExpressions(ExpressionList<? extends Expression> expressions, S context) {
         if (expressions != null) {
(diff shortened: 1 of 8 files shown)


manticore-projects:
Thank you much!

PR #2581, opened 2026-09-10
Support SQL Server table variables through shared table references

### LINKED ISSUE #911: SQL Server table variables not supported | SELECT columnName FROM @table
SQL Server supports storing data in table variables which are in-memory variants to temporary tables:

`SELECT columnName FROM @table`

https://docs.microsoft.com/en-us/sql/t-sql/data-types/table-transact-sql?view=sql-server-ver15#examples

minleejae:
`SELECT columnName FROM @table` and DML targeting a declared table variable fail to parse. Add a shared table-reference production for query sources and DML targets, and reuse the existing alias production for UPDATE's table/index-hint path.

Table references carry an explicit `tableVariable` flag while preserving existing Table-based APIs and visitor/deparser dispatch. Cloning preserves the flag, default catalog/schema qualification leaves local references alone, and TablesNamesFinder separates local variables from catalog tables while retaining them in `findTablesOrOtherSources`. Validation checks the SQL Server table-variable feature and skips catalog-table metadata lookups for these references.

Catalog-only grammar remains unchanged. This adds parsing and reference classification; declaration binding and table-variable column/type resolution are outside this change.

Validation: full Gradle `check` and Maven `verify`; SELECT/JOIN/CTE, INSERT, UPDATE, DELETE and MERGE roundtrips, existing DML models, metadata lookups, dialect validation, cloning/qualification, scalar variables, declarations and following statements, normal tables/index hints, and unsupported catalog-only forms.

Syntax reference: [Microsoft table documentation](https://learn.microsoft.com/en-us/sql/t-sql/data-types/table-transact-sql).

Fixes #911.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/parser/feature/Feature.java
+++ b/src/main/java/net/sf/jsqlparser/parser/feature/Feature.java
@@ -670,6 +670,9 @@ public enum Feature {
      * @see DeclareStatement
      */
     declare,
+
+    /** SQL Server local table variables in queries and DML targets. */
+    tableVariable,
     /**
      * @see SetStatement
      */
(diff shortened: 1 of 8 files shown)


manticore-projects:
Thank you!

Issue #1994, opened 2024-04-19
[BUG] JSQLParser 4.9 fails to parse subsequent statements after parsing FUNCTION statement


kxed:
hi there！

I am using the JSQLParser4.9 tool to parse SQL DDL statements, but I encountered a problem when parsing the following statement:

### SQL Example:
- Simplified Query Example, focusing on the failing feature
    ```sql
    -- Replace with your ACTUAL example
   DROP FUNCTION IF EXISTS "fin"."restore_fund_data"("in_fund_id" int8);
  CREATE OR REPLACE FUNCTION "fin"."restore_fund_data"("in_fund_id" int8)
    RETURNS "pg_catalog"."varchar" AS $BODY$
  DECLARE
	  message VARCHAR;
	  open rec_cur;
	  close proj_info_cur;
	  ********
	  RETURN 'success';
  END $BODY$
    LANGUAGE plpgsql VOLATILE
    COST 100;
    ```

### Software Information:
- JSqlParser 4.9 and 4.8
- Database PostgreSQL

Thank you very much!

manticore-projects:
Greetings.

[Both statements get parsed and no error message is thrown.](http://jsqlformatter.manticore-projects.com/jsqlformatter/demo.html?args=-c%20LTAECUFMAcBsEMDGlQHcCWAXAFqAngPYCuATqAIIDCAKgKrkAyokAHvALZyQBQAIuAHkACqABitAHI0AkgImhpo0AFEAGtIDK1DaABEAM3QA7XQDpdJSAGdMBSwH19RIwBN7L+Jni6AFLuOOzm7oLrqgxpgAHACUANzc3JTgyuTUyqAC4BDKQgxU6eJS1LLyBsZmFta2Dk6u7p7efgG1waHhRlHR3Ml04BI6utAA5vaIDbAEQxUAbvAkiNhzYeQ6ACQAQgK8AJqrfMqUecncoKDs1lbwQygAauTglAAS9-GnBNCQRqCWiKOkr6BEBMrChoCQCAArezGfQEP4kAEAKmRKORJ2yvXkAHIrEREMgrFYsfFlBJeKANltdtw8hIAOL0OnpODDKwAR1goBuAjyxQYykSAi0oAAjAAGMWxIA)

What exactly is the challenge please?

kxed:
The "INDEX" keyword also failed to convert.
``` sql
[...]

Issue #1563, opened 2022-06-11
TSQL/MS SQL Server statements/syntax not supported.


manticore-projects:
From https://github.com/dbeaver/dbeaver/issues/16756.
Two large TSQL files:

https://github.com/jOOQ/sakila/blob/main/sql-server-sakila-db/sql-server-sakila-schema.sql
https://github.com/jOOQ/sakila/blob/main/sql-server-sakila-db/sql-server-sakila-insert-data.sql

wumpz:
Did you investigate this issue further?

minleejae:
@manticore-projects I revisited the two Sakila scripts linked in this issue and verified them against current `master` (8e2d660df46617a9dc6383357968e7abdf4e86cf), using `Dialect.SQLSERVER` with unsupported-statement fallback disabled.

Both scripts parse successfully without modifying or preprocessing the input:

| Input | Parsed statements | Result |
| --- | ---: | --- |
| Schema script | 70 | Pass |
| Current insert-data script | 46,326 | Pass |
| Insert-data script as it existed when this issue was opened | 46,326 | Pass |

I also deparsed and reparsed each complete script and checked that statement types, counts, and rendered SQL were preserved. No `UnsupportedStatement` nodes were produced.

The relevant syntax gaps were addressed by:
- #2586: SQL Server index definitions
- #2603: trailing commas in SQL Server `CREATE TABLE`
- #2605: `SET IDENTITY_INSERT`

All three fixes are included in the `jsqlparser-5.4` tag; the verification above was run on current `master`.

Since the original reproductions now pass, would you agree that this issue can be closed as resolved?

manticore-projects:
Absolutely, great job!
We need to mention this on the Website as well!

PR #2586, opened 2026-09-11
Support dialect-specific CREATE INDEX definitions

### LINKED ISSUE #652: Indices with multiple parameters are not parsed correctly
### Actual Behavior
`create unique null_filtered index foo on bar (baz)` is parsed successfully, but the `unique` parameter is lost.

### Expected Behavior
`create unique null_filtered index foo on bar (baz)` should be parsed successfully and both `unique` and `null_filtered` should be part of the parsed index definition.

### Steps to Reproduce the Problem

  1. Parse `create unique null_filtered index foo on bar (baz)`
  1. Verify that the parsed CreateIndex statement does not contain the `unique` part

### Specifications

  - Version: 1.2
  - Platform: Ubuntu 18.04
  - Subsystem:

minleejae:
`CREATE INDEX` rejects Spanner's `UNIQUE NULL_FILTERED` combination, SQL Server's `UNIQUE NONCLUSTERED` combination, and the schema-qualified PostgreSQL collation/operator-class example in #1994. This change parses and preserves these definitions with an explicit dialect:

- `SPANNER`: keep uniqueness in `Index.getType()` and null filtering in `CreateIndex.isNullFiltered()`.
- `SQLSERVER`: reuse `Index.getClustering()` independently of uniqueness.
- `POSTGRESQL`: accept qualified collation/operator-class names and bare function keys. Store key attributes once in their structured fields so AST edits are reflected in SQL output.

Share key/option expression rendering between the AST and deparser, and route `CREATE INDEX` through the common definition traversal used by visitors, table discovery and validation. Preserve the default parser's single-type and raw key-parameter behavior, including MySQL prefix lengths.

Fixes #652. Addresses the index example in #1994 and the `CREATE UNIQUE NONCLUSTERED INDEX` failure in #1563. This does not resolve all failures in the Sakila scripts linked from #1563. A regression test also covers a PostgreSQL function followed by an index and another statement.

Validation:
- Gradle `check`: 6,272 tests, no failures, 25 skipped; formatting and static checks passed. The local formatter ratchet uses the upstream base commit because the fork's `origin/master` is behind upstream.
- Maven `verify` passed.
- Coverage includes dialect isolation, modifier combinations, AST edits, expression visitors/deparsers, validation and statement boundaries.

Syntax references: [Spanner](https://docs.cloud.google.com/spanner/docs/reference/standard-sql/data-definition-language#_create_index), [SQL Server](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-index-transact-sql?view=sql-server-ver17), [PostgreSQL](https://www.postgresql.org/docs/18/sql-createindex.html).

### DIFF
--- a/src/main/java/net/sf/jsqlparser/parser/AbstractJSqlParser.java
+++ b/src/main/java/net/sf/jsqlparser/parser/AbstractJSqlParser.java
@@ -46,7 +46,7 @@ public abstract class AbstractJSqlParser<P> {
                                                         AdjacentStringLiterals.WHITESPACE,
                                                         Feature.allowDoubleQuotedStrings,
                                                         Feature.allowBackslashEscapeCharacter), SNOWFLAKE(
-                                                                Feature.allowBackslashEscapeCharacter), INFORMIX;
+                                                                Feature.allowBackslashEscapeCharacter), INFORMIX, SPANNER;
 
         private final Set<Feature> lexerFeatures;
         private final AdjacentStringLiterals adjacentStringLiterals;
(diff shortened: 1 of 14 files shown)


manticore-projects:
thank you!

PR #2588, opened 2026-09-11
Split PostgreSQL hash operators from adjacent identifiers

### LINKED ISSUE #2163: [BUG] JSQLParser Version : 5.1   RDBMS : PostgreSQL 10   mix the JSON and relational operators, it outputs the wrong AST and sql.
### Description:
- When using a mix of JSON and relational operators, there is a priority error in the parsing results.
- The new sql lost some blank spaces. 

### SQL Example:
- Original SQL:
    ```sql
    select  *  from  t  where  js  #>>  '{a,b,1}' <> 'bar'
    ```
- Error 1:
   In the AST, the filter likes this:  ```js  #>>  ('{a,b,1}' <> 'bar')```
   but what we need is:  ```(js  #>>  '{a,b,1}') <> 'bar'```
 
- Error 2:
   JSQLParser generates the new sql:
   ```sql
    SELECT * FROM t WHERE js#>>'{a,b,1}' <> 'bar'
   ```
   but what we need is: ```SELECT * FROM t WHERE js #>> '{a,b,1}' <> 'bar'```
   JSQLParser lost the blank spaces.

### Software Information:
- JSqlParser version: **5.1**
- Database: **PostgreSQL**

minleejae:
With `Dialect.POSTGRESQL`, `js#>>'{a,b,1}' <> 'bar'` now has the same JSON-expression AST and precedence as the spaced form. Split unquoted identifiers at `#`, then let the existing lexer recognize `#>`, `#>>` or the standalone hash operator. Quoted identifiers and literal bodies retain their contents; default/SQL Server names and MySQL hash comments retain their behavior.

Share identifier truncation with hash-comment handling and update token end positions after rewinding. Regression tests cover spacing variants, qualified/quoted names, chained JSON operators, token offsets, statement boundaries and dialect isolation. Full Gradle check and Maven verify passed, using upstream base `eddb1fb5` for the local formatter ratchet.

Fixes #2163 (the remaining unspaced variant described in the comments; the original spaced query already works).

Reference: [PostgreSQL identifier and operator rules](https://www.postgresql.org/docs/18/sql-syntax-lexical.html).

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -1769,6 +1769,14 @@ TOKEN_MGR_DECLS : {
         token.endColumn = input_stream.getEndColumn();
     }
 
+    /** Splits a longest-match identifier without losing token positions. */
+    private void truncateIdentifierToken(Token token, int length) {
+        input_stream.backup(token.image.length() - length);
+        token.image = token.image.substring(0, length);
+        token.endLine = input_stream.getEndLine();
+        token.endColumn = input_stream.getEndColumn();
+    }
+
     /**
      * Consumes the body of a block comment after the opening delimiter has been matched,
      * honouring nesting, up to and including the outermost closing delimiter. Then backs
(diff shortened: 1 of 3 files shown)


manticore-projects:
Thank you much!

Issue #1946, opened 2024-01-15
[Feature] Prostgres Procedural `DO $$BEGIN ... END$$`


zdhuangelephant:
Always check against the **Latest SNAPSHOT of JSQLParser** and the [Syntax Diagram](https://jsqlparser.github.io/JSqlParser/syntax_snapshot.html)

### Failing SQL Feature:
- Brief description of the failing SQL feature
- Example: `Encountered unexpected token: "DO" "DO"` can't be parsed

### SQL Example:
- Simplified Query Example, focusing on the failing feature
    ```sql
DO $$
BEGIN
	IF NOT EXISTS( select 1 from comm.permission_operation where permission_operation_code = 'ecg_report_time_modify') and EXISTS( select 1 from comm.permission where permission_code = 'data_modify') 
	 THEN		
		INSERT INTO comm.permission_operation 
		  (permission_operation_id,
		   permission_id,
		   permission_operation_code,
		   permission_operation_name,
		   "type",
		   "version",
		   his_org_id,
		   his_creater_id,
		   his_creater_name,
		   his_create_time,
		   his_updater_id,
		   his_update_time)
		   VALUES
			 ((select max(permission_operation_id) + 1 from comm.permission_operation),
			 (select permission_id from comm.permission where permission_code = 'data_modify' limit 1),
			 'ecg_report_time_modify',
			 '心电报告时间修改',
			 '1',
			 0,
			 (select his_org_id from comm.hospital limit 1),
			 1,
			 '系统管理员',
			 now(),
			 1,
			 now()) on conflict(permission_operation_id) do nothing;
	END IF;
END $$;
    ```

### Software Information:
- JSqlParser 4.7
- Database PostgreSQL

`Exception in thread "main" net.sf.jsqlparser.JSQLParserException: net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: "DO" "DO"
    at line 1, column 1.`

zdhuangelephant:
`DO $$
BEGIN
	IF NOT EXISTS( select 1 from comm.permission_operation where permission_operation_code = 'ecg_report_time_modify') and EXISTS( select 1 from comm.permission where permission_code = 'data_modify') 
	 THEN		
		INSERT INTO comm.permission_operation 
		  (permission_operation_id,
		   permission_id,
		   permission_operation_code,
		   permission_operation_name,
		   "type",
		   "version",
		   his_org_id,
		   his_creater_id,
		   his_creater_name,
[...]

PR #2589, opened 2026-09-11
Support PostgreSQL DO wrappers with preserved routine body literals

minleejae:
With `Dialect.POSTGRESQL`, `DO` now parses into a dedicated `DoStatement` that preserves its body literal, dollar-quote tag, optional language and language-clause position. The wrapper uses shared rendering for the model and deparser; visitors can inspect or replace the body literal, and validation exposes a PostgreSQL `doStatement` capability.

This is the wrapper/body-preservation stage of #1946. It does not parse the procedural language inside the string into an INSERT/IF/loop AST. Feature analysis reports the statement as opaque, and table discovery explicitly reports unsupported analysis rather than claiming an empty set of tables.

Tests cover both LANGUAGE positions, tagged/untagged dollar and single-quoted bodies, a representative issue block followed by another statement, AST edits, visitor/deparser customization, dialect isolation, capability validation and malformed wrappers. Full Gradle check and Maven verify passed (local formatter ratchet: upstream base `eddb1fb5`).

Refs #1946.

Syntax: [PostgreSQL DO](https://www.postgresql.org/docs/18/sql-do.html).

### DIFF
--- a/src/main/java/net/sf/jsqlparser/parser/feature/Feature.java
+++ b/src/main/java/net/sf/jsqlparser/parser/feature/Feature.java
@@ -457,7 +457,7 @@ public enum Feature {
      */
     oracleBlock,
 
-    execute, executeExec, executeCall, executeExecute,
+    execute, executeExec, executeCall, executeExecute, doStatement,
 
     /**
      * SQL "EXECUTE" statement is allowed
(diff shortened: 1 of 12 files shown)


manticore-projects:
This is big! Thank you so much!

PR #2594, opened 2026-09-11
Align empty input handling for statement parsing

### LINKED ISSUE #2576: parseStatements returns null instead of throwing JSQLParserException when withAllowComplexParsing(false) and the input does not parse
`CCJSqlParserUtil.parseStatements(sql, consumer)` returns `null` instead of throwing `JSQLParserException` when complex parsing is switched off and the input does not parse. A caller that relies on the declared exception dereferences the result and fails with a `NullPointerException` of its own; a guard that treats "no exception" as "parsed" is misled.

**Reproduce**

```java
// default configuration: the documented exception
CCJSqlParserUtil.parseStatements("PRAGMA foo");
// -> JSQLParserException: Encountered: <S_IDENTIFIER> / "PRAGMA", at line 1, column 1 ...

// complex parsing off: null, no exception
Statements s = CCJSqlParserUtil.parseStatements("PRAGMA foo", p -> p.withAllowComplexParsing(false));
// s == null

// same for a broken script and for an empty statement between separators
CCJSqlParserUtil.parseStatements("SELECT * FROM foo; PRAGMA x", p -> p.withAllowComplexParsing(false)); // null
CCJSqlParserUtil.parseStatements("SELECT * FROM foo;;",         p -> p.withAllowComplexParsing(false)); // null

// the single-statement entry point throws as expected with the same option
CCJSqlParserUtil.parse("PRAGMA foo", p -> p.withAllowComplexParsing(false));
// -> JSQLParserException
```

Also observed: `parseStatements("/* nothing */")` (default configuration) returns an empty `Statements` while `parse("/* nothing */")` throws `JSQLParserException`; the two entry points disagree on comment-only input.

**Where**

`CCJSqlParserUtil.parseStatements(String, ExecutorService, Consumer)` (`CCJSqlParserUtil.java:471-501`): the simple-parse attempt fails, the complex re-parse is skipped because `allowComplexParsing` is false, and the method falls through returning the unassigned result instead of rethrowing the first `ParseException`.

**Suggestion**

Rethrow the first parse failure as `JSQLParserException` when the complex re-parse is disabled (the way `parse(String, ExecutorService, Consumer)` does), and align the two entry points on comment-only input, either both empty or both throwing.

minleejae:
Null and empty strings currently return `null` from the String `parse` and `parseStatements` overloads, even though blank and comment-only input already follows the single-statement versus statement-list contract. Make `parse("")` and `parse((String) null)` throw `JSQLParserException`, and make `parseStatements` return a fresh, mutable empty `Statements` list for either input.

Share single-statement input validation and remove the redundant guard in the default statement-list overload. Apply the contract to the configuration-callback and caller-executor overloads, preserving executor ownership. Document the intentional change from the previous null returns.

Regression coverage includes null, empty, blank and comment-only input with complex parsing enabled/disabled, Reader/InputStream consistency, caller-executor reuse after parsing, and independent mutable empty results.

Validation:
- The new regression suite failed on the unchanged implementation (7 failures) and passes with the fix.
- Full Gradle `check`, including formatting, static analysis and coverage: passed.
- Maven `clean verify`: passed.

Fixes #2576.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/parser/CCJSqlParserUtil.java
+++ b/src/main/java/net/sf/jsqlparser/parser/CCJSqlParserUtil.java
@@ -60,6 +60,13 @@ public final class CCJSqlParserUtil {
         return statement;
     }
 
+    /**
+     * Parses a single SQL statement.
+     *
+     * @param sql the SQL statement to parse
+     * @return the parsed statement
+     * @throws JSQLParserException if the input is null, empty, or cannot be parsed
+     */
     public static Statement parse(String sql) throws JSQLParserException {
         return parse(sql, null);
     }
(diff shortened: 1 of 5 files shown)


manticore-projects:
Thank you for this fine tuning! Good job!

PR #2595, opened 2026-09-11
Support opt-in legacy MySQL GROUP BY ordering

### LINKED ISSUE #1169: net.sf.jsqlparser.JSQLParserException: Encountered unexpected token: "desc" "DESC"
**Describe the bug**
parser error.

**To Reproduce**
1. Example SQL : select a from b group by c desc;
3. Exception : JSQLParserException

**Expected behavior**
parser success.

**System**
 - mysql 5.7
- Java 1.8
- JSqlParser 4.0

minleejae:
MySQL versions before 8.0.13 allow explicit `ASC`/`DESC` on `GROUP BY` items. Parse and preserve this syntax only when callers select `Dialect.MYSQL` and opt in with `withLegacyMySqlGroupBy(true)`; default and other-dialect parsing keep rejecting it.

Keep the existing grouping expression list and expose directions by list position, so visitors still see the original expressions. Share clause rendering between `GroupByElement` and `GroupByDeParser`, preserve directions when appending items, and validate ordering with a separate `selectGroupByOrdering` feature. Document the option and AST contract.

Validation: full Gradle `check` and Maven `verify`; regression tests cover dialect/option isolation, mixed explicit and omitted directions, round trips, custom expression deparsing, column traversal, mutable AST operations, and malformed input.

Integration validation: all twelve existing PRs plus these three independent fixes pass full Gradle `check` together (6,495 tests; 0 failures/errors; 25 skipped). Existing dialect-enum, shared-index-helper and documentation conflicts were resolved only in the local validation branch.

Fixes #1169.

Syntax reference: https://dev.mysql.com/doc/refman/8.0/en/select.html (the extension was removed in 8.0.13).

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,9 @@ out. `is()` answers "did the grammar prove it", `may()` answers "could it be rul
 guard uses `may()` and a dispatcher uses `is()`. Function volatility is not a syntactic property,
 so anything the caller has not declared pure stays unproven and is listed by name.
 
+Legacy MySQL `GROUP BY ... ASC/DESC` is available with `Dialect.MYSQL` and the explicit
+`withLegacyMySqlGroupBy(true)` option; modern/default parsing keeps it disabled.
+
 ## Piped SQL
 
 Support is progressing for Piped SQL, which writes queries in the order they actually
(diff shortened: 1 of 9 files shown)


manticore-projects:
The MySQL folks will be happy! Thank you!

PR #2596, opened 2026-09-11
Classify ODBC timestamp interval arguments as time units

### LINKED ISSUE #1139: support for (date({fn timestampadd(SQL_TSI_YEAR, 2, date("travel_date"))}))
Hi Want to know,
do we have support for expressions like (date({fn timestampadd(SQL_TSI_YEAR, 2, date("travel_date"))}))
i want to extract column name from the given expression i.e travel_date
whenever i parse this expression SQL_TSI_PARSER is also getting considered as column name

minleejae:
In `{fn TIMESTAMPADD(SQL_TSI_YEAR, 2, travel_date)}`, the interval keyword is currently visited as a column. Classify the nine standard `SQL_TSI_*` keywords as `DateUnitExpression` values in the first argument of unqualified ODBC-escaped `TIMESTAMPADD`/`TIMESTAMPDIFF` calls with three arguments.

Both functions share the same normalization and existing time-unit visitor/deparser support. Output retains the ODBC keyword. Ordinary calls, qualified function or argument names, quoted identifiers, unknown units, and other argument positions retain their existing interpretation; this does not infer DB-specific semantics for arbitrary function names.

Validation: full Gradle `check` and Maven `clean verify`; regression coverage includes all nine units in both functions and parsing modes, round trips, column traversal, AST nodes, nested escapes, case handling, and the scope boundaries above.

Integration validation: all twelve existing PRs plus these three independent fixes pass full Gradle `check` together (6,495 tests; 0 failures/errors; 25 skipped). Existing dialect-enum, shared-index-helper and documentation conflicts were resolved only in the local validation branch.

Fixes #1139.

ODBC reference: https://learn.microsoft.com/en-us/sql/odbc/reference/appendixes/time-date-and-interval-functions

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,9 @@ Background reading: the [Google research paper](https://storage.googleapis.com/g
 [BigQuery pipe syntax](https://cloud.google.com/bigquery/docs/reference/standard-sql/pipe-syntax)
 and [DuckDB FROM-first syntax](https://duckdb.org/docs/sql/query_syntax/from.html#from-first-syntax).
 
+ODBC `{fn TIMESTAMPADD(...)}` and `{fn TIMESTAMPDIFF(...)}` expose standard
+`SQL_TSI_*` interval arguments as time-unit expressions, preserving column traversal.
+
 ## Java version
 
 | JSqlParser | Runtime | Notes |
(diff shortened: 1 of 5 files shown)


manticore-projects:
Thanks!

PR #2605, opened 2026-09-11
Support SQL Server SET IDENTITY_INSERT with a structured table target

minleejae:
With `Dialect.SQLSERVER`, `SET IDENTITY_INSERT [database.][schema.]table ON | OFF` now parses into `SetIdentityInsertStatement`. Its editable target reuses the existing `Table` AST, including qualified and bracket-quoted names, and its ON/OFF value is separate from assignment expressions.

Both SQL renderers share statement formatting while table visitors retain context. Table discovery and metadata validation use the structured target, and feature analysis classifies the directive as a session setting. Generic SET assignments and other dialects retain their existing behavior.

Validation: full Gradle `check` passed (6,563 tests, 0 failures/errors, 25 skipped), covering AST edits, custom visitors/deparsers, metadata, invalid syntax, statement boundaries, and dialect isolation. The original Sakila insert-data script parses and round-trips all 46,326 statements, including 46,273 INSERTs and 26 IDENTITY_INSERT directives. The updated usage page builds with Sphinx warnings treated as errors.

Refs #1563. Syntax reference: [Microsoft SET IDENTITY_INSERT](https://learn.microsoft.com/en-us/sql/t-sql/statements/set-identity-insert-transact-sql?view=sql-server-ver17).

### DIFF
--- a/src/main/java/net/sf/jsqlparser/parser/feature/Feature.java
+++ b/src/main/java/net/sf/jsqlparser/parser/feature/Feature.java
@@ -644,6 +644,8 @@ public enum Feature {
      * @see UseStatement
      */
     use,
+    /** SQL Server session permission to insert explicit identity values into a table. */
+    setIdentityInsert,
     /**
      * @see Grant
      */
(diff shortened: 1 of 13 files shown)


manticore-projects:
Thank you!

PR #2606, opened 2026-09-12
fix: keep CREATE TABLE option tokens in sync with typed options

minleejae:
Editing a typed CREATE TABLE option (for example, changing ENGINE from InnoDB to MyISAM) currently changes `toString()` but leaves StatementDeParser using the old value. Derive legacy option tokens from the current typed options instead of retaining a stale copy, and document the typed-token snapshot API.

Regression tests cover editing, adding and clearing typed options, plus replacement through the mutable legacy string API.


manticore-projects:
Thank you!

PR #2607, opened 2026-09-12
fix: preserve ordered table elements through legacy list edits

minleejae:
Editing CREATE TABLE column/index lists can leave the ordered table definition unchanged, and replacing the column list can silently remove PostgreSQL LIKE clauses. Back the legacy lists with mutable filtered views of the ordered elements and replace only elements of the requested type, preserving other declarations and their positions.

Regression tests cover list mutations in both directions, replacement and fluent APIs, LIKE/table traversal preservation, index bounds, and consistent rendering after clearing the definition.


manticore-projects:
Thank you!

PR #2608, opened 2026-09-12
fix: preserve structured column options when editing specifications

minleejae:
Clearing structured column options can restore stale raw IDENTITY text, while appending legacy column specifications can discard structured REFERENCES information and remove referenced tables from traversal. Keep only the active option representation and preserve existing structured or raw options when appending through either API.

Regression tests cover identity removal, legacy appends retaining foreign-key traversal, promotion of raw specifications, and explicit replacement/clearing.


manticore-projects:
Thank you!

PR #2609, opened 2026-09-12
fix: share expression-aware table constraint rendering

minleejae:
ALTER TABLE ADD CHECK and inline functional indexes currently bypass a custom ExpressionDeParser even though their expressions are structured AST nodes. Share callback-aware index/constraint rendering between model `toString()` methods and TableElementDeParser, and route ALTER ADD index definitions through the same renderer.

The shared rendering retains foreign-key references, constraint attributes, index comments and storage options. Tests cover expression rewrites in CHECK, functional indexes and EXCLUDE, including programmatically constructed ALTER constraints, plus unchanged default rendering across constraint forms.


manticore-projects:
Thank you!

PR #2610, opened 2026-09-12
fix: traverse and deparse structured table partitions

minleejae:
TablesNamesFinder omits the child table in PostgreSQL ATTACH/DETACH PARTITION and the exchange table in MySQL EXCHANGE PARTITION. Extend shared table-definition traversal to those references and to structured partition keys and PostgreSQL bounds.

Reuse callback-aware rendering for CREATE/ALTER partition keys and bounds, including subpartition keys, while preserving raw MySQL partition-definition values and options. Regression tests cover table discovery, visitor context, custom expression rendering, and unchanged SQL output. MINVALUE/MAXVALUE range markers are preserved and excluded from column-reference traversal.


manticore-projects:
Thank you!

PR #2611, opened 2026-09-12
feat: preserve ALTER COLUMN default expressions in the AST

minleejae:
ALTER COLUMN SET DEFAULT currently converts the parsed expression into a String, preventing expression traversal and custom deparsing. Retain the Expression, expose it for AST edits, and derive the existing getDefaultValue() text from it. The legacy String constructor remains available and opaque, including nullable calls.

Share ALTER COLUMN action rendering and route ALTER validation through the existing table-definition traversal. Regression tests cover arithmetic, casts, functions, NULL, multiple actions, visitor context, custom deparsing and AST edits.

This change covers ALTER COLUMN SET DEFAULT; CREATE TABLE column defaults retain their existing representation.


manticore-projects:
Thank you!

PR #2613, opened 2026-09-12
refactor: share PostgreSQL routine argument grammar

minleejae:
Privilege targets and extension members build the same RoutineReference.Argument AST through duplicated grammar. Share mode/name/type parsing while retaining the existing parser entry points and keeping ORDER BY argument boundaries restricted to aggregate-signature contexts.

Tests compare argument ASTs across GRANT and ALTER EXTENSION for modes, qualified and array types, and quoted argument names. They also cover ordered-set aggregates and rejection of aggregate-only syntax in FUNCTION/GRANT contexts.


manticore-projects:
Thank you!

PR #2617, opened 2026-09-12
Support SQL large-object names through shared column type parsing

minleejae:
EvoMaster works around `CHARACTER LARGE OBJECT` in H2 conditions by replacing the type name with `VARCHAR` before parsing. Its [conversion code](https://github.com/WebFuzzing/EvoMaster/blob/ab59ec974c195cdb2cef11f69574a559201d2202/core-extra/dbconstraint/src/main/java/org/evomaster/dbconstraint/parser/jsql/JSqlConditionParser.java#L121) notes that a global replacement can also change literal contents. The original condition with `CAST('hi' AS CHARACTER LARGE OBJECT(2))` currently fails at `LARGE`.

Add a shared large-object type-name production to `ColDataType`, preserving the complete name and existing length properties across CAST, CREATE/ALTER column definitions and `parseColDataType`. It handles CHARACTER/CHAR/NCHAR LARGE OBJECT, NATIONAL CHARACTER LARGE OBJECT and BINARY LARGE OBJECT using contextual tokens and the existing suffix parser. These names are documented in the [H2 type reference](https://www.h2database.com/html/datatypes.html#character_large_object_type).

Tests cover the downstream condition, structured type and length fields, expression traversal, round trips, comments/whitespace, unchanged literal contents and nearby existing type names. No source-text preprocessing is needed.

Validation: Java 17 `./gradlew --console=plain --max-workers=2 check`, including the complete test suite, JavaCC choice-conflict gate, Checkstyle, PMD, SpotBugs and Spotless.

### DIFF
--- a/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
+++ b/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt
@@ -1712,6 +1712,21 @@ public class CCJSqlParser extends AbstractJSqlParser<CCJSqlParser> {
         }
     }
 
+    private boolean isLargeObjectTypeAhead() {
+        int offset = 1;
+        if (isKeywordAhead("NATIONAL")) {
+            offset++;
+            if (!"CHARACTER".equalsIgnoreCase(getToken(offset).image)) {
+                return false;
+            }
+        }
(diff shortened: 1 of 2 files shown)


manticore-projects:
Thank you for supporting H2!

PR #2619, opened 2026-09-12
Support structured XMLFOREST arguments with aliases

minleejae:
XMLFOREST arguments with aliases, such as `XMLFOREST(customer.name AS "Name", balance AS "Balance")`, currently fail to parse. Jailer works around this by replacing unsupported XML expressions during SQL analysis ([pinned consumer code](https://github.com/Wisser/Jailer/blob/6f1531a4556eb9ca2215db6482596fbbb73071e2/src/main/engine/net/sf/jailer/util/SqlUtil.java#L636)). Preserving the original expressions and aliases lets consumers inspect and rewrite these queries without discarding their structure.

This change implements the [XMLFOREST argument syntax](https://www.postgresql.org/docs/current/functions-xml.html#FUNCTIONS-XML-XMLFOREST) using the existing `Function` parameter list and a reusable `AliasedExpression`. Unaliased arguments retain their existing expression types. Default visitor forwarding preserves compatibility with existing visitors, and a shared renderer keeps model output and custom expression deparsers consistent. Table discovery also traverses aliased arguments, including nested subqueries.

Validation: the full Java 17 Gradle `check` passes, including the zero-conflict JavaCC grammar gate, tests, formatting, Checkstyle, PMD, and SpotBugs. Regression tests cover mixed and nested arguments, quoted aliases, AST mutation, visitor context, custom rendering, table discovery, validation, round trips, legacy XML forms, and rejection of malformed aliases.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/expression/AliasedExpression.java
+++ b/src/main/java/net/sf/jsqlparser/expression/AliasedExpression.java
@@ -1,0 +1,59 @@
+/*-
+ * #%L
+ * JSQLParser library
+ * %%
+ * Copyright (C) 2004 - 2026 JSQLParser
+ * %%
+ * Dual licensed under GNU LGPL 2.1 or Apache License 2.0
+ * #L%
+ */
+package net.sf.jsqlparser.expression;
+
+import java.util.function.Consumer;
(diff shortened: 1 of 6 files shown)


manticore-projects:
Thank you much!

Issue #2516, opened 2026-08-31
[BUG] JSQLParser 5.4-SNAPSHOT : all RDBMS : TablesNamesFinder silently misses tables in many SELECT and DML clauses (INTO, GROUP BY, QUALIFY, WINDOW, ORDER BY, LIMIT, DISTINCT ON, LATERAL VIEW, PIVOT XML, SETTINGS)


fudianchn:
AI disclosure: this issue was prepared with AI coding agents, reviewed and revised line by line by me.

### Failing SQL Feature:

`TablesNamesFinder` silently misses tables that only appear inside SELECT clauses it never traverses. `SelectVisitorAdapter` covers most of these clauses, `TablesNamesFinder` does not. All shapes below parse fine on master `e7167cc7` (5.4-SNAPSHOT) and return an incomplete table set:

| Clause | Example | Returned | Missed |
|---|---|---|---|
| INTO target | `SELECT * INTO t1 FROM t2` | `[t2]` | `t1` |
| INTO TEMP | `SELECT * FROM t2 INTO TEMP t1` | `[t2]` | `t1` |
| GROUP BY | `SELECT a FROM t GROUP BY a, (SELECT b FROM u)` | `[t]` | `u` |
| GROUPING SETS | `SELECT a FROM t GROUP BY GROUPING SETS ((a, (SELECT b FROM u)), (c))` | `[t]` | `u` |
| QUALIFY | `SELECT * FROM t QUALIFY (SELECT b FROM u) = 1` | `[t]` | `u` |
| WINDOW | `SELECT * FROM t WINDOW w AS (PARTITION BY (SELECT b FROM u))` | `[t]` | `u` |
| ORDER BY | `SELECT a FROM t ORDER BY a, (SELECT b FROM u)` | `[t]` | `u` |
| LIMIT / LIMIT BY / OFFSET / FETCH | `SELECT a FROM t LIMIT (SELECT b FROM u)` etc. | `[t]` | `u` |
| DISTINCT ON | `SELECT DISTINCT ON ((SELECT b FROM u)) a FROM t` | `[t]` | `u` |
| LATERAL VIEW | `SELECT * FROM t LATERAL VIEW explode(array((SELECT b FROM u))) v AS x` | `[t]` | `u` |
| SETTINGS | `SELECT a FROM t SETTINGS x = (SELECT b FROM u)` | `[t]` | `u` |
[...]

PR #2620, opened 2026-09-12
Expose structured ORDER BY fragment parsing

minleejae:
Consumers that assemble queries from independent sort specifications currently need to wrap fragments in a SELECT or retain them as opaque expressions. For example, DBeaver's ORDER BY handling uses `CustomExpression` for a sort fragment ([pinned consumer code](https://github.com/dbeaver/dbeaver/blob/b59edd36ff94abfea48f5151a5ea10b3791137f0/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/parser/SQLSemanticProcessor.java#L302)). A fragment API exposes the expression and sort flags directly, so callers can inspect and rewrite them through the ordinary AST.

Add `CCJSqlParserUtil.parseOrderByElements(String)` and its parser-configuration overload. Input such as `COALESCE(score, 0) DESC NULLS LAST, id ASC` produces a `List<OrderByElement>`; the input excludes the ORDER BY keywords. Null or empty input returns an empty list, and trailing tokens are rejected.

The existing ORDER BY clause now delegates to a shared element-list production. Column-type parsing and the new API also share fragment parser setup, fast/complex retry, full-input checking, and exception handling. Existing column-type behavior is preserved, and each invocation starts with fresh parser state.

Validation: full Java 17 Gradle `check` passes, including the zero-conflict JavaCC grammar gate, tests, formatting, Checkstyle, PMD, and SpotBugs. Tests compare fragments with complete ORDER BY clauses, exercise configured quoting and parameter indices, verify custom expression deparsing and round trips, and reject incomplete or trailing SQL.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/parser/CCJSqlParserUtil.java
+++ b/src/main/java/net/sf/jsqlparser/parser/CCJSqlParserUtil.java
@@ -13,6 +13,8 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
 import java.util.Stack;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -31,6 +33,7 @@ import net.sf.jsqlparser.parser.feature.Feature;
 import net.sf.jsqlparser.statement.Statement;
 import net.sf.jsqlparser.statement.Statements;
 import net.sf.jsqlparser.statement.create.table.ColDataType;
+import net.sf.jsqlparser.statement.select.OrderByElement;
 
 /**
  * Toolfunctions to start and use JSqlParser.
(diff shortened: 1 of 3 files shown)


manticore-projects:
Mhhh, I don't like this. I believe I do understand the requirement: simpler access to certain fragments, the walking the AST is cumbersome. But instead of specific solutions, I wonder if a kind of "querying" might be the better and more comprehensive solution here. Like XSL for an XML, something like `*/WHERE[1]/EXPRESSION[*]`. In fact, I drafted something crude like this in JSQLFormatter years ago.

Suggestion: lets put this on small flame for a while and focus on getting a 5.4 release out. And then craft a proper AST querying mechanism for this together.
(Unless it was really urgent for you, then we would just take that in and deprecate it later. Just let me know.)

minleejae:
@manticore-projects 
This isn't urgent for me. I agree that a general-purpose AST querying mechanism would be more useful in the long term, and I'd be happy to work on that together. Let's leave this PR on hold and focus on the 5.4 release first.

minleejae:
@manticore-projects Following up on the AST querying idea, I compared 59 DDL/expression cases on 5.4 and current master (`6312f9e5`). Some consumer-side scanning comes from information that is still unstructured:

- `ColumnOption` already structures defaults, references and identity declarations, but column `COLLATE`, `COMMENT`, `ON UPDATE`, generated expressions and several flags remain raw tokens.
- `CURRENT_TIMESTAMP`, `CURRENT_TIMESTAMP(6)` and bare `LOCALTIME` produce `TimeKeyExpression`, `Function` and `Column` respectively. A common temporal representation needs to preserve dialect rules and quoted/qualified identifier behavior.
- `DROP INDEX x ON app.t` stores the owner in `Drop.getParameters()` as `[ON, app.t]`. There is no structured owner table.
- `ColDataType` exposes precision/scale, but `getDataType()` can still contain the parameters, e.g. `decimal (10, 2)`, so extracting the base type still requires string handling.

These seem complementary to AST querying: enriching the model would give the querying layer structured fields to traverse. I suggest handling them in focused PRs that preserve existing accessors and rendering, while keeping general selection/traversal design in this discussion. The quoted-name and negative-scale parser fixes are already in #2638.

Does that separation fit your plan, or are there node/property conventions you would like those AST additions to follow for the future querying API?

manticore-projects:
I spent some tokens quickly to mock something:

*get me the where expression of the second union query of the 3rd with clause* becomes a simple path `/withItemsList[2]/select/selects[1]/where`:

```java
Expression w = AstPath.get(root, "/withItemsList[2]/select/selects[1]/where", Expression.class);
```

Three variants, all verified against the mock:

`/withItemsList[2]/select/selects[1]/where` — positional, your literal phrasing
`/withItemsList[name='w2']/select/selects[1]/where` — by CTE name, survives reordering
`//SetOperationList/selects[1]/where` — shape-agnostic, doesn't care how deep the union sits

Main problem is: our path expressions would be unstable. Any API change and a formerly working path possible won't work. Also there is no schema to verify a path against. Maybe we would need to create such schemas using JAXB. Also, there's no sibling axis, so "the union branch after the one matching X" isn't expressible. We would need to add following-sibling/preceding-sibling yet.

[AstPath.zip](https://github.com/user-attachments/files/32222319/AstPath.zip)

manticore-projects:
> Does that separation fit your plan, or are there node/property conventions you would like those AST additions to follow for the future querying API?

In my opinion, developing a generic `AST Path Query` mechanism and `enriching the API where needed` are two different things. DDL won't have much benefit from our `AST Query` because they are not nested (unless you really want to stress it). But Queries can be deeply nested structures and here `AST Path` can make a lot of sense.

fudianchn:
Observations on the path approach:

1. Positional steps are the fragile part: of the three variants above, `/withItemsList[2]/...` breaks on reordering, `[name='w2']` survives reordering, and `//SetOperationList` additionally survives new intermediate nodes. Node type + field predicates + descendant relationships should be the primary semantics, with positional indices as an optional narrowing operator.

2. Compiling paths to typed matchers would address the missing-schema problem without a generated schema. Parse and compile a path once at load time: unknown node or property names fail immediately instead of silently matching nothing. Most of the validation benefit, and there is no schema artifact to maintain in parallel with the AST.

3. With paths compiled to matchers, following-sibling/preceding-sibling becomes an extension of the matcher; the SQL grammar stays untouched.

4. The matcher layer could itself be a public typed traversal API, `findAll(PlainSelect.class, predicate)` with default full recursion: not a parallel entry point that the query language would later replace, but the layer it compiles to. The recurring TablesNamesFinder gaps (for example #2516) come from visitor methods missing recursion; default full recursion removes that failure mode.

5. One boundary to decide early: shape-agnostic matches such as `//Table` also hit CTE references, where no physical table exists. A scope-aware variant that does not descend into subqueries, or resolves CTE names first, keeps structural search usable for table-level questions.

manticore-projects:
> One boundary to decide early: shape-agnostic matches such as //Table also hit CTE references, where no physical table exists. A scope-aware variant that does not descend into subqueries, or resolves CTE names first, keeps structural search usable for table-level questions.

You can NOT solve that in JSQLParser alone (because there is no information on the actual schema). 
We have solved that correctly in `JSQLTranspiler` though, where we read the actual schema information and resolve tables and columns against it.

Maybe we need two solution instances: one approximation in JSQLParser, that returns "anything that looks like a table" and a precise one in JSQLTranspiler return actual tables only. (Similar to `TableNamesFinder`).

fudianchn:
Structural queries can stay in JSQLParser, with resolution against database metadata in JSQLTranspiler. By scope-aware in point 5, I meant binding CTE references to declarations within the SQL statement, without guaranteeing that the remaining table references identify actual database tables.

manticore-projects:
@minleejae: Would you mind closing this and we open a Discussion for it? Unfortunately Github does not support transfer into a discussion automatically.

minleejae:
Moved the design discussion to [Discussion #2665: AST path queries and typed traversal](https://github.com/JSQLParser/JSqlParser/discussions/2665), as requested. It summarizes the proposals and open questions, with links to the prototype and original comments.

Closing this PR so we can continue the design discussion there. Thanks for the feedback.

PR #2621, opened 2026-09-12
Preserve ASOF joins through shared join rendering

minleejae:
Stratum currently rewrites ASOF joins before handing SQL to JSqlParser ([pinned rewrite code](https://github.com/replikativ/stratum/blob/91a436694248dc6740fb5b7bb9b38d0cc7465919/src/stratum/sql/rewrite.clj#L1)). Parsing this syntax directly preserves the distinction between an ordinary join and a nearest-match join, which consumers need to inspect or transform temporal queries without losing their meaning.

Add structured support for DuckDB-style `ASOF JOIN`, including explicit INNER and directional OUTER forms, with the existing ON expressions and USING columns. For example, `trades t ASOF LEFT JOIN prices p ON t.symbol = p.symbol AND t.ts >= p.ts` now retains the ASOF modifier in `Join.isAsOf()` and in both rendering paths. ASOF is contextual, so ordinary identifiers and explicit aliases remain available. The [DuckDB syntax documentation](https://duckdb.org/docs/stable/sql/query_syntax/from#as-of-joins) describes the nearest-match behavior; expression validity and database execution semantics remain the database's responsibility.

Refactor duplicated join-modifier rendering into `Join.appendJoinTypeTo`, shared by `Join.toString()` and `SelectDeParser`. Existing hints, FETCH, ordinary qualifiers, and expression visitor traversal continue through their existing paths. Add a separate validation feature so callers can allow or reject ASOF independently of ordinary joins.

Validation: full Java 17 Gradle `check` passes, including the zero-conflict JavaCC grammar gate, tests, formatting, Checkstyle, PMD, and SpotBugs. Regression coverage includes unaliased sources, subqueries, chained and pipe joins, USING column structure, AST mutation, custom expression rendering, table discovery, feature validation, and ordinary ASOF identifiers.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/parser/feature/Feature.java
+++ b/src/main/java/net/sf/jsqlparser/parser/feature/Feature.java
@@ -195,6 +195,9 @@ public enum Feature {
      */
     joinApply,
 
+    /** Nearest-match ASOF join. */
+    joinAsOf,
+
     joinWindow, joinUsingColumns,
 
     /**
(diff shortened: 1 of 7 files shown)


manticore-projects:
Kudos for supporting DuckDB! Thank you!

PR #2622, opened 2026-09-12
Structure FOR PORTION intervals in UPDATE and DELETE

minleejae:
Temporal SQL consumers need to retain the interval affected by an UPDATE or DELETE. Stratum currently extracts and removes `FOR PORTION OF` before parsing, then carries that information separately ([pinned consumer workaround](https://github.com/replikativ/stratum/blob/91a436694248dc6740fb5b7bb9b38d0cc7465919/src/stratum/sql/rewrite.clj#L449)). A structured clause keeps the period and bound expressions attached to the statement, making them available to inspection, rewriting, and validation.

Support application-time `FOR PORTION OF period FROM expression TO expression` on UPDATE and DELETE targets, as documented for [MariaDB application-time periods](https://mariadb.com/docs/server/reference/sql-structure/temporal-tables/application-time-periods). For example, `UPDATE prices FOR PORTION OF valid_time FROM ? TO ? SET price = ?` now preserves the period name, both expressions, and their parameter ordering.

Both statements reuse `ForPortionClause`, one grammar production, shared deparser handling, and shared validation. Its renderer is also used by model output, while its expression traversal carries visitor context to both bounds. Table discovery visits nested bound expressions, and a dedicated validation feature is included in the MariaDB feature set. This change concerns the DML interval clause; it does not add temporal SELECT syntax or enforce database-specific restrictions on the values of the bounds.

Validation: full Java 17 Gradle `check` passes, including the zero-conflict JavaCC grammar gate, tests, formatting, Checkstyle, PMD, and SpotBugs. Tests cover both statements, aliases and quoted period names, typed literals and parameters, AST mutation, custom deparsers, nested-expression traversal, feature validation, WHERE/ORDER BY/LIMIT/RETURNING boundaries, round trips, and malformed clauses.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/parser/feature/Feature.java
+++ b/src/main/java/net/sf/jsqlparser/parser/feature/Feature.java
@@ -337,6 +337,9 @@ public enum Feature {
      * @see Update
      */
     update,
+
+    /** Application-time FOR PORTION OF in UPDATE and DELETE. */
+    forPortion,
     /**
      * "UPDATE table1 SET ... FROM table2
      */
--- a/src/main/java/net/sf/jsqlparser/statement/ForPortionClause.java
+++ b/src/main/java/net/sf/jsqlparser/statement/ForPortionClause.java
@@ -1,0 +1,72 @@
+/*-
+ * #%L
+ * JSQLParser library
+ * %%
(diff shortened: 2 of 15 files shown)


manticore-projects:
Thank you!

PR #2623, opened 2026-09-12
Validate every UPDATE assignment through shared UpdateSet traversal

minleejae:
`UPDATE t SET a = 1, b = ?` currently passes validation even when JDBC parameters are forbidden. The validator reads deprecated accessors that expose only the first assignment, so equivalent expressions are checked differently depending on their position.

Validate the complete `UpdateSet` list and share assignment traversal with MERGE. Columns and values in every assignment are visited; direct SELECT values use the select validator. This also removes the obsolete first-assignment/select compatibility branch without changing the public UPDATE AST API.

Validation: full Java 17 Gradle `check` passes, including tests, formatting, Checkstyle, PMD, SpotBugs, coverage, and the grammar ambiguity gate. Regression tests cover first/middle/last assignments, tuples, subqueries, independent feature errors across assignments, and the existing MERGE path.


manticore-projects:
Thank you!

PR #2624, opened 2026-09-12
Render DML RETURNING expressions through shared visitor handling

minleejae:
A custom expression deparser changes the value in `UPDATE t SET a = 7 RETURNING a + 8` but leaves the RETURNING expression untouched. INSERT and DELETE have the same inconsistency, preventing consumers from applying one expression transformation across an entire DML statement.

Route the three RETURNING clauses through a shared deparser helper that uses the existing callback renderer and select-item visitor. Expressions, aliases, wildcards, and output targets retain the existing AST and SQL layout; nested queries use the configured expression/SELECT visitors.

Validation: full Java 17 Gradle `check` passes, including tests, formatting, Checkstyle, PMD, SpotBugs, coverage, and the grammar ambiguity gate. Tests verify one callback per expression, nested SELECTs, preserved aliases and OLD/NEW qualifiers, INTO targets, unchanged source ASTs, and round trips.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/util/deparser/AbstractDeParser.java
+++ b/src/main/java/net/sf/jsqlparser/util/deparser/AbstractDeParser.java
@@ -9,6 +9,7 @@
  */
 package net.sf.jsqlparser.util.deparser;
 
+import net.sf.jsqlparser.statement.ReturningClause;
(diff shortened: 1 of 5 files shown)


manticore-projects:
Thank your for fixing this!

PR #2625, opened 2026-09-12
Reuse common JOIN traversal when discovering UPDATE tables

minleejae:
`UPDATE target t JOIN source s ON s.id IN (SELECT id FROM hidden) SET t.a = 1` currently reports only `target` and `source` through `TablesNamesFinder`. The UPDATE-specific join loop visits the right-hand table but omits the ON expressions, unlike SELECT and DELETE.

Reuse the existing common JOIN traversal for both UPDATE join positions. This removes duplicate loops and includes tables inside join predicates while retaining CTE filtering and visitor context. Consumers inspecting query dependencies can now see the same referenced tables across equivalent statement forms.

Validation: full Java 17 Gradle `check` passes, including tests, formatting, Checkstyle, PMD, SpotBugs, coverage, and the grammar ambiguity gate. Tests cover joins before SET and after FROM, outer joins, predicate subqueries, CTEs, context propagation, and one visit per source.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/util/TablesNamesFinder.java
+++ b/src/main/java/net/sf/jsqlparser/util/TablesNamesFinder.java
@@ -1393,11 +1393,7 @@ public class TablesNamesFinder<Void>
             visit(update.getTable(), context);
         }
 
-        if (update.getStartJoins() != null) {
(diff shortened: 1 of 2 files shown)


manticore-projects:
Thank you for fixing this!

PR #2626, opened 2026-09-12
Share parser execution and preserve cancellation signals

minleejae:
Waiting for a parser task currently swallows the caller’s interruption and leaves the submitted task running. The single-statement and statement-list utilities duplicate the same execution code, so both paths have this behavior.

This change shares the executor submission/wait logic, cancels the task when the caller is interrupted, restores the caller’s interrupt flag, and makes the parser stop flag visible to the worker thread. Timeout, interruption, and cancellation stop the convenience APIs from submitting a complex-parser retry; syntax errors retain the existing retry path. The supplied executor remains caller-owned and public method signatures are unchanged.

Validation: `./gradlew check` passed, including deterministic tests for both entry points, interruption, zero-timeout cancellation, retry counts, and reuse of a caller-owned executor after valid and invalid SQL.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/parser/CCJSqlParserUtil.java
+++ b/src/main/java/net/sf/jsqlparser/parser/CCJSqlParserUtil.java
@@ -14,6 +14,8 @@ import java.io.InputStream;
 import java.io.Reader;
 import java.util.Stack;
 import java.util.concurrent.Callable;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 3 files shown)


manticore-projects:
Thank you very much for improving this!

Btw, I am suggesting:

1) to release 5.4 stable first (I have asked @JSQLParser/owners already)
2) then remove the `SimpleExpression` completely, everything becomes an `Expression`
3) then remove the `Complex` vs `Simple` Parsing as well. 

It should not be needed anymore since we got the `PRATT parser` and the grammar fixed.

PR #2627, opened 2026-09-12
Protect shared dialect presets from external mutation

minleejae:
`Dialect.getLexerFeatures()` exposes the mutable set stored in a shared enum constant. A caller can therefore change the defaults of parsers created elsewhere in the application, making behavior depend on unrelated configuration code.

Expose an unmodifiable view of the preset defaults and document how `withDialect` applies enabled defaults. Parser-specific overrides remain local, and the existing additive application order is preserved.

Validation: `./gradlew check` passed. Regression tests cover mutation attempts on every preset, isolation between parser instances, successive dialect application, and explicit overrides.


manticore-projects:
Thank you!

PR #2628, opened 2026-09-12
Keep ALTER key accessors connected to structured index state

minleejae:
After parsing an ALTER PRIMARY KEY or UNIQUE action, the legacy key-column accessors can disagree with the structured `Index`: changing the index leaves a stale getter, and calling a legacy setter updates a separate list that does not affect generated SQL. Delegate matching key accessors to `Index.getColumnsNames()` / `setColumnsNames()` so inspection and explicit replacement use the same index. Legacy-only actions retain their existing list behavior.

The string API follows the existing `Index` replacement semantics:

- Index-backed getters return mutable snapshots. Editing a snapshot does not update the index.
- String setters and fluent additions rebuild every key as a plain `ColumnParams`. Even an unchanged getter/setter round trip preserves only rendered SQL, not structured expressions or sort options; `a DESC` becomes a plain rendered name. Use `Index.getColumns()` / `setColumns()` for edits that must preserve structured elements.
- Parsing attaches the structured index directly, without calling those string setters to initialize duplicate legacy lists. Parsed attributes therefore remain intact until an explicit string replacement.
- An index with null columns returns an empty snapshot, so getters and fluent additions work for newly constructed indexes.

Validation: `./gradlew check` passed (6,728 tests, 0 failures/errors, 25 skipped). Regression coverage includes PK and UNIQUE snapshots, setters and fluent additions, bare-name reordering, unchanged string round trips, structured reordering, initial parsed sort attributes, index metadata, null columns, and agreement between AST and deparser output.

### DIFF
--- a/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java
+++ b/src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java
@@ -11,6 +11,7 @@ package net.sf.jsqlparser.statement.alter;
 
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.AbstractList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java
```diff
@@ -505,20 +506,98 @@ public void setUsingIfExists(boolean usingIfExists) {
...
+    private static class KeyColumnNames extends AbstractList<String> {
+        private final Index index;
+
+        KeyColumnNames(Index index) {
+            this.index = index;
+        }
+
+        @Override
+        public String get(int position) {
+            return index.getColumns().get(position).toString();
```
manticore-projects: Isn't this a `String` already?!

src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java
```diff
@@ -505,20 +506,98 @@ public void setUsingIfExists(boolean usingIfExists) {
...
+                    : new ArrayList<>(index.getColumns());
+            columns.add(position, new Index.ColumnParams(name));
+            index.setColumns(columns);
+            modCount++;
+        }
+
+        @Override
+        public String remove(int position) {
+            List<Index.ColumnParams> columns = new ArrayList<>(index.getColumns());
+            String previous = columns.remove(position).toString();
```
manticore-projects: Isn't this a `String` already?!

src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java
```diff
@@ -505,20 +506,98 @@ public void setUsingIfExists(boolean usingIfExists) {
...
+    }
+
+    private void replaceKeyColumns(List<String> names) {
+        List<Index.ColumnParams> replacement = new ArrayList<>();
+        if (names != null) {
+            List<Index.ColumnParams> previous = index.getColumns();
+            for (int i = 0; i < names.size(); i++) {
+                String name = names.get(i);
+                replacement.add(previous != null && i < previous.size()
+                        && previous.get(i).toString().equals(name) ? previous.get(i)
```
manticore-projects: Isn't this a `String` already?

manticore-projects (changes requested):

src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java
minleejae: `getColumns()` returns `List<Index.ColumnParams>`. I've made the element type explicit; `toString()` renders it for the legacy `List<String>` view.

src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java
minleejae: `remove()` returns an `Index.ColumnParams` here. I've made that type explicit and kept the conversion required by the legacy `List<String>` view.

src/main/java/net/sf/jsqlparser/statement/alter/AlterExpression.java
minleejae: The elements are `Index.ColumnParams`. I've made that explicit; comparing their rendered SQL lets unchanged elements retain their structured expressions and options.

fudianchn:
The type clarifications read clearly now. One issue remains in the current approach:

**Reordering columns silently drops structured attributes.** `replaceKeyColumns` decides "unchanged" by position: `previousColumn.toString().equals(name)`. A reorder mismatches every position and rebuilds each element as a plain `ColumnParams`:

```sql
ALTER TABLE t ADD UNIQUE (a DESC, b)
```

```java
action.setUkColumns(List.of("b", "a"));
```

This is inherent to the bridge, not a bug in the check: there is no lossless mapping from `List<String>` to `List<ColumnParams>`, the string `"a"` cannot carry "keep the previous sort order". If that loss is intended, it should be a stated semantic. If not, a smaller alternative covers the same reported breakage without the view class:

```java
public List<String> getPkColumns() {
    return hasKeyIndex(Index.Kind.PRIMARY_KEY) ? index.getColumnsNames() : pkColumns;
}

public void setPkColumns(List<String> pkColumns) {
    if (hasKeyIndex(Index.Kind.PRIMARY_KEY)) {
        index.setColumnsNames(pkColumns);
        this.pkColumns = null;
    } else {
        this.pkColumns = pkColumns;
    }
}
```

Notes:

- `getColumnsNames()` / `setColumnsNames()` already exist on `Index`; the legacy accessors become thin adapters and the diff shrinks a lot.
- The NPE caveat is specific to this alternative: `Index.getColumnsNames()` NPEs when `columns` is null (unguarded `columns.stream()`), and this getter calls it. The view class never does; the NPE is still real for any caller of the existing method today.
- Trade-off, stated plainly: this alternative loses structure even on round-trip edits. `setUkColumns(new ArrayList<>(getUkColumns()))` rebuilds every element as a plain `ColumnParams`, so `a DESC` becomes a column whose rendered name is the string `"a DESC"` (same SQL, degraded AST). The view class keeps the original object there. Predictable loss versus positional-matching loss: different trade-offs, not a strict improvement.
- This alternative does not restore mutation through the returned list for index-backed expressions; that path does not work today anyway, since `toString()` renders from the index.

> AI disclosure: drafted and edited by me, translated by AI.

minleejae:
Switched to the thin `Index` adapters and removed the view and positional matching in [cc6e5285](https://github.com/JSQLParser/JSqlParser/pull/2628/commits/cc6e528501de7d45c3e075bcdf4268a71345bd60). The contract is now explicit in Javadoc and the PR description: index-backed getters return snapshots; string setters and fluent additions rebuild every element, even for unchanged strings. They preserve rendered SQL on an unchanged round trip, but not the structured AST. Reordering with attributes intact should use `Index.getColumns()` / `setColumns()`.

I also made `Index.getColumnsNames()` null-safe and removed the parser's duplicate legacy-column initialization, which would otherwise immediately flatten freshly parsed keys through these setters. Regression tests cover both PK and UNIQUE replacement, structured reordering, parsed sort attributes, snapshots, metadata, and empty indexes. Full `./gradlew check` passed.

manticore-projects:
Thank you!

PR #2629, opened 2026-09-12
Fix benchmark configuration and isolated parser class loading

minleejae:
The historical-version benchmark runner discards its parser configuration and casts AST objects from an isolated class loader to the current version’s `Statements` type. Even loading the same build through an isolated loader can fail with `ClassCastException`, while ignored options make version comparisons unreliable.

Keep AST results behind an `Object` boundary, represent the benchmark’s configurations explicitly, and apply each option through the target loader’s parser methods. Share those configuration choices with the current-version runner and close the isolated loader at trial teardown. Setup now parses the corpus before measurement and rejects both exceptions and null ASTs. Native profiling becomes opt-in via the `jmhProfiler` Gradle property instead of requiring a fixed Linux library path.

Validation:
- `./gradlew check` passed, including JMH compilation and static analysis.
- An isolated-loader regression test verifies distinct AST classes, equivalent SQL output, backslash configuration, invalid SQL, and caller-owned executor lifetime.
- Manual smoke checks with the current build and the actual 5.3/5.1 release JARs apply backslash escaping successfully. The current build accepts the complete performance corpus with SIMPLE configuration; 5.3 and 5.1 return no AST for that combination, and setup now rejects those trials before measurement.

No performance result is claimed. Historical versions need a mutually supported corpus before a valid comparison. Profiling can be enabled with `./gradlew jmh -PjmhProfiler='async:libPath=/path/to/libasyncProfiler.so;output=tree;dir=build/reports/jmh'`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -710,7 +710,8 @@ jmh {
             , "-XX:+DebugNonSafepoints"
     ]
 
-    profilers = ['async:libPath=/opt/async-profiler/lib/libasyncProfiler.so;output=tree;dir=build/reports/jmh']
+    // Profiling is opt-in because the native library path depends on the host.
+    profilers = providers.gradleProperty('jmhProfiler').map { [it] }.getOrElse([])
 
     includes = ['.*JSQLParserBenchmark.*']
     warmupIterations = 2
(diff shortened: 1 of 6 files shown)


manticore-projects:
Thanks!

manticore-projects:
Well, in a way your broke that benchmark now: "java.lang.IllegalStateException: Parser 5.1 returned no statements for the benchmark corpus with SIMPLE configuration".
But it does not really matter: we focus on 5.4 now and everything before that will be history. I also understand, that the numbers were somewhat meaningless before, but the business man in me did not care :-D

PR #2630, opened 2026-09-12
Render DML source clauses through shared visitor paths

minleejae:
UPDATE and DELETE deparsers append CTEs and source clauses as raw AST strings. A custom expression deparser therefore rewrites SET/WHERE values but misses expressions in CTEs, joined subqueries, ON predicates, UPDATE FROM, and DELETE USING. SQL rewriting applications can produce a mixture of transformed and original expressions within one statement.

Route those sources through the configured `SelectDeParser` using common DML rendering support. Reuse the same WITH-item renderer from SELECT, including data-modifying CTEs, without casting a generic expression visitor to `ExpressionDeParser`. Existing UPDATE/DELETE constructors, visitor setters, and protected clause hooks remain available; an overload lets callers supply their select deparser explicitly. UPDATE FROM retains its position before or after SET.

Validation: `./gradlew check` passed. Regression tests cover CTEs containing SELECT/INSERT/UPDATE/DELETE, multiple CTEs, joined and USING subqueries, ON predicates, Teradata FROM placement, direct and statement-level deparsers, generic visitors, configured select visitors, visit counts, SQL reparsing, and unchanged source ASTs.


manticore-projects:
Thank you for cleaning and improving this!

PR #2638, opened 2026-09-14
fix: preserve PostgreSQL quoted names, numeric scales and comment literals

minleejae:
PostgreSQL quoted identifiers such as `"a.b"` are currently split again while constructing table and column AST nodes. For example, `DROP INDEX "a.b"` reports `"b"` as its name and invents schema `"a"`. Numeric types with negative scales fail in type fragments, DDL and casts, and COMMENT rejects escape strings containing an escaped quote and newline-separated ordinary string literals.

This change preserves lexer-separated name components with `Dialect.POSTGRESQL`, including table-name copies. It accepts negative numeric scales through the shared type grammar, preserves precision/scale metadata for qualified types, and distinguishes scale `-1` from an omitted scale. `ColDataType.fromNumericParameters` supports nullable parameters while preserving the existing primitive constructor's sentinel behavior.

Identifier copying is separate from reference cloning: `copyIdentifierPartsTo(Table)` copies parsed name components and separators into an existing table without reparsing quoted names. `clone()` additionally preserves the table-variable flag and resolved identifier, while `setResolvedTable()` stores a detached identifier-only copy. This keeps the shared name-copying logic explicit without introducing a second table-cloning method.

COMMENT and SELECT share character-literal parsing. PostgreSQL `E'...'` enables escapes for that literal, and newline concatenation accepts ordinary quoted continuations while rejecting dollar-quoted continuations. Tagged dollar comments already work in 5.4 with the PostgreSQL preset; the existing default-off behavior is retained and covered by regression tests. Default and BigQuery dotted-name behavior is also retained.

Validation:

- Added 75 parameterized/regression tests covering quoted object and column names, clone/resolved-table names, numeric aliases and boundaries, arrays, DDL/casts, dollar tags and bodies, E strings, literal continuation, malformed input and dialect isolation.
- Gradle `check`: 6,979 tests, no failures/errors, 25 skipped; formatting, static analysis and grammar ambiguity checks passed.
- Maven `clean verify`: 6,961 tests, no failures/errors, 25 skipped.
- Differential audit against PostgreSQL 18.6 and both the 5.4 release and master `6312f9e5`: 253 selected cases, 237 accepted by PostgreSQL. AST/deparse checks improve from 93 to 229 passing cases, with no regressions among previously passing valid cases. All 229 regenerated valid SQL cases execute successfully on PostgreSQL.

The remaining eight valid audit failures are unsupported COMMENT target kinds (`INDEX`, `SCHEMA`, `SEQUENCE`, `DOMAIN`, `TYPE`, `MATERIALIZED VIEW`, `FUNCTION`, `CONSTRAINT`), which require an extension of the Comment target AST. PostgreSQL precision/scale range validation is left to the database.

Syntax references: [numeric types](https://www.postgresql.org/docs/18/datatype-numeric.html), [lexical structure](https://www.postgresql.org/docs/18/sql-syntax-lexical.html), [COMMENT](https://www.postgresql.org/docs/18/sql-comment.html).


src/main/java/net/sf/jsqlparser/schema/Table.java
```diff
@@ -579,9 +579,16 @@ public static Table[] setUnsetCatalogAndSchema(String currentCatalogName,
...
     @Override
     public Table clone() {
-        Table clone = new Table(this.getFullyQualifiedName());
+        Table clone = copyName();
         clone.setTableVariable(tableVariable);
         clone.setResolvedTable(this.resolvedTable != null ? this.resolvedTable.clone() : null);
         return clone;
     }
+
+    private Table copyName() {
```
manticore-projects: This is a very unfortunate name because it hides what it does. Please find a better name which expresses the actual purpose, e. g. `getClonedTable()` (you certainly will find something better).

And why exactly do we have 2 methods `clone` and `copyName` now (when they appear to do the same)?

manticore-projects (changes requested):

src/main/java/net/sf/jsqlparser/schema/Table.java
minleejae: Addressed in 4ad229ee. I replaced `copyName()` with `copyIdentifierPartsTo(Table)`, a `void` helper that copies only the parsed identifier components and separators into an existing table.

manticore-projects:
Thank you!

PR #2639, opened 2026-09-14
feat: support additional PostgreSQL COMMENT targets

minleejae:
PostgreSQL `COMMENT ON INDEX idx IS 'description'` and seven other catalog target kinds currently fail to parse. This adds structured support for `INDEX`, `SCHEMA`, `SEQUENCE`, `DOMAIN`, `TYPE`, `MATERIALIZED VIEW`, `FUNCTION`, and `CONSTRAINT`, including constraints owned by a table or domain.

`CommentTarget` preserves object names and constraint ownership, and reuses `RoutineReference` for function signatures, argument modes, names, and types. The existing table/column/view accessors remain available. Comment rendering is shared with the statement deparser, visitors can traverse explicit relations and literals, and table discovery avoids treating indexes, types, domains, sequences, or functions as tables. A dedicated validator exposes target-specific capabilities.

Regression coverage includes quoted dotted identifiers, qualified and Unicode names, optional function signatures, `ON DOMAIN` versus a table named `domain`, comment removal, tagged dollar strings, AST round trips, and custom visitor/deparser behavior. Unambiguous target syntax works without a dialect preset; tagged dollar strings retain their existing PostgreSQL/explicit-option requirement.

### Validation

- `./gradlew spotlessApply check --max-workers=2 --no-daemon` with the formatting ratchet pinned to base `6312f9e5`: 6,994 tests, zero failures/errors, 25 skipped; formatting, PMD and SpotBugs passed.
- `EXPORT_TEST_TO_FILE=False mvn -B clean verify`: 6,976 tests, zero failures/errors, 25 skipped.
- PostgreSQL 18.6 differential audit: all 228 valid cases passed parsing, AST checks, both rendering paths, and execution with the expected stored comment; 11 malformed cases were rejected by both PostgreSQL and the parser. JSqlParser 5.4 rejected all 228 valid cases.
- Merged locally with #2638 without conflicts: 7,069 Gradle tests, zero failures/errors, 25 skipped. All 237 valid cases from the earlier audit now pass, alongside all 228 valid cases in this audit. Numeric range validation remains outside the parser's scope.

This follows up on the eight unsupported COMMENT targets identified during #2638's audit. It is based directly on master and can merge independently of #2638. Other PostgreSQL COMMENT target kinds are outside this change.

Reference: [PostgreSQL 18 COMMENT syntax](https://www.postgresql.org/docs/18/sql-comment.html).

PR #2640, opened 2026-09-14
feat: preserve DROP INDEX owner tables in the AST

minleejae:
`DROP INDEX ix ON app.t` currently puts the owner in raw parameter tokens, so consumers cannot obtain or rewrite it through a `Table` node. Table discovery also reports the index name as a table. This exposes the explicit owner through `Drop.getTable()` and makes visitors, validation and custom deparsers use it. Index names and DROP names with quoted dots retain their identifier components.

DROP rendering is shared by `Drop` and `DropDeParser`. The legacy `getParameters()` still includes `ON` and the table, but returns a snapshot when an owner is structured. Owner mutations update that snapshot and SQL output; `addParameters()` preserves the owner; `setParameters()` replaces the complete clause with raw tokens and clears the structured owner. MySQL ALGORITHM/LOCK order and optional equals signs are preserved.

This covers the single-owner ON form used by MySQL and SQL Server. PostgreSQL DROP INDEX has no explicit owner, so `getTable()` is null and table discovery does not invent one. SQL Server's multi-owner list and WITH options remain outside this change.

### Validation

- Merged locally with #2638, #2639 and #2641 without conflicts: 7,126 Gradle tests, zero failures/errors, 25 skipped; the grammar ambiguity check passed. All 193 new audit cases produced identical parser/rendering results in the combined build.

- Full Gradle `check`, including the JavaCC ambiguity gate, formatting, Checkstyle, PMD and SpotBugs: 6,919 tests, zero failures/errors, 25 skipped. Spotless ratchet pinned to base `6312f9e5`.
- `EXPORT_TEST_TO_FILE=False mvn -B clean verify`: 6,901 tests, zero failures/errors, 25 skipped.
- MySQL 8.4.11 and PostgreSQL 18.6 differential checks: 34 valid statements retained the same index-deletion result after parsing and both rendering paths; two malformed ON clauses were rejected by the parser and MySQL. Coverage includes quoted/qualified names, the quoted PRIMARY index and ALGORITHM/LOCK variants.
- SQL Server bracketed and database-qualified owners, IF EXISTS, AST mutations, legacy setters, custom deparsing and statement boundaries are covered by regression tests. SQL Server was not executed locally.

This is independent of the general AST querying API discussed in #2620. Related PostgreSQL quoted-name fixes in #2638 remain applicable to other statement contexts.

Syntax references: [MySQL DROP INDEX](https://dev.mysql.com/doc/refman/8.4/en/drop-index.html), [SQL Server DROP INDEX](https://learn.microsoft.com/en-us/sql/t-sql/statements/drop-index-transact-sql?view=sql-server-ver17), [PostgreSQL DROP INDEX](https://www.postgresql.org/docs/18/sql-dropindex.html).


manticore-projects:
Thank you, this is much better now.

PR #2641, opened 2026-09-14
feat: structure column attributes and generated expressions

minleejae:
Column definitions currently retain many attributes as raw strings. For example, `ON UPDATE CURRENT_TIMESTAMP(6)` is split into tokens and a generated column's expression cannot be visited or rewritten. Extend the existing `ColumnOption` model to preserve these attributes in CREATE and ALTER column definitions.

This adds structured nullability, collation, comment literals, ON UPDATE expressions, AUTO_INCREMENT, visibility and inline PRIMARY KEY. `GeneratedColumnDefinition` preserves the expression, optional GENERATED ALWAYS prefix and explicit STORED/VIRTUAL choice; existing identity declarations keep their own model. Shared table-definition traversal and rendering now expose the new expressions to visitors, validators and custom deparsers.

Charset remains on `ColDataType`, while column COLLATE is a `ColumnOption`; consumers can read both directly. Unknown extensions retain their raw fallback and source order. The legacy `getColumnSpecs()` supplies a snapshot for structured options, with canonical keyword capitalization and expression fragments rendered as one token. `setColumnSpecs()` still explicitly replaces structured options; `addColumnSpecs()` preserves them. Existing tests that asserted the old raw-only representation were updated, with raw-setter coverage retained.

### Validation

- Merged locally with #2638, #2639 and #2640 without conflicts: 7,126 Gradle tests, zero failures/errors, 25 skipped; the grammar ambiguity check passed. All 193 new audit cases produced identical parser/rendering results in the combined build.

- Full Gradle `check`, including the JavaCC ambiguity gate, formatting, Checkstyle, PMD and SpotBugs: 6,946 tests, zero failures/errors, 25 skipped. Spotless ratchet pinned to base `6312f9e5`.
- `EXPORT_TEST_TO_FILE=False mvn -B clean verify`: 6,928 tests, zero failures/errors, 25 skipped.
- Database differential audit: 132 valid MySQL 8.4.11 cases and 20 valid PostgreSQL 18.6 cases produced identical column metadata from original and regenerated SQL; five malformed definitions were rejected. Checks cover CREATE/ADD/MODIFY/CHANGE, charset/collation, stored comments, ON UPDATE and generated-column expressions/storage.
- Regression tests check AST payloads after both rendering paths, expression visitors, validation, mutations, identity/reference/default boundaries and legacy token behavior.

No server defaults or expression-validity rules are inferred. SQL Server's unparenthesized computed-column forms, complete column-constraint modeling and semantic literal decoding remain outside this change. This enriches the existing AST independently of the querying API discussed in #2620.

Syntax references: [MySQL column definitions](https://dev.mysql.com/doc/refman/8.4/en/create-table.html), [MySQL generated columns](https://dev.mysql.com/doc/refman/8.4/en/create-table-generated-columns.html), [MySQL automatic timestamp updating](https://dev.mysql.com/doc/refman/8.4/en/timestamp-initialization.html), [PostgreSQL column definitions](https://www.postgresql.org/docs/18/sql-createtable.html).


manticore-projects:
This is also much better now, thank you much!

PR #2645, opened 2026-09-18
fix: refresh index classification when replacing its type

minleejae:
Changing an existing index from `UNIQUE` to `INDEX` updates its rendered type but leaves `getKind()` at `UNIQUE`. Consumers rewriting CREATE/ALTER TABLE statements can therefore observe a classification that no longer matches the type they assigned.

Derive the classification on every `setType`/`withType` call through one helper. Null and custom types clear stale classifications, keyword matching respects word boundaries, and original spelling is preserved. `setKind` remains an explicit metadata override without rewriting SQL, supporting declarations that store the index keyword separately and specialized constraint nodes.

Validation: Gradle `check` passed (7502 tests, 0 failures, 0 errors, 25 skipped), including formatting, static analysis and the grammar check. Regression tests cover repeated edits, all prior classifications, custom/null types, subclass defaults and CREATE TABLE deparse/reparse. Eight original/rewritten DDL cases executed successfully on MySQL 8.4.11 and PostgreSQL 18.6.

Syntax references: [MySQL CREATE TABLE](https://dev.mysql.com/doc/refman/8.4/en/create-table.html), [PostgreSQL CREATE TABLE](https://www.postgresql.org/docs/18/sql-createtable.html).


manticore-projects:
Thank you!

PR #2646, opened 2026-09-18
fix: preserve lambda grouping and dialect JSON arrows

minleejae:
A parenthesized single-parameter lambda such as `arrayMap((x) -> x * 2, [1, 2])` loses its parameter parentheses when rendered, so reparsing changes its AST from `LambdaExpression` to `JsonExpression`. PostgreSQL and MySQL JSON access in later function arguments can also be incorrectly classified as a lambda.

Preserve explicit parameter grouping in `LambdaExpression` and share its rendering between `toString()` and the expression deparser, including the caller's body visitor. In PostgreSQL and MySQL/MariaDB presets, expression-list and parenthesized-operand arrow handling uses JSON access. Existing unparenthesized lambda rendering in the default parser remains available.

Validation: Gradle `check` passed (7498 tests, 0 failures, 0 errors, 25 skipped), including grammar ambiguity and static checks. New tests cover grouped single/multiple parameters, nesting, argument positions, custom deparsers, fast/complex parsing and missing JSON operands. Thirteen PostgreSQL/MySQL JSON fixtures and both rendered forms executed successfully on PostgreSQL 18.6 and MySQL 8.4.11 (39 executions).

Syntax references: [PostgreSQL JSON operators](https://www.postgresql.org/docs/18/functions-json.html), [MySQL JSON extraction](https://dev.mysql.com/doc/refman/8.4/en/json-search-functions.html), [ClickHouse lambda syntax](https://clickhouse.com/docs/reference/functions/regular-functions/overview).


manticore-projects:
Thank you for fixing this, silent errors like those are the worst.

PR #2647, opened 2026-09-18
feat: expose base type names and retain large numeric lengths

minleejae:
MySQL accepts `BLOB(4294967295)`, but parsing it overflows an Integer. The valid numeric length `2147483647` is also incorrectly rendered as `MAX`. Separately, consumers reading a type name must strip parameters from the legacy `getDataType()` result, such as `decimal (10, 2)`.

Use exact numeric precision metadata and a shared parameter factory that distinguishes numeric lengths from the MAX keyword. Add `getNumericPrecision()` and `isMaxPrecision()` while retaining the existing Integer accessor and primitive-constructor sentinel behavior. The Integer accessor returns null for numeric lengths outside its range. `getBaseTypeName()` provides a shared, quote-aware view of the unparameterized name while preserving legacy spelling and rendering.

Validation: Gradle `check` passed (7512 tests, 0 failures, 0 errors, 25 skipped), including grammar and static checks. Tests cover MySQL BLOB/TEXT family boundaries, the Integer boundary, MAX, metadata edits, PostgreSQL numeric/temporal/array types, quoted names, fragments, casts and DDL round trips. The reflection-based test printer compares numeric values without reflecting into encapsulated JDK fields.

All 44 valid SQL fixtures and both rendered forms execute on MySQL 8.4.11 or PostgreSQL 18.6 (132 executions). An additional 30-case MySQL BLOB/TEXT audit distinguishes accepted lengths from negative syntax and server range errors. Database-specific semantic limits remain the database's responsibility; syntactically numeric lengths are retained exactly.

Syntax references: [MySQL string types](https://dev.mysql.com/doc/refman/8.4/en/string-type-syntax.html), [PostgreSQL numeric types](https://www.postgresql.org/docs/18/datatype-numeric.html).


manticore-projects:
Please resolve conflicts.

minleejae:
@manticore-projects Thanks for flagging this!

> Please resolve conflicts.

Merged the latest `master` and resolved the conflicts. The type metadata changes now preserve negative scales and the existing `fromNumericParameters()` API, including its legacy `MAX` sentinel. I also retained master's signed generic type-argument handling and added regression coverage for the combined behavior.

The full Gradle `check` passes on Java 17: 7,772 tests, 0 failures/errors, 25 skipped, including grammar ambiguity and static checks. GitHub reports no merge conflicts; CI is running on the updated head.

Commit: [a65e84cd](https://github.com/JSQLParser/JSqlParser/commit/a65e84cd915837c34948b1849c263c00bb87807f)

manticore-projects:
Thank you very much and sorry for the extra round. I wished GitHub would support semantic, AST based merges.

PR #2648, opened 2026-09-18
feat: support typed result columns in PostgreSQL ROWS FROM

minleejae:
PostgreSQL allows each record-returning function in `ROWS FROM` to declare its own result columns, for example `ROWS FROM (json_to_recordset('[{"a":1}]') AS (a integer)) WITH ORDINALITY`. This currently fails to parse.

Attach typed result-column definitions to their Function nodes through `getResultColumnDefinitions()`, reusing ColumnDefinition, ColDataType and structured COLLATE options. The existing mutable ROWS FROM function list and outer alias/ordinality remain intact, so moving functions also moves their definitions. A dedicated production accepts column names, types and optional collations without accepting table constraints.

Share result-column rendering between Function and its deparser, and route TableFunction arguments through the supplied expression visitor. This supports AST edits and custom expression rewriting inside ordinary and ROWS FROM table functions.

Validation: Gradle `check` passed (7500 tests, 0 failures, 0 errors, 25 skipped), including grammar and static checks. Maven `clean verify` also passed (7482 tests, 0 failures, 0 errors, 25 skipped). PostgreSQL 18.6 accepted all 12 positive fixtures and both rendered forms (36 executions), and rejected all eight negative fixtures. Positive and negative inputs are stored as TSV test data. Tests cover multiple functions, typed/untyped mixtures, lateral references, aliases, ordinality, numeric/array/time-zone types, collations, quoted names, mutable-list edits, custom deparsers and table discovery inside function arguments.

Syntax reference: [PostgreSQL table functions and ROWS FROM](https://www.postgresql.org/docs/18/queries-table-expressions.html#QUERIES-TABLEFUNCTIONS).


manticore-projects:
Nice, thank you much!

PR #2649, opened 2026-09-18
feat: expose dialect-aware current-time expression metadata

minleejae:
Consumers inspecting current-time expressions must currently handle TimeKeyExpression, Function and Column separately. For example, PostgreSQL `CURRENT_TIMESTAMP`, `CURRENT_TIMESTAMP(6)` and bare `LOCALTIME` use three different node classes, and MySQL's LOCALTIME alias has different semantics from PostgreSQL's.

Add `TemporalExpressionInfo.from(expression, dialect)` as a common read-only metadata API for MySQL and PostgreSQL. It reports the normalized kind, original name, explicit precision and parentheses while retaining existing AST classes and rendering. It handles MySQL NOW/CURDATE/CURTIME aliases, PostgreSQL now(), grouping, and the dialect-specific LOCALTIME/LOCALTIMESTAMP meanings. Quoted/qualified identifiers, unrelated expressions and unsupported call shapes return an empty result.

Results are snapshots and precision is the requested value. This API does not apply database range validation or clamping: PostgreSQL accepts precision 7 with a warning and clamps it to 6, while MySQL rejects it. Usage and these boundaries are documented.

Validation: Gradle `check` passed (7549 tests, 0 failures, 0 errors, 25 skipped), including formatting, grammar and static checks. A 160-case execution matrix on PostgreSQL 18.6 and MySQL 8.4.11 supplied 43 accepted fixtures. All accepted originals and both rendered forms executed successfully (129 executions). Tests also cover omitted/zero precision, aliases, quoted/qualified names, invalid argument shapes, AST-class preservation and edits after a metadata snapshot.

Syntax references: [PostgreSQL current date/time functions](https://www.postgresql.org/docs/18/functions-datetime.html#FUNCTIONS-DATETIME-CURRENT), [MySQL date/time functions](https://dev.mysql.com/doc/refman/8.4/en/date-and-time-functions.html).


manticore-projects:
Thank you!

PR #2651, opened 2026-09-18
fix: avoid recursion when deparsing statement lists

minleejae:
Calling `Statements.accept(new StatementDeParser(builder), context)` currently throws `StackOverflowError`, even for an empty list: the deparser visits the same container again instead of its children. This prevents callers from rendering or rewriting a parsed SQL script through one visitor.

This change visits each child statement and shares list separators between `Statements.toString()` and `StatementDeParser`. `Block` and `IfElseStatement` gain callback-based rendering overloads, so nested lists use the same separator policy and IF conditions and branches reach custom visitors. The supplied context is forwarded at those dispatch points. Existing `toString()` output, branch/block semicolon flags, and error-recovery null placeholders are preserved.

Regression coverage includes empty and repeated calls, adjacent DDL/DML statements, nested blocks, IF/ELSE delimiter flags, custom condition and statement visitors, AST immutability, unsupported statements and error recovery. Dialect round trips cover MySQL, PostgreSQL quoted routine/DO bodies, Oracle blocks and SQL Server IF statements. The usage guide includes a complete-list deparser example.

Validation:

- `./gradlew spotlessApply check --no-daemon`: 7,505 tests, 0 failures/errors, 25 skipped; includes 30 new regression cases.
- `mvn --batch-mode --no-transfer-progress clean verify`: 7,487 tests, 0 failures/errors, 25 skipped.

The change is confined to rendering and visitor dispatch; it does not change the grammar or interpret procedural bodies.


manticore-projects:
Thank you!

PR #2652, opened 2026-09-19
fix: default wildcard exclusions to EXCEPT

jakezwang:
Adding excluded columns to a new or parsed wildcard currently produces SQL such as `SELECT * null( hidden ) FROM t`. Use `EXCEPT` when no keyword was specified, preserving explicit `EXCLUDE` and existing keyword casing.

The regressions cover programmatic `*`, parsed `*` and `t.*`, including SQL rendering, deparsing and roundtrips. All three fail before the fix; empty/null exclusion lists and explicit keywords remain covered.

Validation on JDK17: `mvn spotless:apply verify` and `./gradlew check` passed, including the full test suites and static checks. `./gradlew jmh` completed the standard parser benchmarks.


manticore-projects:
Please resolve conflicts.

manticore-projects:
Thank you very much!

PR #2653, opened 2026-09-19
Fix parsing of ClickHouse transformers (APPLY/EXCEPT/EXCLUDE/REPLACE) after '*'

### LINKED ISSUE #2636: [BUG] 5.4-SNAPSHOT : ClickHouse : asterisk column modifiers fail or mis-parse (APPLY as alias, docs chain, STRICT, EXCEPT regex)
*AI disclosure: this issue was prepared with AI coding agents, revised and verified by me.*

### Failing SQL Feature:
ClickHouse column modifiers on `*` / `t.*`: the documented `APPLY` modifier is silently mis-parsed as a select-item alias, the documented modifier chain from the [SELECT docs](https://clickhouse.com/docs/en/sql-reference/statements/select) fails with a ParseException, and the `STRICT` variants plus the regex-string form of `EXCEPT` (implemented in the ClickHouse grammar, covered by its stateless tests) are unsupported.

### SQL Example:
```sql
-- 1) documented in the ClickHouse SELECT docs, mis-parsed on master:
--    parses as AllColumns with alias "APPLY(sum)"
SELECT * APPLY(sum) FROM t;

-- 2) the modifier chain exactly as shown in the ClickHouse docs, fails:
SELECT * REPLACE(i + 1 AS i) EXCEPT (j) APPLY(sum) FROM columns_transformers;

-- 3) STRICT variants (undocumented, but in the ClickHouse grammar and
--    its stateless tests) and the regex form of EXCEPT, all fail:
SELECT * EXCEPT STRICT (a, b) FROM t;
SELECT * REPLACE STRICT (1 AS a, 2 AS b) FROM t;
SELECT * EXCEPT '^tmp_' FROM t;
```

Baseline that already parses: `SELECT * EXCEPT (a, b) FROM t`, `SELECT * REPLACE (a AS b) FROM t`, `SELECT * EXCLUDE (a) FROM t`.

### Software Information:
- JSqlParser version: current master 5.4-SNAPSHOT, commit 02b9d940 (tested 2026-09-14)
- Database: ClickHouse

### Additional context
- ClickHouse grammar accepts `STRICT` after both EXCEPT and REPLACE, and a string-literal regex pattern for EXCEPT (`ExpressionElementParsers.cpp`: `is_strict`, `ParserStringLiteral` -> `ASTColumnsExceptTransformer::setPattern`, formatted as `EXCEPT STRICT` in `ASTColumnsTransformers.cpp`); covered by the stateless tests `01470_columns_transformers.sql` and `02343_analyzer_column_transformers_strict.sql`.
- The [SELECT modifiers documentation](https://clickhouse.com/docs/en/sql-reference/statements/select) lists only APPLY, EXCEPT and REPLACE; the STRICT and regex forms are undocumented.
- References: [ExpressionElementParsers.cpp](https://github.com/ClickHouse/ClickHouse/blob/master/src/Parsers/ExpressionElementParsers.cpp), [ASTColumnsTransformers.cpp](https://github.com/ClickHouse/ClickHouse/blob/master/src/Parsers/ASTColumnsTransformers.cpp)
- Related: PR #2635 adds the `COLUMNS('regexp')` context for the documented modifiers; the asterisk position is untouched there, so this issue is independent of it.

kyy-logs:
Fixes #2636.

## What

`*` is parsed by `AllColumns`, which modelled the trailing modifiers itself (`exceptColumns` / `replaceExpressions` / `exceptKeyword`) and accepted only a single `EXCEPT`/`EXCLUDE` followed by a single `REPLACE`. `APPLY` was not accepted at all, so

```sql
SELECT * APPLY(sum) FROM t
```

parsed as `AllColumns` with the alias `"APPLY(sum)"` — syntactically fine, round-trips fine, wrong AST. Anything building on the AST sees a wildcard with an alias instead of a transformer.

## How

`AllColumns` now holds a `List<ColumnsTransformer>`, mirroring `ColumnsExpression` from #2635, and the production loops over `ColumnsTransformer()`. Chains and any keyword order come for free.

`ColumnsTransformerType` gains `EXCLUDE` (DuckDB), previously carried as a plain string in `AllColumns.exceptKeyword`.

## Regression risk I checked

The transformer lookahead could not be reused verbatim. `ColumnsTransformerAhead()` returns false for `(<K_APPLY>|<K_REPLACE>) AliasColumns()`, which is right after `COLUMNS(...)` — `COLUMNS('m') APPLY(a, b)` is an alias — but wrong after `*`, where it would turn `* REPLACE (x)` into an alias. `AllColumnsTransformerAhead()` keeps the alias reading for `APPLY` only. `ClickHouseTest` does not cover `* REPLACE (col)`, so this was found by reading the `AliasColumns()` production rather than by a failing test.

Deliberately unchanged: `* APPLY` and `* APPLY(a, b)` stay aliases.

This also widens the shared `ColumnsTransformer()` production: `COLUMNS(...) EXCLUDE (...)` now parses. That is beyond the issue, tell me if you'd rather gate it.

## Relation to #2652

#2652 patches `AllColumns.appendTo()` for a null `exceptKeyword` and adds `addExceptColumn` / `setExceptColumns` tests. Those accessors are the ones this change removes, so the two cannot both land as-is. Under the transformers model the null-keyword state cannot arise — a transformer always knows its own keyword — so #2652's fix is subsumed. Happy to rebase on it if it lands first.

## Verification

```
mvn clean test
```

7457 tests, 0 failures, 0 errors, 25 skipped. `spotless:check`, checkstyle and PMD pass.

New tests in `ClickHouseTest`: `testAllColumnsApplyIssue2636` (fails before the change with `expected: <null> but was: < APPLY(sum)>`) and `testAllColumnsChainedTransformersIssue2636` (fails before with a parse error at `EXCEPT (a)`).

AI disclosure: I use an AI coding assistant; I reviewed and tested everything here and take responsibility for it.


manticore-projects:
Thank you very much and welcome to this project!

PR #2654, opened 2026-09-20
fix: reuse table constraints for MySQL ALTER TABLE

minleejae:
MySQL ALTER TABLE ADD rejects constraint definitions that already work in CREATE TABLE, including `CONSTRAINT uq UNIQUE KEY idx (id)` and unnamed CHECK constraints. With `Dialect.MYSQL`, both paths now use the existing table-constraint parser and preserve the constraint symbol and index name separately.

Reuse NamedConstraint/CheckConstraint instead of extending the legacy ALTER branches. Keep named-index options on their AST, preserve legacy key accessors, and require one CHECK expression. Tests cover quoted and omitted names, KEY/INDEX variants, enforcement, multiple ALTER actions, AST edits, and custom expression deparsing; the existing PostgreSQL, Informix, and SQL Server tests also pass.

Syntax reference: [MySQL 8.4 ALTER TABLE](https://dev.mysql.com/doc/refman/8.4/en/alter-table.html).

Validation:

- Java 17 Gradle `check`: 7,826 XML-reported tests, 0 failures/errors, 25 skipped; formatting, grammar ambiguity, Checkstyle, PMD, and JMH compilation passed. Formatting was checked against upstream/master using a local Spotless ratchet override.
- Actual MySQL 8.4.11 server: 12 accepted and 4 rejected SQL cases matched the expected behavior.
- Executed both AST and StatementDeParser output for the accepted cases on the same server: 24 successful executions.


manticore-projects:
I am really grateful for this consolidation and cleanup, thank you!

PR #2655, opened 2026-09-20
fix: support FILTER after ordered-set aggregates

minleejae:
Valid PostgreSQL expressions such as `percentile_cont(0.5) WITHIN GROUP (ORDER BY score) FILTER (WHERE active)` currently fail to parse. Accept FILTER after WITHIN GROUP and render it in that position in both AST and visitor output, while preserving ordinary aggregate/window FILTER behavior.

Reuse AnalyticExpression and share filter parsing/rendering helpers. In the PostgreSQL dialect, reject the reversed FILTER-before-WITHIN-GROUP form. Tests cover ordered-set and hypothetical-set aggregates, filter/order expression visitors, malformed predicates, and window regressions.

Syntax reference: [PostgreSQL aggregate expressions](https://www.postgresql.org/docs/17/sql-expressions.html#SYNTAX-AGGREGATES).

Validation:

- Java 17 Gradle `check`: 7,820 XML-reported tests, 0 failures/errors, 25 skipped; formatting, grammar ambiguity, Checkstyle, PMD, and JMH compilation passed. Formatting was checked against upstream/master using a local Spotless ratchet override.
- Actual PostgreSQL 17.11 server: 8 accepted and 5 rejected SQL cases matched the expected behavior.
- Executed both AST and StatementDeParser output for the accepted cases on the same server: 16 successful executions.


manticore-projects:
Thank you!

PR #2656, opened 2026-09-20
feat: parse MySQL table option lists and UNION sources

minleejae:
CREATE TABLE currently rejects comma-separated options such as `STATS_PERSISTENT=1, STATS_AUTO_RECALC=1` and MERGE-table `UNION=(t1,t2)` options. Parse these forms, including mixed separators and an empty UNION list, using a shared option-list production.

Expose statistics options through TableOption kinds and UNION sources as mutable Table references. Legacy token/value views and both renderers reflect AST edits, and TableDefinitionTraversal visits UNION sources. Preserve PARTITION BY, AS SELECT, and Spanner interleave boundaries; reject incomplete separators and invalid boolean statistics values.

Syntax reference: [MySQL 8.4 CREATE TABLE](https://dev.mysql.com/doc/refman/8.4/en/create-table.html).

Validation:

- Java 17 Gradle `check`: 7,826 XML-reported tests, 0 failures/errors, 25 skipped; formatting, grammar ambiguity, Checkstyle, PMD, and JMH compilation passed. Formatting was checked against upstream/master using a local Spotless ratchet override.
- Actual MySQL 8.4.11 server: 9 accepted and 7 rejected SQL cases matched the expected behavior.
- Executed both AST and StatementDeParser output for the accepted cases on the same server: 18 successful executions.


manticore-projects:
Thank you!

PR #2657, opened 2026-09-20
feat: support PostgreSQL PREPARE parameter types

minleejae:
`PREPARE p(bigint) AS SELECT ...` currently fails because PrepareStatement only supports inferred parameter types. Add an optional structured parameter-type list, reusing ColDataType for arrays, qualified names, and type modifiers, with PostgreSQL UNKNOWN retained for inference.

Add a statement-writer overload to the existing renderer so StatementDeParser delegates the prepared query/DML to the configured visitors. This fixes nested expressions bypassing a custom ExpressionDeParser. Existing constructors and the untyped form remain available. Tests cover type edits, quoted names, SELECT/INSERT/UPDATE/DELETE, JDBC parameter preservation, and malformed lists.

Syntax reference: [PostgreSQL PREPARE](https://www.postgresql.org/docs/17/sql-prepare.html).

Validation:

- Java 17 Gradle `check`: 7,823 XML-reported tests, 0 failures/errors, 25 skipped; formatting, grammar ambiguity, Checkstyle, PMD, and JMH compilation passed. Formatting was checked against upstream/master using a local Spotless ratchet override.
- Actual PostgreSQL 17.11 server: 10 accepted and 4 rejected SQL cases matched the expected behavior.
- Executed both AST and StatementDeParser output for the accepted cases on the same server: 20 successful executions.


manticore-projects:
Thank you!

PR #2658, opened 2026-09-20
feat: support PostgreSQL COPY WITH options

minleejae:
COPY already models table/query sources and STDIN/STDOUT, but rejects PostgreSQL's `WITH (FORMAT csv, HEADER true)` option syntax. Preserve the optional WITH keyword, including PostgreSQL's accepted WITH-only legacy form, and reject empty option entries in the new WITH form and PostgreSQL dialect.

Reuse CopyStatement's option representation and add a select-writer overload so query expressions reach the configured deparser. Tests cover CSV delimiters/nulls, column-list and wildcard FORCE_QUOTE, HEADER MATCH, ON_ERROR, CTE/UNION queries, AST edits, and malformed lists. Existing DuckDB tests pass. PROGRAM, WHERE, COPY data-body parsing, and DML RETURNING query sources are outside this change.

Syntax reference: [PostgreSQL COPY](https://www.postgresql.org/docs/17/sql-copy.html).

Validation:

- Java 17 Gradle `check`: 7,829 XML-reported tests, 0 failures/errors, 25 skipped; formatting, grammar ambiguity, Checkstyle, PMD, and JMH compilation passed. Formatting was checked against upstream/master using a local Spotless ratchet override.
- Actual PostgreSQL 17.11 server: 13 accepted and 6 rejected SQL cases matched the expected behavior.
- Executed both AST and StatementDeParser output for the accepted cases on the same server: 26 successful executions. COPY FROM STDIN cases also loaded CSV/text rows and verified row counts.


manticore-projects:
Thank you much!

PR #2661, opened 2026-09-20
fix: share SET assignment parsing and support MySQL character set selection

minleejae:
`SET @a := 1` fails, while subsequent user-variable assignments are stored inside the first assignment's value list: `SET @a=1, @b=2` reports only one entry. `SET CHARACTER SET utf8mb4` is also unsupported.

Share one assignment production between the first and subsequent entries, and represent punctuation with AssignmentOperator (NONE, EQUALS, COLON_EQUALS). Each top-level assignment can now be inspected and modified independently. Existing boolean useEqual methods retain their equals/no-equals behavior; the new accessors preserve `:=`. Nested assignments remain inside their value expression, and PostgreSQL comma-separated setting values remain a single entry.

Add CHARACTER SET/CHARSET selection with quoted names and DEFAULT, including valid combinations such as `SET CHARSET DEFAULT, @a=1`. Statement punctuation continues to be shared with expression deparsers. Existing SQL Server option groups and PostgreSQL settings have regression coverage.

References: [MySQL SET assignments](https://dev.mysql.com/doc/refman/8.4/en/set-variable.html), [SET CHARACTER SET](https://dev.mysql.com/doc/refman/8.4/en/set-character-set.html).

Validation:

- Full Gradle `check` passed on Java 17, including grammar ambiguity, formatting, static analysis, coverage and the full test suite. JUnit XML totals: 7831 tests, 0 failures/errors, 25 skipped.
- MySQL 8.4.11: 19 cases (14 valid, 5 invalid). For all 14 valid cases, the original SQL, toString output and StatementDeParser output executed with identical results.


manticore-projects:
Nice refactoring, thank you!

PR #2664, opened 2026-09-20
fix: parse PostgreSQL JSON path deletion and existence operators

minleejae:
PostgreSQL `jsonb #- text[]` currently becomes an Intersects expression followed by unary minus, so both output paths emit `# -` and the SQL fails on PostgreSQL. `jsonb @? jsonpath` is also missing; without whitespace, the lexer can absorb `@` into the preceding identifier.

Recognize `#-` and `@?` as complete tokens and reuse JsonOperator plus the existing expression precedence loop. Preserve left-associative chains, comparisons and JDBC parameters. In the PostgreSQL dialect, handle existing JSON access operators (`->`, `->>`, `#>`, `#>>`) and concatenation at the same precedence, so mixed chains preserve their left-to-right AST structure in both parser modes, and split `@` out of unquoted identifiers as already done for `#`. Existing hash operators and identifier behavior in other dialects retain regression coverage. The legacy `-#` branch remains accepted for compatibility.

Tests check the actual operator and child AST types, visitors, mutations, missing operands, no-whitespace forms and token locations. JSON path contents remain string expressions; this does not add a JSON path parser.

References: [PostgreSQL JSON operators](https://www.postgresql.org/docs/17/functions-json.html), [operator precedence](https://www.postgresql.org/docs/17/sql-syntax-lexical.html#SQL-PRECEDENCE).

Validation:

- Full Gradle `check` passed on Java 17, including grammar ambiguity, formatting, static analysis, coverage and the full test suite. JUnit XML totals: 7822 tests, 0 failures/errors, 25 skipped.
- PostgreSQL 17.11: 24 cases (21 valid, 3 invalid). For all 21 valid cases, the original SQL, toString output and StatementDeParser output executed with identical results.


manticore-projects:
Thank you!

PR #2666, opened 2026-09-21
feat: support MySQL TABLE query sources in CREATE TABLE

minleejae:
MySQL accepts `CREATE TABLE target [AS] TABLE source`, but the CREATE TABLE grammar currently requires an AS SELECT source. This prevents tools from inspecting or rewriting valid table-copy statements.

Reuse TableStatement for the source query, including qualified and quoted names, ORDER BY and LIMIT/OFFSET. Preserve whether AS was present for both TABLE and SELECT sources. Share query-suffix rendering between CreateTable and CreateTableDeParser so AST mutations use the same punctuation in both output paths; programmatically constructed CreateTable objects still use AS by default.

Regression tests cover source AST types, table discovery, mutation, table options, temporary tables, SELECT controls and malformed source boundaries.

Syntax reference: [MySQL CREATE TABLE ... SELECT/TABLE](https://dev.mysql.com/doc/refman/8.4/en/create-table-select.html).

Validation:

- Full Gradle `check` passed on Java 17, including grammar ambiguity, formatting, static analysis, coverage and the full test suite. JUnit XML totals: 7994 tests, 0 failures/errors, 25 skipped.
- MySQL 8.4.11: 13 cases (9 valid, 4 invalid). For all 9 valid cases, the original SQL, `toString()` output and `StatementDeParser` output executed successfully with matching query results or resulting schema state. Invalid inputs were rejected by the server.


manticore-projects:
Thank you!

PR #2667, opened 2026-09-21
feat: model MySQL expression-first CONVERT syntax

minleejae:
MySQL `CONVERT(expression, type)` is not represented by the existing CONVERT grammar, which primarily handles the SQL Server type-first form and character-set conversion with USING. Valid casts such as `CONVERT(amount, DECIMAL(12,4))` therefore fail or are interpreted in the wrong argument order.

Replace the internal boolean syntax choice in TranscodingFunction with USING, TYPE_FIRST and TYPE_LAST variants, retaining the existing constructors and boolean accessors. Share its rendering with ExpressionDeParser so operand visitors and type mutations work consistently. Reuse ColDataType for lengths, precision and character-set clauses.

The MYSQL dialect selects expression-first parsing for ambiguous inputs such as `CONVERT(value, CHAR)`. Default and SQLSERVER parsing retain the existing type-first precedence; callers processing ambiguous MySQL input should explicitly select `Dialect.MYSQL`. Regression coverage includes USING, SQL Server styles and TRY_CONVERT, nested conversions, NULL, parameters, type mutation and incomplete arguments.

Syntax reference: [MySQL conversion functions](https://dev.mysql.com/doc/refman/8.4/en/cast-functions.html).

Validation:

- Full Gradle `check` passed on Java 17, including grammar ambiguity, formatting, static analysis, coverage and the full test suite. JUnit XML totals: 7997 tests, 0 failures/errors, 25 skipped.
- MySQL 8.4.11: 20 cases (16 valid, 4 invalid). For all 16 valid cases, the original SQL, `toString()` output and `StatementDeParser` output executed successfully with matching query results or resulting schema state. Invalid inputs were rejected by the server.


manticore-projects:
Thank you!

PR #2668, opened 2026-09-21
feat: model PostgreSQL LOCK targets and lock modes

minleejae:
PostgreSQL LOCK supports multiple targets, inheritance scopes, omitted TABLE/IN MODE clauses and three modes missing from the existing single-table model. This prevents parsing common statements such as `LOCK ONLY a, b * NOWAIT`.

Introduce structured lock targets with DEFAULT, ONLY and INCLUDING_DESCENDANTS scopes, and support all eight PostgreSQL lock modes. Factor target and mode parsing into separate productions and share statement rendering with StatementDeParser. TablesNamesFinder now visits every target. The existing table getters/setters remain a view of the first target, and existing Oracle SHARE UPDATE and WAIT forms retain regression coverage.

Tests cover all modes, multiple targets, optional clauses, quoted and qualified names, AST mutation, legacy constructors/accessors and conflicting or incomplete syntax. Oracle compatibility is covered by parser tests; Oracle server execution was not performed.

Syntax reference: [PostgreSQL LOCK](https://www.postgresql.org/docs/17/sql-lock.html).

Validation:

- Full Gradle `check` passed on Java 17, including grammar ambiguity, formatting, static analysis, coverage and the full test suite. JUnit XML totals: 7999 tests, 0 failures/errors, 25 skipped.
- PostgreSQL 17.11: 19 cases (13 valid, 6 invalid). For all 13 valid cases, the original SQL, `toString()` output and `StatementDeParser` output executed successfully with matching query results or resulting schema state. Invalid inputs were rejected by the server.


manticore-projects:
Thank you!

PR #2669, opened 2026-09-21
feat: add a structured SQL JSON predicate

minleejae:
PostgreSQL accepts `expression IS [NOT] JSON` with optional VALUE/SCALAR/ARRAY/OBJECT and key-uniqueness clauses. These predicates currently fail to parse, including when used in WHERE and CHECK expressions.

Add IsJsonExpression with a structured operand, JSON type and WITH/WITHOUT UNIQUE options. Preserve optional KEYS spelling and share rendering with ExpressionDeParser. The visitor default traverses the operand, preserving compatibility for existing visitor implementations, and TablesNamesFinder explicitly handles the new node.

Tests cover both expression-parser modes, negation, option combinations, boolean precedence, NULL, subqueries, visitor context, AST mutation, operand validation and malformed options. This models the predicate syntax; JSON contents are evaluated by the database, not by JSqlParser.

Syntax reference: [PostgreSQL JSON predicates](https://www.postgresql.org/docs/17/functions-json.html).

Validation:

- Full Gradle `check` passed on Java 17, including grammar ambiguity, formatting, static analysis, coverage and the full test suite. JUnit XML totals: 8000 tests, 0 failures/errors, 25 skipped.
- PostgreSQL 17.11: 21 cases (17 valid, 4 invalid). For all 17 valid cases, the original SQL, `toString()` output and `StatementDeParser` output executed successfully with matching query results or resulting schema state. Invalid inputs were rejected by the server.


manticore-projects:
Thank you!

PR #2671, opened 2026-09-21
fix: preserve visitor context in expression deparsing

minleejae:
ExpressionDeParser accepts a caller context but drops it at binary operators and several helper boundaries. A custom visitor processing `COALESCE(7, ABS(8 + 9))`, for example, receives the context for 7 and null for 8 and 9. This makes context-dependent rewriting inconsistent within a single expression.

Forward the supplied context through binary operators, expression lists/row constructors, function and analytic ORDER BY/LIMIT helpers, and parenthesized SELECT/CTE visitor handoffs. Add context-aware helper overloads while retaining the existing null-context entry points and legacy ORDER BY extension hooks.

Regression tests assert context identity through nested arithmetic, boolean, bitwise, CASE, IN, aggregate and window expressions, with unchanged default SQL output and legacy helper behavior. This is an expression traversal fix and does not change database syntax.

Validation:

- Full Gradle `check` passed on Java 17, including grammar ambiguity, formatting, static analysis, coverage and the full test suite. JUnit XML totals: 7993 tests, 0 failures/errors, 25 skipped.


manticore-projects:
Thank you for fixing this!

PR #2672, opened 2026-09-21
fix: parse compound expressions in array constructors

minleejae:
Valid PostgreSQL input such as `ARRAY['a', (1 + 0)::text]` fails because a short syntactic lookahead cannot see through the parenthesized arithmetic expression and chooses the nested-array fallback. A bare nested array beginning with a parenthesized element can also be mistaken for a function call.

Parse every array element through a shared ArrayElement production, reusing the existing expression and range paths for the first and subsequent elements. Dispatch bare array constructors by their opening bracket and exclude that token from both the function-name probe and its cache eligibility guard. The existing ArrayConstructor AST and visitor APIs are unchanged.

Regression tests cover both expression-parser modes, first/middle/last positions, arithmetic and casts, CASE, scalar subqueries, NULL, row constructors, multidimensional arrays, AST mutation and visitor/table traversal. Existing bare/typed array, range and subscript forms retain coverage. Type compatibility and rectangular dimensions remain database checks.

Syntax reference: [PostgreSQL array constructors](https://www.postgresql.org/docs/17/sql-expressions.html#SQL-SYNTAX-ARRAY-CONSTRUCTORS).

Validation:

- PostgreSQL 17.11: 33 cases, 89 executions, all verified. The 25 valid cases produced matching results for the original SQL, `toString()` output and `StatementDeParser` output. Five malformed inputs were rejected by the server; three additional type/dimension errors were rejected for all three output forms. Regressions include SELECT, INSERT, UPDATE and a JSON path-deletion expression using an array operand.
- Full Java 17 Gradle `check` passed, including grammar ambiguity, formatting, Checkstyle, PMD, SpotBugs, coverage and 8006 JUnit tests (0 failures/errors, 25 skipped).


manticore-projects:
Thank you!

PR #2673, opened 2026-09-21
feat: support SQL Server table locking hints

minleejae:
Common SQL Server queries such as `SELECT * FROM jobs WITH (UPDLOCK, ROWLOCK)` fail because SQLServerHints only recognizes INDEX and NOLOCK. This prevents parsing locking reads and work-queue queries that combine UPDLOCK, ROWLOCK and READPAST.

Extend the existing SQLServerHints model with a typed, mutable list of 15 additional argument-free locking hints. Preserve their order within that list and retain the existing INDEX and nullable NOLOCK accessors/output. Reuse the same hint production for table sources and the shared aliased DML-target path, covering UPDATE and DELETE targets as well as SELECT, JOIN and CTE sources. Unknown hint names and argument forms are rejected rather than stored as arbitrary strings.

The scope is comma-separated locking hints inside `WITH (...)`. Parameterized access-path hints, additional INDEX argument forms and deprecated WITH/comma omissions are outside this change. Hint compatibility, isolation-level requirements and execution effects remain database checks.

Syntax reference: [Microsoft SQL Server table hints](https://learn.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-table?view=sql-server-ver17).

Validation:

- Regression tests cover every new locking hint, mixed-case names, combinations with INDEX, aliases, quoted names, JOINs, CTEs, UPDATE/DELETE targets, AST mutation, table discovery, legacy NOLOCK behavior, round trips and malformed hints. SQL Server server execution was not performed.
- Full Java 17 Gradle `check` passed, including grammar ambiguity, formatting, Checkstyle, PMD, SpotBugs, coverage and 8015 JUnit tests (0 failures/errors, 25 skipped).


manticore-projects:
Thank you!

PR #2674, opened 2026-09-22
fix: support comment column names in ALTER TABLE

minleejae:
`ALTER TABLE t MODIFY COLUMN comment TEXT` previously entered the column-comment grammar branch, while DROP/RENAME/CHANGE could reject the unquoted name. Require the complete COMMENT-literal lookahead and accept `comment` in the shared ALTER identifier helper.

The existing column-definition/name APIs remain usable for inspecting and editing these actions. Includes a usage example and regressions for ADD/MODIFY/CHANGE/DROP/RENAME, parsed AST mutation, statement deparsing and reparsing, with default and MYSQL configurations.

Validation: full Gradle `check` on Java 17 (including grammar ambiguity, formatting, Checkstyle, PMD and tests); six valid DDL fixtures executed both before and after parser rendering on MySQL 8.4, with catalog checks for resulting names/types. The grammar fix is shared; no dialect restriction is added to previously supported syntax.


manticore-projects:
Thank you!

PR #2675, opened 2026-09-22
feat: support MySQL named foreign key and check drops

minleejae:
In MYSQL mode, support `ALTER TABLE t DROP FOREIGN KEY fk1` and `DROP CHECK c1`, including quoted names and combined ALTER actions. Expose the deletion target through `getConstraintName()` and the action through `DROP_FOREIGN_KEY` / new `DROP_CHECK`, so callers can inspect, rename or construct the action without editing raw SQL.

Share special DROP rendering between parser-created actions and directly constructed `AlterExpression` nodes. Preserve the existing parenthesized foreign-key form and other DROP branches. Includes usage documentation, AST mutation/construction tests and parse/deparse/reparse checks.

Validation: full Gradle `check` on Java 17; six valid DDL fixtures executed in original and rendered form on MySQL 8.4, with catalog checks confirming the target constraints were removed. New named alternatives are selected only by `Dialect.MYSQL`.

Syntax reference: https://dev.mysql.com/doc/refman/8.4/en/alter-table.html


manticore-projects:
Thank you!

PR #2676, opened 2026-09-22
feat: attach PostgreSQL constraints to existing indexes

minleejae:
In POSTGRESQL mode, parse `ALTER TABLE t ADD [CONSTRAINT name] UNIQUE|PRIMARY KEY USING INDEX index_name`, including deferrability attributes and combined ALTER actions.

Expose this as `ConstraintUsingIndex`, with the new constraint name separated from `getExistingIndexName()`. Callers can inspect, edit or construct the reference; rendering is shared by `toString()` and the statement deparser. Metadata validation distinguishes the newly declared constraint from the existing index, and table discovery does not report the index as a table. Includes API usage documentation and regressions for ordinary constraints and Oracle's existing USING INDEX tail.

Validation: full Gradle `check` on Java 17; four fixtures executed in original and rendered form on PostgreSQL 18, checking `pg_constraint` for the resulting name, kind, attached index and deferrability.

Syntax reference: https://www.postgresql.org/docs/18/sql-altertable.html


manticore-projects:
Thank you!

PR #2677, opened 2026-09-22
fix: accept identity RESTART values without WITH

minleejae:
Accept the optional-WITH form `ALTER TABLE t ALTER COLUMN id RESTART 20`, reusing the existing signed sequence-value parser and `IdentityAlteration` AST. Both spellings expose `getRestartWith()` and render consistently as `RESTART WITH 20`; a null value continues to represent bare RESTART.

Add `withRestartWith` for direct construction, document the value semantics, and cover editing, signed values, adjacent identity operations, multi-action ALTER and default/POSTGRESQL parsing. This common grammar correction preserves the existing RESTART WITH behavior.

Validation: full Gradle `check` on Java 17; original and rendered DDL executed on PostgreSQL 18, followed by an INSERT verifying that the next generated identity value is 20.

Syntax reference: https://www.postgresql.org/docs/18/sql-altertable.html


manticore-projects:
Thank you!

PR #2678, opened 2026-09-22
feat: model qualified PostgreSQL exclusion operators

minleejae:
In POSTGRESQL mode, support exclusion keys such as `c WITH OPERATOR(pg_catalog.&&)` in both CREATE TABLE and ALTER TABLE ADD, including quoted schemas and unqualified OPERATOR wrappers.

Add editable `ExclusionOperator` references with separate schema, symbol and wrapper flag, exposed through `Index.ColumnParams.getExclusionOperatorReference()`. Keep the existing string getter/setter compatible and share qualified-operator grammar with extension DDL. Both statement renderers consume the same AST. Includes construction/mutation, multiple-key, expression-key and legacy-syntax tests plus usage documentation.

Validation: full Gradle `check` on Java 17; original and rendered DDL executed on PostgreSQL 18, checking that the exclusion constraint uses `pg_catalog.&&` in `pg_constraint`/`pg_operator`.

Syntax reference: https://www.postgresql.org/docs/18/sql-createtable.html


manticore-projects:
Thank you!

PR #2679, opened 2026-09-22
feat: support PostgreSQL CHECK NO INHERIT

minleejae:
In POSTGRESQL mode, support `CHECK (expression) NO INHERIT`, including ALTER ADD with `NOT VALID`. Expose inheritance as `CheckConstraint.isNoInherit()` / `setNoInherit()` / `withNoInherit()`, independently of the existing validation flag.

Reuse the CHECK production for PostgreSQL column constraints as well as CREATE/ALTER table constraints, so column-level expressions remain structured and visitable. Both renderers read the editable AST. Includes construction/mutation, existing ENFORCED options, expression traversal and parse/deparse/reparse tests plus usage documentation.

Validation: full Gradle `check` on Java 17; original and rendered DDL executed on PostgreSQL 18 for ALTER plus named/unnamed column and table CREATE forms. Verified the ALTER constraint's `connoinherit=true` and `convalidated=false` in `pg_constraint`.

Syntax references: https://www.postgresql.org/docs/18/sql-createtable.html and https://www.postgresql.org/docs/18/sql-altertable.html


manticore-projects:
Thank you!

PR #2680, opened 2026-09-22
feat: model PostgreSQL table-level NOT NULL constraints

minleejae:
In POSTGRESQL mode, support PostgreSQL 18 table constraints such as `ALTER TABLE t ADD CONSTRAINT nn NOT NULL id NOT VALID` and `CREATE TABLE t (id INT, CONSTRAINT nn NOT NULL id)`, including unnamed constraints and NO INHERIT.

Expose a `NotNullConstraint` with an editable target `Column`, optional constraint name, inheritance flag and existing `ConstraintAttributes` validation flag. Add `Index.Kind.NOT_NULL`; wire the column through table traversal, expression deparsers and validation, and distinguish the newly declared constraint from an index. Includes construction/mutation, custom deparsing, metadata validation, legacy column nullability and round-trip tests plus usage documentation.

Validation: full Gradle `check` on Java 17; original and rendered DDL executed on PostgreSQL 18 for ALTER and CREATE forms, including quoted identifiers, NO INHERIT and combined ALTER actions. Verified the named ALTER constraint is recorded as `contype=n`, `convalidated=false`.

Syntax references: https://www.postgresql.org/docs/18/sql-createtable.html and https://www.postgresql.org/docs/18/sql-altertable.html


manticore-projects:
Thank you!

